PR #20, opened 2023-07-13
Pause Rules Bool  added to handler

### LINKED ISSUE #15: Access Module - Global Rules - Pause Rules Toggle On/Off
This ticket was created from design meeting 6/26/23 in order to implement functionality to toggle pause rules in App Manager to keep parity with the access Tier and Risk rules. 

Pause rules check should be turned on/off via boolean in protocolApplicationHandler.sol.

TJ-Everett:
moved Pause Rules changes over to new branch


TJ-Everett:
This is the same changes made in [ticket 342](https://github.com/thrackle-io/tron_legacy/pull/356).

ShaneDuncan602:
Are we sure this should go into internal?

ShaneDuncan602 (changes requested):
Can we add a little statement to the Pause Window section of the Rules Guide that explains this toggle on/off? Also, I verified with @jjbailey-io. We do want this to go to external.

TJ-Everett:
> Are we sure this should go into internal?

This is a great question. It was pointed to internal because we were not going to get it into the audit, now that is not a concern. 
Easy path is to put it into internal now, then sort through the conflicts later. 
Given the audit is over the "normal" path would be to branch from external so I totally see your point.

TJ-Everett:
PR moved to new branch [here](https://github.com/thrackle-io/tron/pull/137).

Issue #25, opened 2023-07-17
Audit - Missing access control in handlers (3)


ShaneDuncan602:
For detailed findings on this issue, reference [Finding 3](https://drive.google.com/file/d/1YEockBPB2IcMVV3iLXCRsipiSDLbd3Rp/view?ts=64b5552c) in the document.

Fix Plan
- Add Ownable to the Handlers and put onlyOwner on checkAllRules and checkAllApplicationRules
- Add a parameter to the Handlers' constructor for the owner contract(this will be the asset contract)
- For ProtocolERC721, Protocol721U, Protocol721Umin, and Protocol721a, access setERC721Address from the constructor to save unnecessary deployment steps.
- Thoroughly test deployments and include fail tests for accessing checkAllRules and checkAllApplicationRules directly. 

These changes should be done in branch, TOB-THR-3, cut from External.

ShaneDuncan602:
Ready for merge

ShaneDuncan602:
I missed the application handler. It needs the same mods

ShaneDuncan602:
Ready for merge

PR #48, opened 2023-07-19
Audit 3 - Make handlers ownable by their controlling asset

ShaneDuncan602:


TJ-Everett (commented):
Since this will be after my zero address changes we should check that the _assetAddress is not zero.

ShaneDuncan602:
I agree. Good call.

TJ-Everett (approved):
Approved

TJ-Everett (approved):
Approved App Handler Changes

jjbailey-io:
#25

ShaneDuncan602:
merged with external

PR #51, opened 2023-07-20
refactor _removeAddress logic and test

TJ-Everett:


src/application/AppManager.sol
```diff
@@ -624,17 +630,17 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
         if (_addressArray.length == 0) {
```
ShaneDuncan602, line 630: Can we add the ToB suggested NatSpec to this function as well?

ShaneDuncan602 (changes requested):
Looks good. Just one NatSpec thing

ShaneDuncan602 (approved):
Perfect!

ShaneDuncan602:
merged with external

PR #60, opened 2023-07-25
inconsistent implementation of risk processing (17)

TJ-Everett:


ShaneDuncan602 (approved):
Looks good.

src/economic/ruleProcessor/RiskTaggedRuleProcessorFacet.sol
```diff
@@ -21,6 +21,18 @@ contract RiskTaggedRuleProcessorFacet is IRuleProcessorErrors, IRiskErrors {
...
      * @param _amountToTransfer total USD amount to be transferred with 18 decimals of precision
+     * @notice _transactionSize size must be equal to _riskLevel + 1 since the _transactionSize must
+     * specify the maximum tx size for anything below the first level and between the highest risk score and 100. This also
+     * means that the positioning of the arrays is ascendant in terms of risk levels, and
+     * descendant in the size of transactions. (i.e. if highest risk level is 99, the last balanceLimit
+     * will apply to all risk scores of 100.)
+     * eg.
+     * risk scores      TxLimit         resultant logic
+     * -----------      --------         ---------------
+     *    25             1000            0-24  =  1000
```
oscarsernarosero, line 32: Haven't seen the code yet, but this documentation is already wrong. There shouldn't be any limit in the first range which makes the txLimit array and the risk score array to have the same length (not +1 on the TxLimit as illustrated here). Therefore 0-24 should have no limit other than the imposed by the data size.

src/economic/ruleProcessor/RiskTaggedRuleProcessorFacet.sol
```diff
@@ -29,9 +41,9 @@ contract RiskTaggedRuleProcessorFacet is IRuleProcessorErrors, IRiskErrors {
         TaggedRules.TransactionSizeToRiskRule memory rule = data.getTransactionLimitByRiskRule(_ruleId);
         ///If risk score is within the rule riskLevel array, find the maxSize for that risk Score
         for (uint256 i; i < rule.riskLevel.length; ) {
-            if (_riskScore <= rule.riskLevel[i]) {
+            if (_riskScore < rule.riskLevel[i]) {
```
oscarsernarosero, line 44: Actual table should be something like:

| Risk Score | TxLimit | Means |
| - | - | - |
| 0* | No Limit * | 0 -> 24: No Limit * |
| 25 | 1000 | 25 -> 49: 1000 max |
| 50 | 500 | 50 -> 74: 500 max |
| 75 | 250 |  75 -> 100: 250 max |

##### * This refers to implied values and therefore it is not necessary to be written as part of the rule (They are already part of every rule implicitly )

This means that this logic here should be refactored since current logic would behave differently than expected.

oscarsernarosero: Assuming that both input arrays have the same length, there are a couple things that this in bringing up here and most likely needs to be fixed in the other rules as well:

- We might want to check if the risk score is less than the lowest risk score in the rule since this would mean that the risk score is too low, which means it has no limit ( 0 - lowest risk score) and therefore it automatically passes the check (basically no check needed) This would reward well-behaved users with low-cost transactions.

- This also means that the for loop should start with *i* on *1* and not on *0*.  ```for (uint256 i=1; i < rule.riskLevel.length ; )```

It would be beneficial to discuss with @ShaneDuncan602  about the the limits being inclusive or exclusive at the bottom or at the top, and then making sure all the rules have the same strategy. I personally keep going back and forth with this, but right now --if you ask me-- I think the rules being inclusive at the bottom and exclusive at the top makes sense, but this also mean that rules can have risk scores of 100. Since these rules have been touched many times by some of us, this is the chance we have to standardize them.

oscarsernarosero (changes requested):
Some changes to the logic and documentation is needed. We should make sure all the risk rules are written in the same manner in this ticket, and make sure the tests are thorough enough to catch any edge-case misbehavior.

src/economic/ruleProcessor/RiskTaggedRuleProcessorFacet.sol
oscarsernarosero: Ok sorry. I think I was confused about the state of the implementation of the rules. Since we still have the implementation of the rule with different array sizes, this logic is perfect. We're good to go!

oscarsernarosero (approved):
Since we still have the old implementation in place, this logic adjustments are perfect.

TJ-Everett:
This is now issue 17 in the ToB final report

ShaneDuncan602:
merged with external

PR #61, opened 2023-07-27
External 55

### LINKED ISSUE #55: Audit - Access level zero rule checks can be bypassed(19)
For detailed findings on this issue, reference [Week 2 Finding 20](https://drive.google.com/file/d/1Bb3Bb9IVtHyieh0uXDiBDpRM5t_2DiZ6/view?usp=sharing) ->19 in the document.

Fix Plan
- Change logic as suggested so that AccessLevel0 rule may not be bypassed for anything except AMM.

These changes should be done in branch, TOB-THR-19, cut from External.

TJ-Everett:
access Level 0 bypassed

### DIFF
--- a/src/application/ProtocolApplicationHandler.sol
+++ b/src/application/ProtocolApplicationHandler.sol
@@ -126,8 +126,13 @@ contract ProtocolApplicationHandler is Ownable, AppAdministratorOnly, IApplicati
     function _checkAccessLevelRules(address _from, address _to, uint128 _usdBalanceValuation, uint128 _usdAmountTransferring) internal {
         uint8 score = appManager.getAccessLevel(_to);
         uint8 fromScore = appManager.getAccessLevel(_from);
+        /// Check if recipient is AMM (AMM is not exempt but does not have an accessLevel)
         if (AccessLevel0RuleActive && appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(fromScore);
-        if (AccessLevel0RuleActive && !appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(score);
+        /// if reciepent is not an AMM check both to and from access level) 
+        if (AccessLevel0RuleActive && !appManager.isRegisteredAMM(_to)) {
+            ruleProcessor.checkAccessLevel0Passes(score);
+            ruleProcessor.checkAccessLevel0Passes(fromScore);
+        }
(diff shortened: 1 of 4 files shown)


src/application/ProtocolApplicationHandler.sol
```diff
@@ -126,8 +126,13 @@ contract ProtocolApplicationHandler is Ownable, AppAdministratorOnly, IApplicati
     function _checkAccessLevelRules(address _from, address _to, uint128 _usdBalanceValuation, uint128 _usdAmountTransferring) internal {
         uint8 score = appManager.getAccessLevel(_to);
         uint8 fromScore = appManager.getAccessLevel(_from);
+        /// Check if recipient is AMM (AMM is not exempt but does not have an accessLevel)
         if (AccessLevel0RuleActive && appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(fromScore);
-        if (AccessLevel0RuleActive && !appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(score);
+        /// if reciepent is not an AMM check both to and from access level)
```
ShaneDuncan602: typo on recipient

src/application/ProtocolApplicationHandler.sol
```diff
@@ -126,8 +126,13 @@ contract ProtocolApplicationHandler is Ownable, AppAdministratorOnly, IApplicati
...
         uint8 score = appManager.getAccessLevel(_to);
         uint8 fromScore = appManager.getAccessLevel(_from);
+        /// Check if recipient is AMM (AMM is not exempt but does not have an accessLevel)
         if (AccessLevel0RuleActive && appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(fromScore);
-        if (AccessLevel0RuleActive && !appManager.isRegisteredAMM(_to)) ruleProcessor.checkAccessLevel0Passes(score);
+        /// if reciepent is not an AMM check both to and from access level) 
+        if (AccessLevel0RuleActive && !appManager.isRegisteredAMM(_to)) {
+            ruleProcessor.checkAccessLevel0Passes(score);
+            ruleProcessor.checkAccessLevel0Passes(fromScore);
+        }
```
ShaneDuncan602: Wouldn't there be an issue if the rule is active and the from address is an AMM? When the from score is checked it will be an AMM and not have a score. Does it not work correctly if we make the change they suggest?

TJ-Everett: Yes, I think I miss-understood their suggestion in my confusion of making sure the rule still applied the way we want it too. I have that updated to reflect their suggested change and our new updated tests.

ShaneDuncan602 (approved):
Looks good

TJ-Everett:
This is now issue 19 in ToB final report

ShaneDuncan602:
merged with external

PR #62, opened 2023-07-28
Audit 6 - 2 Step process for critical ops

ShaneDuncan602:
Recommendations
Short term, implement a two-step process for all irrecoverable critical operations. Split all
functions that set contracts owners and administrators into two functions:
1. proposeOwner() or proposeAppManagerAddress() which proposes a new contract
owner or manager which is saved in the contract storage,
2. confirmOwner() or confirmAppManagerAddress() which can only be called by the
proposed owner or manager and is used to finalize the transfer

This will guarantee that the owner-setting party must be able to call the contract from the
address in question before it is actually set as the new owner.

### DIFF
--- a/src/application/AppManager.sol
+++ b/src/application/AppManager.sol
@@ -2,20 +2,21 @@
 pragma solidity 0.8.17;
 
 import "openzeppelin-contracts/contracts/access/AccessControlEnumerable.sol";
-import "../data/Accounts.sol";
-import "../data/IAccounts.sol";
-import "../data/IAccessLevels.sol";
-import "../data/AccessLevels.sol";
-import "../data/IRiskScores.sol";
-import "../data/RiskScores.sol";
-import "../data/IGeneralTags.sol";
-import "../data/GeneralTags.sol";
-import "../data/IPauseRules.sol";
-import "../data/PauseRules.sol";
-import "./ProtocolApplicationHandler.sol";
+import "src/data/Accounts.sol";
+import "src/data/IAccounts.sol";
+import "src/data/IAccessLevels.sol";
+import "src/data/AccessLevels.sol";
+import "src/data/IRiskScores.sol";
+import "src/data/RiskScores.sol";
+import "src/data/IGeneralTags.sol";
+import "src/data/GeneralTags.sol";
+import "src/data/IPauseRules.sol";
+import "src/data/PauseRules.sol";
+import "src/application/ProtocolApplicationHandler.sol";
 import "src/economic/ruleProcessor/ActionEnum.sol";
-import {IAppLevelEvents} from "../interfaces/IEvents.sol";
-import "./IAppManagerUser.sol";
+import {IAppLevelEvents} from "src/interfaces/IEvents.sol";
+import "src/application/IAppManagerUser.sol";
+import "src/data/IDataModule.sol";
 
 /**
  * @title App Manager Contract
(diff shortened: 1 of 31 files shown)


src/application/IAppManagerUser.sol
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @title App Manager User Interface
+ * @dev This interface is implemented by all contracts that use AppManager. It provides the common function for setting up a new link to an AppManager
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett
+ * @notice Interface for app manager user functions.
+ */
+interface IAppManagerUser {
+    /**
+     * @dev this function confirms a new appManagerAddress that was put in storageIt can only be confirmed by the proposed address
```
TJ-Everett: This* and missing a . after storage

TJ-Everett (approved):
Great work Shane this was high effort for sure. Just the one little typo. Approving

src/data/DataModule.sol
```diff
@@ -27,19 +30,28 @@ contract DataModule is IDataModule, Ownable {
...
     /**
-     * @dev Transfers ownership of the contract to a new account (`newOwner`).
-     * Can only be called by the current owner.
+     * @dev this function confirms a new appManagerAddress that was put in storage. It can only be confirmed by the proposed address
      */
-    function transferDataOwnership(address newOwner) public appAdminstratorOrOwnerOnly {
-        transferOwnership(newOwner);
+    function confirmOwner() external {
+        if (newOwner == address(0)) revert NoProposalHasBeenMade();
+        if (msg.sender != newOwner) revert ConfirmerDoesNotMatchProposedAddress();
```
oscarsernarosero, line 46: Interesting. So the confirmation has to be made from the new owner? kinda like accepting the ownership?

oscarsernarosero: I love it Shane

src/data/DataModule.sol
```diff
@@ -27,19 +30,28 @@ contract DataModule is IDataModule, Ownable {
...
+        _transferOwnership(newOwner);
+    }
+
+    /**
+     * @dev Part of the two step process to set a new Data Provider within a Protocol AppManager
+     * @param _providerType the type of data provider
+     */
+    function confirmDataProvider(ProviderType _providerType) external virtual appAdminstratorOrOwnerOnly {
```
oscarsernarosero, line 54: Shouldn't we re-enter the address here to make sure we proposed the right address in the first step?

src/application/AppManager.sol
```diff
@@ -812,47 +818,87 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
         appName = _appName;
     }
 
+    /**
+     * @dev Part of the two step process to set a new AppManager within a Protocol Entity
+     * @param _assetAddress address of a protocol entity that uses AppManager
+     */
+    function confirmAppManager(address _assetAddress) external onlyAppAdministrator {
+        IAppManagerUser(_assetAddress).confirmAppManagerAddress();
+    }
```
oscarsernarosero, line 827: Super nice

oscarsernarosero (approved):
Shane this looks really really amazing. Fantastic job

src/data/DataModule.sol
ShaneDuncan602: It has a check in the appManager to make sure the sender is the new proposed data provider.

ShaneDuncan602:
merged with external

PR #86, opened 2023-08-03
dataModuleAppManagerAddress can’t be set in DataModule contract(28)

### LINKED ISSUE #83: Audit - dataModuleAppManagerAddress can’t be set in DataModule contract(28)
See issue 28 in [ToB Final Report](https://drive.google.com/file/d/1JmY1U7RlM43-t1Dz_wcGBicltCMxy33A/view?usp=sharing)

The dataModuleAppManagerAddress variable cannot be set in the DataModule contract because the setter function is protected by a modifier that will revert if the value of dataModuleAppManagerAddress is equal to address(0). Since the contract instantiates address variables with address(0) as the default, the setter function to update the variable will always revert. This issue will impact the transferDataOwnership function, which is protected by the same modifier. As a result, the contract will be left in an unusable state.

Fix Plan
- mark the DataModule contract as abstract and set the dataModuleAppManagerAddress in the constructor rather than relying on the developer to set it in the constructor of the top-level contract.

TJ-Everett:


TJ-Everett:
This PR is closed as we moved to [here](https://github.com/thrackle-io/tron/pull/92).

PR #87, opened 2023-08-03
Rule validations, Audit 25

### LINKED ISSUE #80: Audit - RuleDataFacet functions and data structures are misconfigured(25)
See issue 25 in [ToB Final Report](https://drive.google.com/file/d/1JmY1U7RlM43-t1Dz_wcGBicltCMxy33A/view?usp=sharing)

Multiple functions in the RuleDataFacet contract are missing validation or are not aligned with the data structure definitions. This indicates that either the function implementations or the data structures are incorrect, which can lead to unexpected behavior.

Fix Plan
- Every rule should have data validation to ensure only expected values may be entered into the rules. This validation should match any inline comments and be included in the natspec comment

ShaneDuncan602:
Some of the requested rule validations were part of audit issue 1 and were already addressed. Some of the setter parameter changes requested were not done. This was already addressed in audit issue 1. It pertains to the parameter being named startingTime but it only storing the hour. Additionally, I changed all the rules to be the same format where it is a mapping within a struct. It didn't make sense for them to be different. Also removed a failing test because it will be addressed by audit 23

This closes #80

### DIFF
--- a/src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
+++ b/src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
@@ -14,7 +14,7 @@ import "./IOracle.sol";
  * @dev Facet in charge of the logic to check token rules compliance
  * @notice Implements Token Fee Rules on Accounts.
  */
-contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors{
+contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors {
     /**
      * @dev Check if transaction passes minTransfer rule.
      * @param _ruleId Rule Identifier for rule arguments
@@ -24,8 +24,8 @@ contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors{
         RuleDataFacet data = RuleDataFacet(Diamond.ruleDataStorage().rules);
 
         if (data.getTotalMinimumTransferRules() != 0) {
-            try data.getMinimumTransferRule(_ruleId) returns (uint min) {
-                if (min > amountToTransfer) revert BelowMinTransfer();
+            try data.getMinimumTransferRule(_ruleId) returns (NonTaggedRules.TokenMinimumTransferRule memory rule) {
+                if (rule.minTransferAmount > amountToTransfer) revert BelowMinTransfer();
             } catch {
                 revert RuleDoesNotExist();
             }
@@ -155,8 +155,8 @@ contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors{
             if (rule.totalSupply != 0) {
                 _supply = rule.totalSupply;
             }
-            // we check the numbers against the rule  
-            if ((_volume * 100000000) / _supply >= uint(rule.maxVolume) * 10000 ) {
+            // we check the numbers against the rule
+            if ((_volume * 100000000) / _supply >= uint(rule.maxVolume) * 10000) {
                 revert TransferExceedsMaxVolumeAllowed();
             }
         } catch {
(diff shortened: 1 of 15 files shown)


src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
```diff
@@ -24,8 +24,8 @@ contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors{
         RuleDataFacet data = RuleDataFacet(Diamond.ruleDataStorage().rules);
 
         if (data.getTotalMinimumTransferRules() != 0) {
```
oscarsernarosero, line 26: I think we should also validate here that the ruleId is valid ( < data.getTotalMinimumTransferRules() ). If this rule is stored in an array, this can be done using a try-catch clause, but it if it is stored in a mapping, then we can just compare total rules against the ruleId.

src/economic/ruleStorage/RuleDataFacet.sol
```diff
@@ -170,14 +170,21 @@ contract RuleDataFacet is Context, AppAdministratorOnly, IEconomicEvents, IInput
...
-        if (_maxVolatility == 0 || _blocksPerPeriod == 0 || _hoursFrozen == 0) revert ZeroValueNotPermited();
+    function addVolatilityRule(
+        address _appManagerAddr,
+        uint16 _maxVolatility,
+        uint8 _period,
+        uint8 _startingTime,
+        uint256 _totalSupply
+    ) external appAdministratorOnly(_appManagerAddr) returns (uint32) {
+        if (_maxVolatility == 0 || _period == 0 || _startingTime == 0) revert ZeroValueNotPermited();
+        if (_maxVolatility > 10000) revert ValueOutOfRange(_maxVolatility);
```
oscarsernarosero, line 185: so volatility can't be more than 100% (2 Xs)? we should probably ask about this to NEE since one could argue that some devs might want to allow a max volatility of 3 Xs (200%) for instance

src/economic/ruleStorage/RuleDataFacet.sol
```diff
@@ -228,10 +235,12 @@ contract RuleDataFacet is Context, AppAdministratorOnly, IEconomicEvents, IInput
         if (_maxVolumePercentage > 9999) revert ValueOutOfRange(_maxVolumePercentage);
         if (_maxVolumePercentage < 100) revert ValueOutOfRange(_maxVolumePercentage);
```
oscarsernarosero: _maxVolumePercentage can't be set to less than 1%? maybe it's a very well distributed coin and they want to limit transfers to 0.5% which could be a lot of money. I would use ZeroValueNorPermited error instead

test/ApplicationERC20.t.sol
```diff
@@ -238,13 +238,9 @@ contract ApplicationERC20Test is DiamondTestUtil, RuleProcessorDiamondTestUtil {
         // Finally, check the invalid type
         vm.stopPrank();
         vm.startPrank(defaultAdmin);
+        bytes4 selector = bytes4(keccak256("InvalidOracleType(uint8)"));
+        vm.expectRevert(abi.encodeWithSelector(selector, 2));
```
oscarsernarosero, line 242: nice

test/ApplicationERC20Handler.t.sol
```diff
@@ -408,21 +408,14 @@ contract ApplicationERC20HandlerTest is Test, DiamondTestUtil, RuleProcessorDiam
...
-        _index = RuleDataFacet(address(ruleStorageDiamond)).addOracleRule(ac, 2, address(oracleAllowed));
-        /// connect the rule to this handler
-        applicationCoinHandler.setOracleRuleId(_index);
-        vm.expectRevert(0x2a15491e);
-        applicationCoinHandler.checkAllRules(20, 0, user1, address(69), 10, ActionTypes.TRADE);
-
         /// let's turn the rule off
         applicationCoinHandler.activateOracleRule(false);
-        applicationCoinHandler.checkAllRules(20, 0, user1, address(69), 10, ActionTypes.TRADE);
+        applicationCoinHandler.checkAllRules(20, 0, user1, address(88), 10, ActionTypes.TRADE);
```
oscarsernarosero, line 413: lol

oscarsernarosero (approved):
It looks really good. I just have a couple questions about input validation.

src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
ShaneDuncan602: This will be addressed by the other audit issue. All the rule storage is being changed to use a mapping in this ticket, but one of the other ones adds reverts when the rule is not found. It's audit issue 23

oscarsernarosero: oh ok I see. I would take care of this here since you are switching from array to mapping here, so the issue in this specific file was not in the audited code, and therefore will not be listed in the docs. Just to be safe.

ShaneDuncan602:
merged with external

PR #88, opened 2023-08-04
Audit 23,  Multiple functions return an empty rule instead of reverting

### LINKED ISSUE #78: Audit - Multiple functions return an empty rule instead of reverting(23)
See issue 20 in [ToB Final Report](https://drive.google.com/file/d/1JmY1U7RlM43-t1Dz_wcGBicltCMxy33A/view?usp=sharing)

Multiple rule checking functions use the try/catch syntax expecting invalid rules to revert and trigger the catch clause body in order to revert the transaction. However, some of the getter functions will return an empty rule instead of reverting, which can lead to the system not behaving as expected.

Fix Plan
- All the rule data structures need to be assessed to ensure that they are the right format. All the rule structs within IRuleStorage need to conform to the standard of a mapping of the rule single rule struct(from RuleDataInterfaces) and a uint32 rule index variable within the rule collection struct(in IRuleStorage)
- Add validation to all rule getters to ensure that getting a nonexistent rule reverts. This validation should be done in a common function in an abstract function that all the rule storage facets have access to. This may have to be put into its own facet.
- Add validation to all the rule setters in the handlers to check for the existence of the rule prior to allowing it to be set. 

- NOTE: this should use a standard function that can be shared by all the handlers.

ShaneDuncan602:
Added custom errors to all rule getters. NOTE: This does not include the tagged rules because they must check differently due to a user having a tag that is not covered by the rule.

This closes #78

### DIFF
--- a/src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
+++ b/src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
@@ -1,0 +1,226 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import {RuleProcessorDiamondLib as Diamond, RuleDataStorage} from "./RuleProcessorDiamondLib.sol";
+import {FeeRuleDataFacet} from "src/economic/ruleStorage/FeeRuleDataFacet.sol";
+import {TaggedRuleDataFacet} from "src/economic/ruleStorage/TaggedRuleDataFacet.sol";
+import {RuleDataFacet} from "src/economic/ruleStorage/RuleDataFacet.sol";
+import {AppRuleDataFacet} from "src/economic/ruleStorage/AppRuleDataFacet.sol";
+import {IFeeRules as Fee, ITaggedRules as TaggedRules, INonTaggedRules as NonTaggedRules} from "../ruleStorage/RuleDataInterfaces.sol";
+import "src/economic/ruleStorage/RuleStorageCommonLib.sol";
+import "src/data/PauseRule.sol";
+import "src/application/IAppManager.sol";
+import "forge-std/console.sol";
+
+/**
+ * @title Fee Rule Processor Facet Contract
+ * @author @ShaneDuncan602 @oscarsernarosero @TJ-Everett
+ * @dev Facet in charge of the logic to check fee rule compliance
+ * @notice Implements Token Fee Rules on Accounts.
+ */
+contract RuleApplicationValidationFacet {
+    using RuleStorageCommonLib for uint32;
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateAMMFee(uint32 _ruleId) external view {
+        FeeRuleDataFacet data = FeeRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
+        _ruleId.checkRuleExistence(data.getTotalAMMFeeRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateTransactionLimitByRiskScore(uint32 _ruleId) external view {
+        TaggedRuleDataFacet data = TaggedRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
+        _ruleId.checkRuleExistence(data.getTotalTransactionLimitByRiskRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateMinMaxAccountBalanceERC721(uint32 _ruleId) external view {
+        TaggedRuleDataFacet data = TaggedRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
+        _ruleId.checkRuleExistence(data.getTotalBalanceLimitRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateNFTTransferCounter(uint32 _ruleId) external view {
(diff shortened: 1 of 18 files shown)


src/economic/ruleStorage/RuleDataFacet.sol
```diff
@@ -62,6 +61,8 @@ contract RuleDataFacet is Context, AppAdministratorOnly, IEconomicEvents, IInput
      */
     function getPctPurchaseRule(uint32 _index) external view returns (NonTaggedRules.TokenPercentagePurchaseRule memory) {
         RuleS.PctPurchaseRuleS storage data = Storage.pctPurchaseStorage();
+        // check one of the required non zero values to check for existence, if not, revert
+        if (data.percentagePurchaseRules[_index].startTime == 0) revert RuleDoesNotExist();
```
oscarsernarosero: This is a way of doing it. I would say a more standard way of doing the same is by making sure the _index (ruleId) is not greater or equal to the total amount of rules (ruleIndex, or percentagePurchaseRuleIndex in this case). This should work with all rules even if they don't have a startTime. This current implementation works as good tho. So I would say it's up to you.

oscarsernarosero (approved):
This implementation works perfectly fine. I just have a suggestion to make the check more standard. I'll approve and I'll let you decide on that.

oscarsernarosero (commented):
I love your new solution Shane

TJ-Everett (approved):
This took a minute to understand the logic here but I like this a lot. Great work Shane.

src/economic/ruleStorage/RuleStorageCommonLib.sol
```diff
@@ -17,4 +18,18 @@ library RuleStorageCommonLib {
...
     }
+
+    /**
+     * @dev generic function to check the existence of a rule
+     * @param _ruleIndex index of the current rule
+     * @param _ruleTotal total rules in existence for the rule type
+     * @return _exists true if it exists, false if not
+     */
+    function checkRuleExistence(uint32 _ruleIndex, uint32 _ruleTotal) internal pure returns (bool) {
+        if ((_ruleTotal > 0 && _ruleTotal <= _ruleIndex) || _ruleTotal == 0) {
```
oscarsernarosero: just for efficiency purposes, it should start with the ```_ruleTotal == 0``` as the first condition, and you wouldn't need ```_ruleTotal > 0```. So it would be:
```
if( _ruleTotal == 0 || _ruleTotal <= _ruleIndex)
```

src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
```diff
@@ -0,0 +1,226 @@
...
+contract RuleApplicationValidationFacet {
+    using RuleStorageCommonLib for uint32;
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateAMMFee(uint32 _ruleId) external view {
+        FeeRuleDataFacet data = FeeRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
```
oscarsernarosero: this documentation needs to be updated.

src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
```diff
@@ -0,0 +1,226 @@
...
+        _ruleId.checkRuleExistence(data.getTotalAMMFeeRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateTransactionLimitByRiskScore(uint32 _ruleId) external view {
+        TaggedRuleDataFacet data = TaggedRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
```
oscarsernarosero: same here

src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
```diff
@@ -0,0 +1,226 @@
...
+        _ruleId.checkRuleExistence(data.getTotalTransactionLimitByRiskRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateMinMaxAccountBalanceERC721(uint32 _ruleId) external view {
+        TaggedRuleDataFacet data = TaggedRuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
```
oscarsernarosero: here too

src/economic/ruleProcessor/RuleApplicationValidationFacet.sol
```diff
@@ -0,0 +1,226 @@
...
+        _ruleId.checkRuleExistence(data.getTotalBalanceLimitRules());
+    }
+
+    /**
+     * @dev Validate the existence of the rule
+     * @param _ruleId Rule Identifier
+     */
+    function validateNFTTransferCounter(uint32 _ruleId) external view {
+        RuleDataFacet data = RuleDataFacet(Diamond.ruleDataStorage().rules);
+        // check one of the required non zero values to check for existence, if not, revert
```
oscarsernarosero: ok you get the idea. Basically all of them

src/economic/ruleStorage/RuleStorageCommonLib.sol
oscarsernarosero: Actually I think that just ```if(_ruleTotal <= _ruleIndex)``` is enough, but I remember adding the check for the zero rule with you once. I just don't know if that scenario still applies. Maybe some tests would help us figure out if the simpler ```if(_ruleTotal <= _ruleIndex)``` is enough. I know for sure ```if( _ruleTotal == 0 || _ruleTotal <= _ruleIndex)``` would do.

oscarsernarosero (approved):
This is brilliant Shane. Great work. I just pointed are some doc fixes, and I do suggest an improvement in the conditional for the ```checkExistance``` function.

ShaneDuncan602:
merged with external

PR #93, opened 2023-08-08
Audit 22, Percentage rules can be bypassed

ShaneDuncan602:
Removed all the percentage rule checks since they are not supported in this release.


TJ-Everett (approved):
Approved.

ShaneDuncan602:
merged to external

PR #96, opened 2023-08-09
Audit 24 - Admin Withdrawal Rule Bypass

ShaneDuncan602:


TJ-Everett (approved):
This is great work Shane.

src/token/ProtocolERC20Handler.sol
```diff
@@ -84,6 +88,13 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
         }
     }
 
+    /**
+     * @dev See {IERC165-supportsInterface}.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165) returns (bool) {
+        return interfaceId == type(IAdminWithdrawalRuleCapable).interfaceId || super.supportsInterface(interfaceId);
+    }
+
```
oscarsernarosero, line 97: cool

src/token/ProtocolERC20Handler.sol
```diff
@@ -496,22 +507,36 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+    function isAdminWithdrawalActiveAndApplicable() public view override returns (bool) {
+        bool active;
+        if (adminWithdrawalActive) {
+            try ruleProcessor.checkAdminWithdrawalRule(adminWithdrawalRuleId, 1, 1) {} catch {
+                active = true;
+            }
+        }
+        return active;
+    }
+
```
oscarsernarosero, line 531: shouldn't we put this function in a common file to avoid duplication of code?

oscarsernarosero (approved):
Super nice Shane. My only question is if we can put some code in one single file to avoid duplication of code and all the dangers associated with it.

src/token/ProtocolERC20Handler.sol
ShaneDuncan602: Well, the problem with that would be the adminWithdrawalRefId would also need to go into the ProtocolHandlerCommon or an interface to store the abi for getAdminWithdrawalRefId(). Since I already needed an abstract contract(IAdminWithdrawalRuleCapable) to define the abi inside the AppManager, I opted to keep these functions in the handler.  I couldn't find a cleaner way.

ShaneDuncan602:
https://github.com/thrackle-io/tron/pull/109/commits

PR #100, opened 2023-08-11
Configure New CHAIN ID for Dev Env

oni001:


bfcoursewool (commented):
Nice @oni001 , only question is where do we specify the actual chain id's now? Is that in AWS or does it need to be added to a .env somewhere or anything?

k8s/anvil/deployment.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         - name: RUST_LOG
           value: backend,api,node,rpc=warn
         - name: CHAIN_ID
-          value: "31338"
+          value: "{{CHAIN_ID}}"
```
oni001, line 31: @bfcoursewool here's where we specify the chain_id and the value is stored under the dev GitHub Environment. The workflow replaces CHAIN_ID with the value before deployment.

PR #101, opened 2023-08-14
Remove staking from branch

### LINKED ISSUE #98: Move Staking contracts from external to internal
Remove staking contracts from the external branch and port them to the internal branch.

TJ-Everett:


ShaneDuncan602 (approved):
Looks good.

oscarsernarosero (changes requested):
Not sure if it is necessary, but the Make file, the SUMMARY.md, and the README still has some **staking** lines. The AppManager has a **stakingList** and some functions to maintain it. Also take a look at the IAppManager. Also we still have the events. Not sure if we need to go that deep, so I would consult with @jjbailey-io

TJ-Everett:
Removed staking from AppManager, IAppManager, ProtocolERC20 (checked for isStakingRegistered()) and removed from the testing (ApplicationAppManager.t.sol)

oscarsernarosero (commented):
Suggested changes were addressed in a commit although not visible here for some reason. I approve the changes.

PR #102, opened 2023-08-15
52 audit missing validation safemint 16

oscarsernarosero:
Changes were made to make the repo npm-package compatible and to address issue 16 from the audit. The main changes were:

- Example ERC721 now has minting restricted to app administrator or contract owner by default (Fix for issue 16 from audit).
- Tests were fixed to accommodate for this new version of ApplicationERC721 example.
- Absolute paths were migrated to relative paths.
- One unnecessary import was removed.
- openzeppelin-contracts was replaced for @openzeppelin to make it npm compatible out of the box.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,5 +2,27 @@
   "devDependencies": {
     "prettier": "^2.8.3",
     "prettier-plugin-solidity": "^1.1.1"
-  }
+  },
+  "name": "@thrackle-io/tron",
+  "description": "[![Project Version][version-image]][version-url]",
+  "version": "1.0.0",
+  "main": "index.js",
+  "directories": {
+    "doc": "docs",
+    "lib": "lib",
+    "test": "test"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/thrackle-io/Tron.git"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/thrackle-io/Tron/issues"
+  },
+  "homepage": "https://github.com/thrackle-io/Tron#readme"
 }
--- a/package.json
+++ b/package.json
@@ -2,27 +2,5 @@
   "devDependencies": {
     "prettier": "^2.8.3",
     "prettier-plugin-solidity": "^1.1.1"
-  },
-  "name": "@thrackle-io/tron",
-  "description": "[![Project Version][version-image]][version-url]",
-  "version": "1.0.0",
-  "main": "index.js",
-  "directories": {
-    "doc": "docs",
-    "lib": "lib",
-    "test": "test"
-  },
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
-  },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/thrackle-io/Tron.git"
-  },
-  "author": "",
-  "license": "ISC",
-  "bugs": {
-    "url": "https://github.com/thrackle-io/Tron/issues"
-  },
-  "homepage": "https://github.com/thrackle-io/Tron#readme"
+  }
 }
--- a/src/application/AppManager.sol
+++ b/src/application/AppManager.sol
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.8.17;
 
-import "openzeppelin-contracts/contracts/access/AccessControlEnumerable.sol";
-import "openzeppelin-contracts/contracts/utils/introspection/ERC165Checker.sol";
+import "@openzeppelin/contracts/access/AccessControlEnumerable.sol";
+import "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
 import "src/data/Accounts.sol";
 import "src/data/IAccounts.sol";
 import "src/data/IAccessLevels.sol";
--- a/src/application/AppManager.sol
+++ b/src/application/AppManager.sol
@@ -3,23 +3,23 @@ pragma solidity 0.8.17;
 
 import "@openzeppelin/contracts/access/AccessControlEnumerable.sol";
 import "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
-import "src/data/Accounts.sol";
-import "src/data/IAccounts.sol";
-import "src/data/IAccessLevels.sol";
-import "src/data/AccessLevels.sol";
-import "src/data/IRiskScores.sol";
-import "src/data/RiskScores.sol";
-import "src/data/IGeneralTags.sol";
-import "src/data/GeneralTags.sol";
-import "src/data/IPauseRules.sol";
-import "src/data/PauseRules.sol";
-import "src/application/ProtocolApplicationHandler.sol";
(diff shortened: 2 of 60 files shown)


src/application/AppManager.sol
```diff
@@ -1,25 +1,25 @@
...
+import "../data/GeneralTags.sol";
+import "../data/IPauseRules.sol";
+import "../data/PauseRules.sol";
+import "../application/ProtocolApplicationHandler.sol";
+import "../economic/ruleProcessor/ActionEnum.sol";
+import {IAppLevelEvents} from "../interfaces/IEvents.sol";
+import "../application/IAppManagerUser.sol";
+import "../data/IDataModule.sol";
+import "../token/IAdminWithdrawalRuleCapable.sol";
+import "../token/ProtocolTokenCommon.sol";
```
ShaneDuncan602, line 22: Do all of these have to be relative paths? We have spent a lot of effort to switch them to absolute.

src/example/ApplicationERC721.sol
```diff
@@ -1,23 +1,32 @@
...
+ * that the one implementation you chose and its variables are enabled (not commented out).
  */
 
-contract ApplicationERC721 is ProtocolERC721 {
+contract ApplicationERC721 
+    is 
+    ProtocolERC721
+    ,AppAdministratorOrOwnerOnly
+    {
+
```
ShaneDuncan602: This formatting got weird

src/application/AppManager.sol
oscarsernarosero: This is a good discussion to have with @jjbailey-io

src/example/ApplicationERC721.sol
oscarsernarosero: I left it like that so it could be easy commented out. Would it be better if I just leave it in a single line? Or do you have anything else in mind?

src/application/AppManager.sol
jjbailey-io: I don't mind which way they go as long as they're consistent. Pick one or the other. If we go relative, that makes the NPM packaging easier.

src/example/ApplicationERC721.sol
```diff
@@ -102,16 +108,15 @@ contract ApplicationERC721 is ProtocolERC721 {
...
-    //     if (msg.sender != owner) revert OnlyOwnerCanMint();
-    //     uint256 tokenId = _tokenIdCounter.current();
-    //     _tokenIdCounter.increment();
-    //     _safeMint(to, tokenId);
-    // }
+      function safeMint(address to) public payable override whenNotPaused appAdministratorOrOwnerOnly(appManagerAddress) {
+        uint256 tokenId = _tokenIdCounter.current();
+        _tokenIdCounter.increment();
+        _safeMint(to, tokenId);
+    }
```
ShaneDuncan602: The access control was supposed to be added to ProtocolERC721 so it acted as a default for all implementing NFT's.

src/application/AppManager.sol
ShaneDuncan602: This is going to make the merge have a ton of conflicts. We have been changing them all to absolutes.

lib/openzeppelin-contracts
ShaneDuncan602, line 1: This worries me a little. I don't know why we needed to update. It scares me because we haven't tested any of the upgradeable implementations with these OZ versions.

src/application/AppManager.sol
oscarsernarosero: I can revert the commit where I migrated to relative. We could still do the package, but not as straight forward as we wanted since we would have to copy the files to a different repo. We have 3 options:

- Create the package from a separate repo where we copy the files from Rules Protocol, and we go from absolute to relative through a script or library every time we copy the files.
- Migrate to relative when we merge to Rules Protocol through a library or script (same as before, but using Rules Protocol repo as the package repo).
- Go relative all the way in Tron.

package.json
```diff
@@ -2,5 +2,27 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/thrackle-io/Tron.git"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/thrackle-io/Tron/issues"
+  },
+  "homepage": "https://github.com/thrackle-io/Tron#readme"
```
ShaneDuncan602: Is this supposed something that was unintentionally pushed?

src/application/AppManager.sol
oscarsernarosero: As agreed with the team. This will stay like this.

src/example/ApplicationERC721.sol
oscarsernarosero: Fixed

src/example/ApplicationERC721.sol
oscarsernarosero: Fixed

lib/openzeppelin-contracts
oscarsernarosero: Will be addressed by Johnathan.

test/ApplicationERC721Fuzz.t.sol
```diff
@@ -116,23 +116,29 @@ contract ApplicationERC721FuzzTest is DiamondTestUtil, RuleProcessorDiamondTestU
         address[] memory addressList = getUniqueAddresses(_addressIndex % ADDRESSES.length, 2);
         address randomUser = addressList[0];
         address randomUser2 = addressList[1];
-        vm.stopPrank();
-        vm.startPrank(randomUser);
+        // vm.stopPrank();
+        // vm.startPrank(randomUser);
```
TJ-Everett: we should remove these lines if they are no longer necessary in the test

test/ApplicationERC721Fuzz.t.sol
```diff
@@ -116,23 +116,29 @@ contract ApplicationERC721FuzzTest is DiamondTestUtil, RuleProcessorDiamondTestU
...
         vm.stopPrank();
         vm.startPrank(randomUser2);
-        /// Burn appAdministrator token
+        // Burn appAdministrator token
         applicationNFT.burn(0);
         ///Return to default admin account
-        vm.stopPrank();
-        vm.startPrank(randomUser);
+        // vm.stopPrank();
+        // vm.startPrank(randomUser);
```
TJ-Everett: same here

test/ApplicationERC721Fuzz.t.sol
```diff
@@ -215,18 +221,20 @@ contract ApplicationERC721FuzzTest is DiamondTestUtil, RuleProcessorDiamondTestU
...
         applicationNFT.safeMint(user1); /// Id 6
         applicationNFT.safeMint(user1); /// Id 7
         applicationNFT.safeMint(user1); /// Id 8
         applicationNFT.safeMint(user1); /// Id 9
         applicationNFT.safeMint(user1); /// Id 10
 
-        vm.stopPrank();
-        vm.startPrank(user2);
+        // vm.stopPrank();
+        // vm.startPrank(user2);
```
TJ-Everett: same here

TJ-Everett (approved):
Just those few test lines that are commented out, otherwise looks good. I appreciate the expanded explanation on the optional mint functions as well.

ShaneDuncan602 (approved):
Looks good.

ShaneDuncan602:
merged with external

PR #109, opened 2023-08-16
Audit final 2 - event emission

TJ-Everett:


oscarsernarosero (commented):
We don't need an event when proposing new App Manager Address? I would say we do need an event there.

TJ-Everett:
> We don't need an event when proposing new App Manager Address? I would say we do need an event there.

This is a good question and happy to drop an event in there if we need/want. gnosis safe UI has pending requests for their multi step approval process, would we want to mimic that? i.e. in our UI would we catch the event and then let the user know they have a pending proposed new address? @oscarsernarosero

ShaneDuncan602:
> We don't need an event when proposing new App Manager Address? I would say we do need an event there.

Since we are already adding events to this, I agree. Good idea, Oscar.

ShaneDuncan602:
Additionally, the ProtocolERC20 has the setAppManagerAddress function in it again. This function should be removed since the process goes through the propose/confirm functions in ProtocolTokenCommon. This was likely a merge issue.

ShaneDuncan602 (changes requested):
Just the removal of the setAppManagerAddress function and adding the propose event.

ShaneDuncan602 (approved):
Looks great!

ShaneDuncan602:
merged to external

PR #110, opened 2023-08-16
Audit Final 9 - Add missing empty array validation to rule creation functions.

### LINKED ISSUE #106: Audit Final 9 - Zero address validation
[TOB-THR-9](https://github.com/thrackle-io/rules-protocol/commit/578ac7f1a76016d36c90bfbf205716cd887902c8) - There are functions which take in array arguments that don't check if the arrays are of zero length. In the case of functions such as [addBalanceLimitRules](https://github.com/thrackle-io/rules-protocol/blob/578ac7f1a76016d36c90bfbf205716cd887902c8/src/economic/ruleStorage/TaggedRuleDataFacet.sol#L178C14-L213) it could result in the event being emitted and the index being incremented even though no rule was added. Also the if (_appManagerAddr == address(0)) revert ZeroAddress(); aren't really necessary in functions that have the appAdministratorOnly(_appManagerAddr) modifier since the modifier performs the check as well. We could have probably been a bit more specific in the TOB-THR-9 finding description since I see specific mention of array length zero checks is missing

ShaneDuncan602:
Went through all the rule storage facets and added empty array validation.

### DIFF
--- a/src/economic/ruleStorage/AppRuleDataFacet.sol
+++ b/src/economic/ruleStorage/AppRuleDataFacet.sol
@@ -160,6 +160,8 @@ contract AppRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents, II
     ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32) {
         /// Validation block
         if (_maxSize.length != _riskLevel.length + 1) revert InputArraysSizesNotValid();
+        // since all the arrays must have matching lengths, it is only necessary to check for one of them being empty.
+        if (_maxSize.length == 0) revert InvalidRuleInput();
         if (_riskLevel[_riskLevel.length - 1] > 99) revert RiskLevelCannotExceed99();
         for (uint256 i = 1; i < _riskLevel.length; ) {
             if (_riskLevel[i] <= _riskLevel[i - 1]) revert WrongArrayOrder();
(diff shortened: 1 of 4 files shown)


TJ-Everett (approved):
Nice work Shane.

src/economic/ruleStorage/AppRuleDataFacet.sol
```diff
@@ -160,6 +160,8 @@ contract AppRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents, II
     ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32) {
         /// Validation block
         if (_maxSize.length != _riskLevel.length + 1) revert InputArraysSizesNotValid();
+        // since all the arrays must have matching lengths, it is only necessary to check for one of them being empty.
+        if (_maxSize.length == 0) revert InvalidRuleInput();
```
oscarsernarosero, line 164: very efficient

src/economic/ruleStorage/AppRuleDataFacet.sol
```diff
@@ -233,6 +235,7 @@ contract AppRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents, II
      */
     function addAccountBalanceByRiskScore(address _appManagerAddr, uint8[] calldata _riskScores, uint48[] calldata _balanceLimits) external ruleAdministratorOnly(_appManagerAddr) returns (uint32) {
         if (_balanceLimits.length != _riskScores.length + 1) revert InputArraysSizesNotValid();
+        if (_balanceLimits.length == 0 || _riskScores.length == 0) revert InvalidRuleInput();
```
oscarsernarosero: I think you can apply same principal here. Since balanceLimits is always smaller than riskScores at this point. You can just check only on balanceLimits and that would be enough.

oscarsernarosero (approved):
Looks good. I think you can still improve one check tho.

src/economic/ruleStorage/AppRuleDataFacet.sol
ShaneDuncan602: changed

ShaneDuncan602:
merged to external

PR #116, opened 2023-08-18
External 58 versioning

### LINKED ISSUE #58: All Modules - Create version variables and associated functions
Create a string variable to store the version information. This must be a string in order to capture semantic versioning, ie 1.0.0. 

This should be added to:
1. All Handlers
2. AppManager
3. RuleStorageDiamond
4. RuleProcessorDiamond

oscarsernarosero:
Version fields added to handlers, AppManager, RuleStorage, RuleProcessor, Data contracts and Pricers. The strategy used is described [here](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/166133773/Tron+Versioning)


TJ-Everett (commented):
Should the versions be at 1.0.0 ? Since we will be cutting these changes over to rules then deploying in this sprint?

ShaneDuncan602:
I think this PR should point to external

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -15,7 +15,6 @@ import {IRuleProcessorErrors, ITagRuleErrors, IMaxTagLimitError} from "../../int
  * @notice  Implements Token Rules on Tagged Accounts.
  */
 contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, ITagRuleErrors, IMaxTagLimitError {
-    uint8 public constant VERSION = 1;
```
ShaneDuncan602, line 18: Good thinking removing these old ones.

src/economic/ruleProcessor/RuleProcessorDiamond.sol
```diff
@@ -53,6 +54,25 @@ contract RuleProcessorDiamond is ERC173Facet {
...
+    function updateVersion(string memory newVersion) external onlyOwner{
+        VersionStorage storage v = DiamondLib.versionStorage();
+        v.VERSION = newVersion;
+    }
+
+    /**
+    * @dev returns the version of the Rule Processor Diamond.
+    * @return string version.
+    */
+    function VERSION() external view returns(string memory){
```
ShaneDuncan602: Are we sure we want this version to be all caps? I usually only see that with enums. For instance, OZ has similar functions for name, owner, etc and they are all lower or camel.

oscarsernarosero:
> Should the versions be at 1.0.0 ? Since we will be cutting these changes over to rules then deploying in this sprint?

Good point. Im debating if I should do that in this ticket, or if I should leave it here as 0.0.6, and then create another ticket dedicated only for upgrading the version as we should do from here onwards.

src/economic/ruleStorage/RuleStorageDiamond.sol
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.17;
 
-import {RuleStorageDiamondLib as DiamondLib, DiamondCutStorage, FacetCut} from "./RuleStorageDiamondLib.sol";
+import {RuleStorageDiamondLib as DiamondLib, DiamondCutStorage, FacetCut, VersionStorage} from "./RuleStorageDiamondLib.sol";
 import {IRuleStorageDiamondEvents} from "../../interfaces/IEvents.sol";
+import {ERC173Facet} from "diamond-std/implementations/ERC173/ERC173Facet.sol";
```
ShaneDuncan602: I don't think we need this in here. ProtocolRawFacet contains 165 and 173 and is always added to the diamonds. That should make the 173 functions available. In [Mudgen's newest examples](https://github.com/mudgen/diamond-1-hardhat/commits?author=mudgen)  he defines the interfaces, IERC165 and IERC173, in the diamond but has the code for them in a facet. We should probably do the same. I think we need to remove 165 and 173 from any remaining facets and define the interfaces in the storage and processor diamonds.

src/economic/ruleStorage/RuleStorageDiamond.sol
```diff
@@ -20,7 +21,7 @@ struct RuleStorageDiamondArgs {
  * @dev main contract of the Rule diamond pattern. Mainly responsible
  * for storing the diamnond-pattern logic and binding together the different facets.
  */
-contract RuleStorageDiamond is IRuleStorageDiamondEvents {
+contract RuleStorageDiamond is IRuleStorageDiamondEvents, ERC173Facet {
```
ShaneDuncan602: see above

src/token/ProtocolHandlerCommon.sol
```diff
@@ -23,6 +23,7 @@ import "./IAdminWithdrawalRuleCapable.sol";
  */
 
 abstract contract ProtocolHandlerCommon is IAppManagerUser, IOwnershipErrors, IZeroAddressError, ITokenHandlerEvents, IAssetHandlerErrors, AppAdministratorOrOwnerOnly {
+    string public constant VERSION="0.0.6";
```
ShaneDuncan602: Great spot for this.

ShaneDuncan602 (changes requested):
Just some questions about implementation and capitolization.

src/economic/ruleProcessor/RuleProcessorDiamond.sol
oscarsernarosero: Since constants are conventionally named in all caps, and since these constants were made public in the application side, their getters will automatically be VERSION(), so I just tried to make it all the same. However, you bring up a good point which is that the version is not constant in the protocol side. So we can make VERSION constant private in the application side, and create a version() getter for those, and that way we can have all getters be lower case.

src/economic/ruleStorage/RuleStorageDiamond.sol
oscarsernarosero: I had to add it cuz it was complaining when I was trying to make the `updateVersion` function onlyOwner. Unless there is another way of doing it, and Im not aware of.

ShaneDuncan602: Will it work with just the interface defined?

ShaneDuncan602: I think that the version stuff should be in its own facet rather than in the diamond itself. We should probably keep the diamond as clean as possible. If you look at his [OwnershipFacet](https://github.com/mudgen/diamond-1-hardhat/blob/main/contracts/facets/OwnershipFacet.sol) you can see an example.

oscarsernarosero: great point. Will do.

ShaneDuncan602:
Can you add some testing to make sure that the ownership logic works as expected in the diamonds?

ShaneDuncan602:
I looked through the existing test for ownership transfer. It doesn't really test it. It calls the transferOwnership but there is no assertion that the new owner is truly the owner. It also seems to transfer ownership from superAdmin to superAdmin. All I think it needs is for the ownership to be transferred to a different address and then that address try to upgrade. These functions are in RuleProcessorDiamond.t.sol and RuleStorageDiamond.t.sol - testUpgrade().

oscarsernarosero:
> I looked through the existing test for ownership transfer. It doesn't really test it. It calls the transferOwnership but there is no assertion that the new owner is truly the owner. It also seems to transfer ownership from superAdmin to superAdmin. All I think it needs is for the ownership to be transferred to a different address and then that address try to upgrade. These functions are in RuleProcessorDiamond.t.sol and RuleStorageDiamond.t.sol - testUpgrade().

ohh ok I see what you're saying. You're right. I haven't tested the ownership transfer. Will do.

ShaneDuncan602 (approved):
Looks great.

PR #121, opened 2023-08-18
Audit Final 3 - Only allow registered handlers to check app rules

### LINKED ISSUE #117: Audit Final 3 - AppManager CheckAllRules permissions
For the [TOB-THR-3 commit](https://github.com/thrackle-io/rules-protocol/commit/a4a91d2ba961e37eaab8b19a60d92ab55f37ef56) it seems the fix only partially resolved the access control issue since the AppManager contract is missing access control on the checkApplicationRules function. So although you have it so only the token can call the checkAllRules function of the token handler, and only the AppManager can call the checkApplicationRules of the applicationHandler, anyone can call checkApplicationRules on the AppManager and bypass the latter access control check.
We're still thinking about the best way to address this, ideally the checkAllRules and checkApplicationRules functions should not make state changes and then there wouldn't be a need for access control, but it seems like this would require the system to be reworked quite a bit. Another option is to implement a registry of token handler contracts so you can ensure the caller of the checkApplicationRules function can only be a valid token handler.
Additionally, the current way of access control through the onlyOwner modifier might open up an attack vector if the token contract is malicious (or upgradeable), since it would also be able to call all the setter functions that have the appAdministratorOrOwnerOnly modifier. This seems unlikely but I wanted to put it out there as an edge case. You could resolve this by having a new modifier instead of onlyOwner for the checkAllRules access control (e

ShaneDuncan602:
Some tests had to be removed. This doesn't affect test coverage but it reduces the layers of testing.


test/ApplicationERC20Handler.t.sol
```diff
@@ -35,8 +35,15 @@ contract ApplicationERC20HandlerTest is TestCommon {
         vm.startPrank(superAdmin);
         setUpProtocolAndAppManagerAndTokens();
         switchToAppAdministrator();
+        /// NOTE: this set up logic must be different because the handler must be owned by appAdministrator so it mayt be called directly. It still
```
TJ-Everett: small typo "so it may" remove the t.

TJ-Everett (approved):
Just that tiny typo. Otherwise looks good.

TJ-Everett (commented):
I like these changes. Question: would we (the ui) want to know if you are skipping one of the addresses you are registering for tokens? 
Should we change the AddressAlreadyRegistered to an event instead of error and emit when we skip so devs could still know for sure?

ShaneDuncan602:
> I like these changes. Question: would we (the ui) want to know if you are skipping one of the addresses you are registering for tokens? Should we change the AddressAlreadyRegistered to an event instead of error and emit when we skip so devs could still know for sure?

That would be a good enhancement for the future.

src/application/AppManager.sol
```diff
@@ -593,27 +594,52 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
         if (_tokenAddress == address(0)) revert ZeroAddress();
+        bool skip;
         tokenToAddress[_token] = _tokenAddress;
         addressToToken[_tokenAddress] = _token;
         for (uint256 i = 0; i < tokenList.length; ) {
-            if (tokenList[i] == _tokenAddress) revert AddressAlreadyRegistered();
+            if (tokenList[i] == _tokenAddress) {
+                skip = true;
+                break;
+            }
```
oscarsernarosero, line 632: I think it's better to check if the token has been registered like this:
```
if(keccak256(abi.encodePacked(addressToToken[_tokenAddress])) != keccak256(abi.encodePacked(""))
    revert AddressAlreadyRegistered();
```
This way we don't have to loop through the array.

src/application/AppManager.sol
```diff
@@ -651,28 +682,30 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+                for (uint256 i = 0; i < _addressArray.length; ) {
+                    if (_addressArray[i] == _address) {
+                        _addressArray[i] = _addressArray[_addressArray.length - 1];
+                        _addressArray.pop();
+                        _removed = true;
+                        break;
+                    }
+                    unchecked {
+                        ++i;
+                    }
```
oscarsernarosero, line 704: For right now, this is a legit implementation, but I think we can try to use something similar to what most ERC721 enumerable implementations use, which is a mapping addressToIndex. This way we always know the index of an address in the array, and we wouldn't have to loop through it to know where it is or to handle it.

src/application/AppManager.sol
oscarsernarosero: Also, this means that once an address has been registered, it won't be possible to change its name, so it would be good to have an updateTokenName function just in case.

oscarsernarosero (commented):
I think we can improve the gas efficiency of the code by checking if an address has been registered in a different way, and also the deregistering process could be more efficient if we support our tokenList array with an index mapping.

ShaneDuncan602:
merged with external

PR #123, opened 2023-08-18
Audit final 24 - AdminWithdrawal Circumvention

ShaneDuncan602:


docs/userGuides/rules/RULE-GUIDE.md
```diff
@@ -17,7 +17,7 @@
...
-| application admin Withdrawal                          | In some instances, an application developer may wish to assure users they will not pillage the ecosystem and then abandon it. For instance, an application developer may wish to promise on-chain that they will not withdraw more than 1000 tokens worth of value from the system until users have collectively withdrawn more than 100,000 tokens.                                                                   |
+| Application Admin Withdrawal                          | In some instances, an application developer may wish to assure users they will not pillage the ecosystem and then abandon it. For instance, an application developer may wish to promise on-chain that they will not withdraw more than 1000 tokens worth of value from the system until users have collectively withdrawn more than 100,000 tokens. NOTE: This rule is dependent upon trust of the "Super Admin" account as this account has permissions to remove the "App Admin" role thus allowing an admin to be released from the rule restriction.                                                                  |
```
TJ-Everett, line 20: This is great wording. Since this was always a "promise on chain"

src/application/AppManager.sol
```diff
@@ -139,7 +124,7 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
      * @dev Add an array of accounts to the app administrator role. Restricted to admins.
      * @param _accounts address array to be added
      */
-    function addMultipleAppAdministrator(address[] memory _accounts) external onlySuperAdmin {
+    function addMultipleAppAdministrator(address[] memory _accounts) external onlyRole(SUPER_ADMIN_ROLE) {
```
TJ-Everett, line 127: This is way cleaner - great thinking on this

TJ-Everett (approved):
This looks great. Great work Shane

src/application/AppManager.sol
oscarsernarosero: Agree. Great idea!

oscarsernarosero (approved):
This is great Shane. For what it's worth, I still think we should just build our own vesting contract for devs to do this instead writing a rule.

ShaneDuncan602:
merged to external

PR #125, opened 2023-08-18
Audit 13 - Reverted the addTransferCounterRule event back where it wa…

ShaneDuncan602:
…s pre-audit


TJ-Everett:
Did we want to also move the event on the diamond cut in this as well? Or are we okay with the event inside the loop for that one?

ShaneDuncan602:
> 

Yep, you're right. I had to read Elvis's reply a few more times to understand it. Good catch.

TJ-Everett (approved):
approved

oscarsernarosero (approved):
Im glad we reverted this change. The indexer can take care of this rule with a single event as it is rn.

ShaneDuncan602:
merged to external

PR #127, opened 2023-08-21
Audit 26 - Added custom error to handle fees over 100 percent

### LINKED ISSUE #81: Audit - Token transfers can revert due to transfer fees exceeding the transfer amount(26)
See issue 26 in [ToB Final Report](https://drive.google.com/file/d/1JmY1U7RlM43-t1Dz_wcGBicltCMxy33A/view?usp=sharing)

The total fees a user needs to pay when initiating a token transfer can exceed 100% which will result in all transfers reverting for that user.

Fix Plan(per product)
- When the fees and discounts have been assessed and they amount to more than 100% of the total transaction, the transfer should revert with a custom error that identifies it as a problem. The custom error should be: FeesAreGreaterThanTransactionAmount(address). This will give users the ability to determine how to fix it by retrieving the account's tags and checking the fees on the token.

ShaneDuncan602:
This closes #81

### DIFF
--- a/src/token/ProtocolERC20Handler.sol
+++ b/src/token/ProtocolERC20Handler.sol
@@ -289,9 +289,10 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
     function getApplicableFees(address _from, uint256 _balanceFrom) public view returns (address[] memory feeCollectorAccounts, int24[] memory feePercentages) {
         Fees.Fee memory fee;
         bytes32[] memory _fromTags = appManager.getAllTags(_from);
+        int24 totalFeePercent;
+        uint24 discount;
         if (_fromTags.length != 0 && !appManager.isAppAdministrator(_from)) {
             uint feeCount;
-            uint24 discount;
             uint discountCount;
             // size the dynamic arrays by maximum possible fees
             feeCollectorAccounts = new address[](_fromTags.length);
@@ -308,6 +309,8 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
                     } else {
                         feePercentages[feeCount] = fee.feePercentage;
                         feeCollectorAccounts[feeCount] = fee.feeCollectorAccount;
+                        // add to the total fee percentage
+                        totalFeePercent += fee.feePercentage;
                         unchecked {
                             ++feeCount;
                         }
@@ -335,6 +338,10 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
                     }
                 }
             }
+        }
+        // if the total fees - discounts is greater than 100 percent, revert
+        if (totalFeePercent - int24(discount) > 10000) {
+            revert FeesAreGreaterThanTransactionAmount(_from);
         }
         return (feeCollectorAccounts, feePercentages);
     }
--- a/test/ApplicationERC20.t.sol
+++ b/test/ApplicationERC20.t.sol
@@ -792,6 +792,53 @@ contract ApplicationERC20Test is TestCommon {
         assertEq(applicationCoin.balanceOf(targetAccount2), 11 * (10 ** 18)); // treasury remains the same
     }
 
+    ///Test transferring coins with fees and discounts where the discounts are greater than the fees
+    function testTransactionFeeTableDiscountsCoin() public {
+        applicationCoin.transfer(user4, 100000 * (10 ** 18));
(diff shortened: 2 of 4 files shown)


src/token/ProtocolERC20Handler.sol
```diff
@@ -336,6 +339,10 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
                 }
             }
         }
+        // if the total fees - discounts is greater than 100 percent, revert
+        if (totalFeePercent - int24(discount) > 10000) {
```
oscarsernarosero, line 343: is there any chances that the discount total is greater than the fees total? If that is a possibility, then we have to check for that scenario before carrying out the substraction since this will cause an underflow.

test/ApplicationERC20.t.sol
```diff
@@ -891,6 +891,70 @@ contract ApplicationERC20Test is TestCommon {
...
+        // make sure standard fee works(other fee will also be assessed)
+        bytes4 selector = bytes4(keccak256("FeesAreGreaterThanTransactionAmount(address)"));
+        vm.expectRevert(abi.encodeWithSelector(selector, user4));
+        applicationCoin.transfer(user3, 200 * (10 ** 18));
+        // make sure nothing changed
+        assertEq(applicationCoin.balanceOf(user4), 99800 * (10 ** 18));
+        assertEq(applicationCoin.balanceOf(user3), 97 * (10 ** 18));
+        assertEq(applicationCoin.balanceOf(targetAccount), 6 * (10 ** 18)); // previous 3 + current 3
+        assertEq(applicationCoin.balanceOf(targetAccount2), 97 * (10 ** 18)); // current 7
+    }
```
oscarsernarosero, line 1003: A test for discounts being greater than fees must be included.

oscarsernarosero (changes requested):
An unexplored scenario here is discounts exceeding fees which might cause unexpected behaviors.

src/token/ProtocolERC20Handler.sol
ShaneDuncan602: Shouldn't be a problem since they are signed integers.

TJ-Everett: I agree it would not be a problem on this check specifically if the discount is greater than the fees, where it would be an underflow is when subtracted from the total token amount in the transfer (I think this is what concerns Oscar as well) where you would then, essentially, be owned tokens in the transfer from since you would be adding a negative to the transfer amount. 
Oscar is that a correct summary?

oscarsernarosero: ah ok. The only problem with that is the gas cost. I went ahead and tested the gas in remix like this:
```

    function signedSub(int a, uint24 b) external pure returns (int z){
        return a - int24(b);
    }

    function unsignedSub(uint a, uint24 b) external pure returns (uint z){
        if(a > b)
            return a - b;
        else
            return 0;
    }
```
    
The input was 1000 for A and 2000 for B. The gas cost for using unsigned integers was 814 and 1105 for using the signed version (35% increase), and that's not even counting the variable creation. I would try to use unsigned integers when possible.

ShaneDuncan602: It always checks to ensure that the fee is larger than the discount slice. I can add a test for this scenario though just to be certain.

oscarsernarosero: yes. I think the most important factor here is the test to make sure there is no issues in that scenario. And if the code always checks that the fees are greater than discounts, then there is no need to use the more expensive signed integer over the unsigned integer.

TJ-Everett (approved):
This looks great Shane. Approved

test/ApplicationERC20.t.sol
```diff
@@ -792,6 +792,53 @@ contract ApplicationERC20Test is TestCommon {
...
+        applicationAppManager.addGeneralTag(user4, "fee1"); ///add tag
+        vm.stopPrank();
+        vm.startPrank(user4);
+        // discounts are greater than fees so it should put fees to 0
+        applicationCoin.transfer(user3, 100 * (10 ** 18));
+        assertEq(applicationCoin.balanceOf(user4), 99900 * (10 ** 18));
+        assertEq(applicationCoin.balanceOf(user3), 100 * (10 ** 18));
+        assertEq(applicationCoin.balanceOf(targetAccount), 0 * (10 ** 18));
+    }
+
```
oscarsernarosero, line 841: Looks great

oscarsernarosero (approved):
Looks great Shane. Now there's no room for doubt.

PR #130, opened 2023-08-23
Se 173 code quality open zeppelin consistency

ShaneDuncan602:
Upgraded both OZ submodules to v4.9. This required a minor override change.

### DIFF
--- a/src/token/ProtocolERC721.sol
+++ b/src/token/ProtocolERC721.sol
@@ -18,7 +18,7 @@ import "../token/IProtocolERC721Handler.sol";
  * @notice This is the base contract for all protocol ERC721s
  */
 
-contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon ,AppAdministratorOrOwnerOnly {
+contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon, AppAdministratorOrOwnerOnly {
     using Counters for Counters.Counter;
     address public handlerAddress;
     IProtocolERC721Handler handler;
@@ -94,7 +94,7 @@ contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, P
      * Function is payable for child contracts to override with priced mint function.
      * @param to Address of recipient
      */
-    function safeMint(address to) public payable virtual whenNotPaused appAdministratorOrOwnerOnly(appManagerAddress){
+    function safeMint(address to) public payable virtual whenNotPaused appAdministratorOrOwnerOnly(appManagerAddress) {
         uint256 tokenId = _tokenIdCounter.current();
         _tokenIdCounter.increment();
         _safeMint(to, tokenId);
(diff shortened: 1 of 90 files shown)


lib/openzeppelin-contracts-upgradeable
oscarsernarosero, line 1: why do we have changes here?

src/token/ProtocolERC721.sol
```diff
@@ -141,7 +141,7 @@ contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, P
      *
      * This function call must use less than 30 000 gas.
      */
-    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable) returns (bool) {
+    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable, ERC721URIStorage) returns (bool) {
         return super.supportsInterface(interfaceId);
```
oscarsernarosero: I would honestly try to be safe and make sure that we are not telling the world we are not supporting one of these interfaces (not sure what route the contract inheritance will take) by doing something like:

```
return ERC721Enumerable.supportsInterface(interfaceId) || ERC721URIStorage.supportsInterface(interfaceId) 
```

You'll have to double check, but I think ERC721Enumerable and/or ERC721URIStorage implementations already have ERC721 in their supportsInterface function, so no need to put that one in the return statement.

oscarsernarosero (changes requested):
Everything looks fantastic. I would just feel more confident if we add a test where we make sure that the supportsInterface function in the ERC721 will return true for both ERC721Enumerable and ERC721Storage. If not, I left a _possible_ implementation in the comments.

lib/openzeppelin-contracts-upgradeable
jjbailey-io: I found only updating the .gitmodules file does not cause the underlying OZ libraries to update to the correct version. You must also check out the appropriate commit in the submodule and commit the change.

oscarsernarosero: I think you have to delete the submodules manually, and then do a 'forge build' to make sure it has the appropriate versions in it. I ask this because we didn't intend to upgrade this particular submodule, so we shouldn't have changes here for what I know.

jjbailey-io: This module is currently at v4.9.2 in the trunk. If we upgrade the other, it'll be at v4.9.3 and they'll be at different versions.

src/token/ProtocolERC721.sol
```diff
@@ -141,8 +141,8 @@ contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, P
      *
      * This function call must use less than 30 000 gas.
      */
-    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable) returns (bool) {
-        return super.supportsInterface(interfaceId);
+    function supportsInterface(bytes4 interfaceId) public view override(ERC721, ERC721Enumerable, ERC721URIStorage) returns (bool) {
+        return ERC721Enumerable.supportsInterface(interfaceId) || ERC721URIStorage.supportsInterface(interfaceId) || ERC721.supportsInterface(interfaceId) || super.supportsInterface(interfaceId);
```
oscarsernarosero: I think you only need ERC721Enumerable.supportsInterface(interfaceId) and ERC721URIStorage.supportsInterface(interfaceId) since ERC721.supportsInterface(interfaceId) is already inside those, and super.supportsInterface(interfaceId) is the same as saying ERC721URIStorage.supportsInterface(interfaceId)

oscarsernarosero (approved):
Looks great. The supportsInterface will work just fine and as efficient as if you only have the first 2 checks due to the short-circuit ORs. Im approving.

PR #131, opened 2023-08-24
Se 175 code quality change registrations to use mappings

oscarsernarosero:
### Main Changes

The following arrays were migrated to use a new mechanism to handle **_isInList_, _addToList_, and _removeFromList_** function:

- ammList
- tokenList
- treasuryList
- tagRecords

The new mechanism is built upon 2 mappings: 

- `mapping(address => uint) addressToIndex`
- `mapping(address => bool) isAddressRegistered`

#### Gas savings

These mappings **reduce greatly** the cost of reading if an address is present in an array such as in functions `isTreasury`, `isRegisteredAMM`, and `hasTag` among others, and **reduces significantly** the cost of removing from these lists as well. 

The savings grow compared to the old method the larger the lists are, and are not so great when the lists are small. However, **these savings do come on a price which is paid when the registration** of an address is made to these lists, although these are paid by devs and not by users.

### Additional change

Another note worthy change is the addition of the **revert** clause when the **removeTag** function is called **but the tag doesn't exist** for such account.


src/application/AppManager.sol
```diff
@@ -667,46 +665,47 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
-                        _addressArray.pop();
-                        _removed = true;
-                        break;
-                    }
-                    unchecked {
-                        ++i;
-                    }
-                }
+    function _removeAddressWithMapping(
+        address[] storage  _addressArray,
```
ShaneDuncan602, line 696: What if the addressArray is empty? We had to put specific logic in here for that condition. If you try to pop an empty array it will revert, I think.
You have a check if the address is registered though. It's in the calling function. So that should cover it. Disregard.

test/ApplicationAppManager.t.sol
```diff
@@ -713,54 +713,186 @@ contract ApplicationAppManagerTest is TestCommon {
...
+        assertEq(address(applicationCoinA), applicationAppManager.getTokenAddress("CoinA"));
+
+        applicationAppManager.registerToken("CoinB", address(applicationCoinB));
+        assertEq(address(applicationCoinB), applicationAppManager.getTokenAddress("CoinB"));
+
+        applicationAppManager.registerToken("CoinC", address(applicationCoinC));
+        assertEq(address(applicationCoinC), applicationAppManager.getTokenAddress("CoinC"));
+    
+        // test updating the token's name
+        applicationAppManager.registerToken("FRANCISCOSTEIN", address(applicationCoin));
```
ShaneDuncan602, line 859: Colombian Frankenstein...I love it.

ShaneDuncan602 (approved):
Great job. I wish we could have gotten rid of the arrays completely but it's just not practical. This is much, much more efficient though.

PR #132, opened 2023-08-24
added to_relative script to installation of the package

oscarsernarosero:
- Minor fixes in the `to_relative.sh` script
- `to_relative.sh` script now runs as part of the installation process for the package to make sure current issue won't happen again.
- Even though the issue would be corrected at installation time in the clients machine, the script was run to fix current issue, so we don't have the wrong import style in Tron.
- The package.json was updated to now point to version 1.0.1 so it is ready to publish.


ShaneDuncan602 (approved):
Great job. Really smart fix for this.

PR #136, opened 2023-08-28
Changed pre image of storage slots to be unknown. Made tests and diam…

ShaneDuncan602:
…ond-std libray compatible with solidity 8.17 and up

### DIFF
--- a/lib/diamond-std/core/DiamondCut/DiamondCutLib.sol
+++ b/lib/diamond-std/core/DiamondCut/DiamondCutLib.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.17;
+pragma solidity ^0.8.17;
 
 import "./FacetCut.sol";
 
@@ -39,8 +39,7 @@ struct DiamondCutStorage {
 }
 
 library DiamondCutLib {
-    bytes32 constant DIAMOND_CUT_STORAGE_POSITION = keccak256("diamond-cut.storage");
-
+    bytes32 constant DIAMOND_CUT_STORAGE_POSITION = bytes32(uint256(keccak256("diamond-cut.storage")) - 1);
     function s() internal pure returns (DiamondCutStorage storage ds) {
         bytes32 position = DIAMOND_CUT_STORAGE_POSITION;
         assembly {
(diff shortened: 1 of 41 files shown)


TJ-Everett (approved):
This looks great Shane. Well done. 

Questions though: Is this a breaking change for Doom at all (this might be answered by Oscar)? I don't know if Doom even calls the storage location for rules at all or just sends the params to the contracts. 
Second, what was the need for updating to ^0.8.17? Or was this just best practice?

lib/diamond-std/core/DiamondCut/DiamondCutLib.sol
```diff
@@ -39,8 +39,7 @@ struct DiamondCutStorage {
 }
 
 library DiamondCutLib {
-    bytes32 constant DIAMOND_CUT_STORAGE_POSITION = keccak256("diamond-cut.storage");
-
+    bytes32 constant DIAMOND_CUT_STORAGE_POSITION = bytes32(uint256(keccak256("diamond-cut.storage")) - 1);
```
oscarsernarosero, line 42: So the only storage we changed was the `diamond-cut.storage` because it is the same in all diamonds out there? I think this makes sense, but if we are bothering ourselves to do this, maybe we can change `diamond-cut.storage` for something like `rules-protocol.diamond-cut.storage` since current exact solution might overlap with future diamond implementers that are advised with the same solution? I honestly don't fully understand the goal of this, so feel free to disregard this if you think this does not apply.

oscarsernarosero (approved):
Looks good. I only have one quessugestion.

oscarsernarosero:
> Is this a breaking change for Doom at all

@TJ-Everett  It is not. The diamond makes it transparent for any user including DOOM.

lib/diamond-std/core/DiamondCut/DiamondCutLib.sol
ShaneDuncan602: This was something I don't think was really necessary but ToB would not drop it. It's not that the diamond-cut.storage constant might overlap with other diamonds. It's because the pre image was known for all of the diamond storage. [EIP-1967](https://eips.ethereum.org/EIPS/eip-1967#rationale) explains why it is a good idea, but they acknowledged there are no known attack vectors out there. This "minus 1" technique was applied to all of our diamond storage positions. You might also want to check the audit report for their explanation. I don't think it will ever be a problem but it will take a bullet out of their gun in future audits.

jjbailey-io:
Closed.

PR #137, opened 2023-08-29
SE 187 access module global rules pause rules toggle on off

TJ-Everett:


src/application/AppManager.sol
```diff
@@ -391,20 +391,30 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
 
     /**
      * @dev Remove a pause rule. Restricted to Application Administrators
+     * @notice If no pause rules exist after removal bool is set to false in handler and pause rules will not be checked until new rule is added. 
      * @param _pauseStart Beginning of the pause window
      * @param _pauseStop End of the pause window
      */
     function removePauseRule(uint256 _pauseStart, uint256 _pauseStop) external onlyRole(RULE_ADMIN_ROLE) {
         pauseRules.removePauseRule(_pauseStart, _pauseStop);
+        /// retieve pause rule array
```
oscarsernarosero: misspelling

src/application/AppManager.sol
```diff
@@ -391,20 +391,30 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
      */
     function removePauseRule(uint256 _pauseStart, uint256 _pauseStop) external onlyRole(RULE_ADMIN_ROLE) {
         pauseRules.removePauseRule(_pauseStart, _pauseStop);
+        /// retieve pause rule array 
+        PauseRule[] memory pauseRule = pauseRules.getPauseRules();
+        /// if length is 0 no pause rules exist
+        if (pauseRule.length == 0){
+            /// set handler bool to false to save gas and prevent pause rule checks when non exist
+            applicationHandler.activatePauseRule(false);
+        }
```
oscarsernarosero: This over here is better to do in the data contract for gas reasons. You can create a function in PauseRules data contract that will return something like `return pauseRules.length == 0;`. It would have the same effect, but returning a boolean instead of returning an array which is really expensive.

src/application/ProtocolApplicationHandler.sol
```diff
@@ -305,6 +307,29 @@ contract ProtocolApplicationHandler is Ownable, RuleAdministratorOnly, IApplicat
...
+     * @notice This function does not use ruleAdminisitratorOnly modifier so adding pause rules via app manager will activate pause rule check automatically. 
+     * @param _on boolean representing if a rule must be checked or not.
+     */
+
+    function activatePauseRule(bool _on) external {
+        if (msg.sender == address(appManagerAddress)) {
+            pauseRuleActive = _on;
+        } else if (!appManager.isRuleAdministrator(msg.sender)){
+        pauseRuleActive = _on;
+        }
```
oscarsernarosero: This can be improved by making an OR conditional: `if (msg.sender == address(appManagerAddress) || !appManager.isRuleAdministrator(msg.sender))`

oscarsernarosero (changes requested):
Looks good. I do see some opportunities for improvements, and I would also add some tests to prove that current implementation works as expected.

src/application/ProtocolApplicationHandler.sol
oscarsernarosero: not sure why there is negation in the second if tho.

src/data/PauseRules.sol
```diff
@@ -95,4 +95,17 @@ contract PauseRules is IPauseRules, DataModule {
...
+     * @notice return true if pause rules is empty and return false if array contains rules 
+     * @return true if empty 
+     */
+    function isPauseRulesEmpty() external view virtual onlyOwner returns(bool) {
+        if (pauseRules.length == 0){
+            return true; 
+        } else {
+            return false; 
+        }
+    }
```
oscarsernarosero: This works perfectly fine. I just think we can avoid 4 lines of code if you simply return the result of the conditional:

```
function isPauseRulesEmpty() external view virtual onlyOwner returns(bool) {
        return pauseRules.length == 0;
        }
```
Im being a bit of a stickler here since one of the audit observations was that sometimes we had unnecessarily long chunks of code which makes it less readable.

src/application/ProtocolApplicationHandler.sol
```diff
@@ -305,6 +307,27 @@ contract ProtocolApplicationHandler is Ownable, RuleAdministratorOnly, IApplicat
...
 
+    /**
+     * @dev enable/disable rule. Disabling a rule will save gas on transfer transactions.
+     * This function does not use ruleAdministratorOnly modifier, the function checks if the caller is the appManager contract or if rule admin. 
+     * @notice This function does not use ruleAdminisitratorOnly modifier so adding pause rules via app manager will activate pause rule check automatically. 
+     * @param _on boolean representing if a rule must be checked or not.
+     */
+
+    function activatePauseRule(bool _on) external {
+        if (msg.sender == address(appManagerAddress) || appManager.isRuleAdministrator(msg.sender)) {
```
oscarsernarosero: ok. I had to look at the code, but I think it's better if this function is an onlyOwner fucntion since `appManagerAddress` is the owner, but you will need to force the devs to call this function through the AppManager as we do in other functions.

oscarsernarosero (changes requested):

oscarsernarosero (approved):
Looks perfect!

docs/userGuides/rules/RULE-GUIDE.md
```diff
@@ -27,7 +27,7 @@
...
-| Pause Window                                   | At times, it may be necessary to disable transactions on a particular token while maintenance functions or other activities are performed on a supporting application such as an application server or dashboard. A maintenance window can be designated by application admin accounts to suspend all application token activity during a specified time period. This would suspend all transactions for a given time period. |
+| Pause Window                                   | At times, it may be necessary to disable transactions on a particular token while maintenance functions or other activities are performed on a supporting application such as an application server or dashboard. A maintenance window can be designated by application admin accounts to suspend all application token activity during a specified time period. This would suspend all transactions for a given time period. Adding a pause rule in the appManager will actiavte the pauseRule check in the appHandler. Removing all pause rules will also deactivate pause rule checks in the appHandler.|
```
ShaneDuncan602: actiavte

ShaneDuncan602: You might also add something about the ability to toggle off the pause checks for efficiency as well as for unpausing without the need for pause rule deletion prior to the window closing.

src/data/PauseRules.sol
```diff
@@ -95,4 +95,13 @@ contract PauseRules is IPauseRules, DataModule {
...
     }
+
+    /**
+     * @dev Return a bool for if the PauseRule array is empty
+     * @notice return true if pause rules is empty and return false if array contains rules 
+     * @return true if empty 
+     */
+    function isPauseRulesEmpty() external view virtual onlyOwner returns(bool) {
+        return pauseRules.length == 0;
+    }
```
ShaneDuncan602, line 106: Does this need the onlyOwner?

ShaneDuncan602 (changes requested):
Just a minor typo and couple suggestions.

src/data/PauseRules.sol
TJ-Everett: My thought was that we might not want users to be able to call this and see if there is no upcoming pause rules, but otherwise no.

docs/userGuides/rules/RULE-GUIDE.md
TJ-Everett: This lead me to realize that I missed a function to do just this. Since the activate function was onlyOwner that meant only the appManager can call the activate function to skip/enable the check. So I added that function and some testing to the most recent commit to address the additional description and the functions match that.

src/data/PauseRules.sol
ShaneDuncan602: Good thinking.

ShaneDuncan602 (approved):
Looks good.

PR #138, opened 2023-08-29
Se 258 create automatic processes for version upgrades

oscarsernarosero:
- Added upgrade_version.sh where it upgrades the versions in `src`, `test` and `package.json`.
- Modified tests for version upgrade in the diamonds to avoid the script messing those up.

### DIFF
--- a/test/RuleProcessorDiamond.t.sol
+++ b/test/RuleProcessorDiamond.t.sol
@@ -120,10 +120,10 @@ contract RuleProcessorDiamondTest is Test, RuleProcessorDiamondTestUtil {
         vm.stopPrank();
         vm.startPrank(superAdmin);
         // update version
-        VersionFacet(address(ruleProcessor)).updateVersion("1,0,1"); // commas are used here to avoid upgrade_version-script replacements
+        VersionFacet(address(ruleProcessor)).updateVersion("1,0,0"); // commas are used here to avoid upgrade_version-script replacements
         string memory version = VersionFacet(address(ruleProcessor)).version();
         console.log(version);
-        assertEq(version, "1,0,1");
+        assertEq(version, "1,0,0");
         // update version again
         VersionFacet(address(ruleProcessor)).updateVersion("2.2.2");// upgrade_version script will replace this version
         version = VersionFacet(address(ruleProcessor)).version();
--- a/test/RuleProcessorDiamond.t.sol
+++ b/test/RuleProcessorDiamond.t.sol
@@ -120,12 +120,12 @@ contract RuleProcessorDiamondTest is Test, RuleProcessorDiamondTestUtil {
         vm.stopPrank();
         vm.startPrank(superAdmin);
         // update version
-        VersionFacet(address(ruleProcessor)).updateVersion("1.0.1");
+        VersionFacet(address(ruleProcessor)).updateVersion("test_version"); // this is done to avoid upgrade_version script replace this version
         string memory version = VersionFacet(address(ruleProcessor)).version();
         console.log(version);
-        assertEq(version, "1.0.1");
+        assertEq(version, "test_version");
         // update version again
-        VersionFacet(address(ruleProcessor)).updateVersion("2.2.2");
+        VersionFacet(address(ruleProcessor)).updateVersion("2.2.2");// upgrade_version script will replace this version
         version = VersionFacet(address(ruleProcessor)).version();
         console.log(version);
         assertEq(version, "2.2.2");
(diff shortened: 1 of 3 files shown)


test/RuleProcessorDiamond.t.sol
```diff
@@ -120,20 +120,20 @@ contract RuleProcessorDiamondTest is Test, RuleProcessorDiamondTestUtil {
...
         // update version
-        VersionFacet(address(ruleProcessor)).updateVersion("1.0.1");
+        VersionFacet(address(ruleProcessor)).updateVersion("1,0,1"); // commas are used here to avoid upgrade_version-script replacements
         string memory version = VersionFacet(address(ruleProcessor)).version();
         console.log(version);
-        assertEq(version, "1.0.1");
+        assertEq(version, "1,0,1");
         // update version again
-        VersionFacet(address(ruleProcessor)).updateVersion("2.2.2");
+        VersionFacet(address(ruleProcessor)).updateVersion("2.2.2");// upgrade_version script will replace this version
```
ShaneDuncan602, line 128: Wouldn't this one also need the commas?

test/RuleProcessorDiamond.t.sol
```diff
@@ -120,20 +120,20 @@ contract RuleProcessorDiamondTest is Test, RuleProcessorDiamondTestUtil {
         vm.stopPrank();
         vm.startPrank(superAdmin);
         // update version
-        VersionFacet(address(ruleProcessor)).updateVersion("1.0.1");
+        VersionFacet(address(ruleProcessor)).updateVersion("1,0,1"); // commas are used here to avoid upgrade_version-script replacements
```
ShaneDuncan602: This is a good solution to this problem. Good thinking.

test/RuleStorageDiamond.t.sol
```diff
@@ -102,20 +102,20 @@ contract RuleStorageDiamondTest is Test, RuleStorageDiamondTestUtil {
...
         // update version
-        VersionFacet(address(ruleStorageDiamond)).updateVersion("1.0.1");
+        VersionFacet(address(ruleStorageDiamond)).updateVersion("1,0,1"); // commas are used here to avoid upgrade_version-script replacements
         string memory version = VersionFacet(address(ruleStorageDiamond)).version();
         console.log(version);
-        assertEq(version, "1.0.1");
+        assertEq(version, "1,0,1");
         // update version again
-        VersionFacet(address(ruleStorageDiamond)).updateVersion("2.2.2");
+        VersionFacet(address(ruleStorageDiamond)).updateVersion("2.2.2"); // upgrade_version script will replace this version
```
ShaneDuncan602, line 110: Commas?

upgrade_version.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+################################################################################################################
+# Script designed specifically for Tron to change versions in the src and test directories as well as in the   #
+# package.json file.                                                                                           #
+#                                                                                                              #
+# Usage:                                                                                                       #
+#           ./upgrade_version.sh <MAJOR.MINOR.MICRO>                                                           #
+#                                                                                                              #
+# Notice that there are no quote marks in the version.                                                         #
```
ShaneDuncan602: quote should be quotation

upgrade_version.sh
```diff
@@ -0,0 +1,64 @@
...
+################################################################################################################
+
+# since all the files in source will have the same declaration pattern, we can use it to accurately change versions
+replace_version_in_src() {
+    local file="$1"
+    local new_version="$2"
+    
+    sed -i "" "s/\(string[[:space:]]*private[[:space:]]*constant[[:space:]]*VERSION[[:space:]]*=[[:space:]]*\"\)[^\"]*\(\"\)/\1$new_version\2/" "$file"
```
ShaneDuncan602, line 18: Are we always going to keep the versions in sync? I thought we would get to a point where the diamond had a version, the handlers had a version, etc?

ShaneDuncan602 (commented):
I just have a some questions. Looks great overall.

test/RuleProcessorDiamond.t.sol
oscarsernarosero: well they could, but I though of just leaving those there since it won't hurt, and would keep us up to date with the version.

test/RuleStorageDiamond.t.sol
oscarsernarosero: same as above

upgrade_version.sh
oscarsernarosero: oh good question. My opinion is we should keep all the same since that way devs know they are using latest version on their contracts which is on sync with latest version of the protocol, but we should prob discuss this with the team.

test/RuleProcessorDiamond.t.sol
oscarsernarosero: Im changing the `1,0,1` to `1,0,0` so they don't get confused with current version

upgrade_version.sh
ShaneDuncan602: Team discussed and decided to keep the version the same throughout the repo.

ShaneDuncan602 (approved):
Looks good.

PR #139, opened 2023-08-29
Se 169 code quality duplicate code and typos

oscarsernarosero:
Half of the ticket tasks had already been taken care of in other tickets. Some condierations:

- `removelevel` function --which original task was to make it camel case-- was removed since this function is never used. Moreover setLevel(0) would do exactly the same.
- The Rule enum was also removed because it is a fossil from when the rules were passed through arrays during the last ice age.
- I took the liberty of removing `defaultFee` from the Fees.sol file since it was never used even though it was not specified in the ticket. I did notice some opportunities for improvement here, but that's more for a design meeting.
- I looked for more stuff to improve, but didn't really find anything, although Im not that good at finding stuff this way, so I kinda limited myself to the scope of the ticket.


ShaneDuncan602 (changes requested):
appAdminstratorOrOwnerOnly is still misspelled. I guess the IToken was already addressed so that all looks good.

oscarsernarosero:
> appAdminstratorOrOwnerOnly is still misspelled. I guess the IToken was already addressed so that all looks good.

oops! I guess I mis-misspelled when I was looking for it. Thank you for that Shane.

ShaneDuncan602 (approved):
Looks good.

TJ-Everett (approved):
Approved. Looks Great Oscar

PR #146, opened 2023-09-06
SE-185-NFT-Module-Add-Memory-Slot-Placeholders-to-Upgradeable-NFT

ShaneDuncan602:


src/token/ProtocolERC721U.sol
```diff
@@ -36,6 +36,8 @@ contract ProtocolERC721U is
 
     /// Base Contract URI
     string public baseUri;
+    /// memory placeholders to allow up variable addition without affecting client upgradeability
+    uint256[49] __gap;
```
oscarsernarosero, line 40: Just to make sure I understand this, we are reserving 49 extra slots in storage this way?

test/helpers/ApplicationERC721UExtra.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import "src/token/ProtocolERC721U.sol";
+
+/**
+ * @title ApplicationERC721U
```
oscarsernarosero: ApplicationERC721UExtra

test/helpers/ApplicationERC721UExtra2.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import "./ProtocolERC721UExtra.sol";
+
+/**
+ * @title ApplicationERC721U
```
oscarsernarosero: ApplicationERC721UExtra2

test/helpers/ProtocolERC721UExtra.sol
```diff
@@ -0,0 +1,169 @@
...
+    PausableUpgradeable
+{
+    using CountersUpgradeable for CountersUpgradeable.Counter;
+    address public handlerAddress;
+    IProtocolERC721Handler handler;
+    CountersUpgradeable.Counter private _tokenIdCounter;
+
+    /// Base Contract URI
+    string public baseUri;
+    /// memory placeholders to allow up variable addition without affecting client upgradeability
```
TJ-Everett, line 39: This reads a little funny 
BUT
My understanding is that we are setting (here) an array of memory and then the variables after that gap to ensure we don't erase these in the upgrade?

TJ-Everett (approved):
This looks good to me, as long as my understanding of the reasoning is correct. One small comment line needs a little more explanation

test/helpers/ApplicationERC721UExtra.sol
```diff
@@ -0,0 +1,37 @@
...
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett
+ * @dev This is an example implementation that App Devs should use.
+ * During deployment, this contract should be deployed first, then initialize should be invoked, then ApplicationERC721UProxy should be deployed and pointed at * this contract. Any special or additional initializations can be done by overriding initialize but all initializations performed in ProtocolERC721U
+ * must be performed
+ */
+
+contract ApplicationERC721UExtra is ProtocolERC721U {
```
oscarsernarosero, line 14: so this one inherits from normal ProtocolERC721U?

test/helpers/ApplicationERC721UExtra.sol
```diff
@@ -0,0 +1,37 @@
...
+    /// Optional Function Variables and Errors. Uncomment these if using option functions:
+    using CountersUpgradeable for CountersUpgradeable.Counter;
+    CountersUpgradeable.Counter private _tokenIdCounter;
+    error OnlyOwnerCanMint();
+    uint256 testVariable1;
+    string testVariable2;
```
oscarsernarosero, line 20: So these would be in slots 51, 52 respectively

test/helpers/ProtocolERC721UExtra.sol
```diff
@@ -0,0 +1,169 @@
...
+    ERC721BurnableUpgradeable,
+    OwnableUpgradeable,
+    UUPSUpgradeable,
+    ProtocolTokenCommonU,
+    PausableUpgradeable
+{
+    using CountersUpgradeable for CountersUpgradeable.Counter;
+    address public handlerAddress;
+    IProtocolERC721Handler handler;
+    CountersUpgradeable.Counter private _tokenIdCounter;
```
oscarsernarosero, line 35: If this contract is meant to be the upgraded version, IProtocolERC721Handler will point to the same slot where baseUri is, and the new baseUri will point to a new unused slot (3)

oscarsernarosero: You have to keep the same order of the variables

test/helpers/ApplicationERC721UExtra2.sol
```diff
@@ -0,0 +1,37 @@
...
+ * must be performed
+ */
+
+contract ApplicationERC721UExtra2 is ProtocolERC721UExtra {
+    /// Optional Function Variables and Errors. Uncomment these if using option functions:
+    using CountersUpgradeable for CountersUpgradeable.Counter;
+    CountersUpgradeable.Counter private _tokenIdCounter;
+    error OnlyOwnerCanMint();
+    uint256 testVariable1;
+    string testVariable2;
```
oscarsernarosero, line 20: and since this contract inherits ProtocolERC721UExtra, then these tow variables will be also on slots 51 and 52. So you are testing that you can add variables like newVariable1, newVariable2 and newVariable3? this looks really good. I just want to make sure Im understanding this well

test/ApplicationERC721U.t.sol
```diff
@@ -971,6 +975,45 @@ contract ApplicationERC721UTest is TestCommon {
...
+
+        vm.stopPrank();
+        vm.startPrank(appAdministrator);
+        ///deploy new modified appliction asset handler contract
+        ApplicationERC721HandlerMod assetHandler = new ApplicationERC721HandlerMod(address(ruleProcessor), address(applicationAppManager), address(applicationNFTProxy), true);
+        ///connect to apptoken
+        ApplicationERC721UExtra(address(applicationNFTProxy)).connectHandlerToToken(address(assetHandler));
+        /// in order to handle upgrades and handler registrations, deregister and re-register with new
+        applicationAppManager.deregisterToken("THRK");
+        applicationAppManager.registerToken("THRK", address(applicationNFTProxy));
```
oscarsernarosero, line 1008: Why is this?

test/helpers/ProtocolERC721UExtra.sol
ShaneDuncan602: That is correct. I can probably make the comment more clear

test/helpers/ProtocolERC721UExtra.sol
oscarsernarosero: Never mind. The original one has the same lines.

test/ApplicationERC721U.t.sol
```diff
@@ -971,6 +975,45 @@ contract ApplicationERC721UTest is TestCommon {
...
+
+        assetHandler.setNFTPricingAddress(address(erc721Pricer));
+        assetHandler.setERC20PricingAddress(address(erc20Pricer));
+        // check to make sure the storage level variables defined after the memory slot are still fine.
+        assertEq(ApplicationERC721UExtra(address(applicationNFTProxy)).getTestVariable1(), 14);
+        assertEq(ApplicationERC721UExtra(address(applicationNFTProxy)).getTestVariable2(), "awesome");
```
oscarsernarosero, line 1014: This looks really good Shane. I would just say, maybe test also setting these variables through this updated contract, and also the new ones (newVariableN) to demonstrate the upgraded variables/functionality.

oscarsernarosero (approved):
Ok took me a little while to understand, but it looks really good Shane. My only suggestion is to add to the test the demonstration of setting the initial test variables through the upgraded contract to see that those still work, and most importantly, test the newVariables by setting them and reading them. Since that's just the lime on the margarita, Im approving it.

src/token/ProtocolERC721U.sol
ShaneDuncan602: Yes. This is how OZ recommends it.

src/token/ProtocolERC721Umin.sol
```diff
@@ -14,7 +14,7 @@ import "./ProtocolTokenCommonU.sol";
 contract ProtocolERC721Umin is Initializable, ERC721EnumerableUpgradeable, ProtocolTokenCommonU {
     address private handlerAddress;
     IProtocolERC721Handler private handler;
-    /// memory placeholders to allow up variable addition without affecting client upgradeability
+    /// memory placeholders to allow variable addition without affecting client upgradeability
```
TJ-Everett, line 17: crystal clear now!

test/helpers/ApplicationERC721UExtra.sol
ShaneDuncan602: Yep, ApplicationERC721UExtra points to the standard ProtocolERC721U. The idea is to reproduce the scenario where the protocol adds a variable which forces an upgrade by the upgradeable nft clients. ApplicationERC721UExtra2 points to ProtocolERC721UExtra which has the extra variables in it.

test/helpers/ApplicationERC721UExtra.sol
ShaneDuncan602: No, those would be the last slots of the array. testVariable1 would be 47, testVariable2 would be 48, and testVariable3 would be 49.

test/helpers/ApplicationERC721UExtra2.sol
ShaneDuncan602: Yes, but no new slots are added. The gap array length gets adjusted. It's a place hold for memory. When a new variable is added, the defined elements in the gap array has to be reduced.

test/ApplicationERC721U.t.sol
ShaneDuncan602: This was a part of the audit. Since the handlers are registered along with the asset, they have to be deregistered and reregistered with the AppManager. If we only try to register, it won't work becaus they are already registered.

test/helpers/ApplicationERC721UExtra.sol
oscarsernarosero: I think you are confused about the variables here, but I know what you're saying.

PR #148, opened 2023-09-06
CREATE-NFT-RULES.md addNFTTransferCounterRule:

TJ-Everett:
added start time param
DEPLOY-APPMANAGER:
create command refactored
DEPLOY-NFT.md:
connectHandlerToToken command refactored


ShaneDuncan602 (changes requested):
This is now the valid deployment of Mumbai for v1.0.1. It needs to be added to DEPLOYMENT-DIRECTORY.md as well as environment variables in the .env file. Also, I think we need to take some of the protocol unit tests(the diamond tests) and put them into a Deployment.t.sol. In there, the setup function would use the .env variables to link to the existing contracts during a fork test. We need a way to test our deployments that is reusable for all environments.

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,604 @@
...
+import {AppRuleDataFacet} from "../src/economic/ruleStorage/AppRuleDataFacet.sol";
+import "src/example/OracleRestricted.sol";
+import "src/example/OracleAllowed.sol";
+import "../src/example/ApplicationERC20Handler.sol";
+import {ApplicationERC20} from "../src/example/ApplicationERC20.sol";
+import "test/helpers/TestCommon.sol";
+
+
+
+contract RuleProcessorDiamondTest is Test, RuleProcessorDiamondTestUtil, TestCommon {
```
ShaneDuncan602: You should only need the TestCommon here. RuleProcessorDiamondTestUtil is being phased out

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,604 @@
...
+            // Diploy the token rule processor diamond
+            ruleProcessor = getRuleProcessorDiamond();
+            console.log("localStorageDiamond ", address(ruleStorageDiamond));
+            console.log("localProcessorDiamond", address(ruleProcessor));
+            forkTest = false;
+            vm.stopPrank();
+        }
+        vm.startPrank(superAdmin);
+        // Connect the ruleProcessor into the ruleStorageDiamond
+        ruleProcessor.setRuleDataDiamond(address(ruleStorageDiamond));
```
ShaneDuncan602: All of this protocol setup is handled in TestCommon. You can refer to some of the other tests to see how the stuff should be set up now. We no longer have to repeat this. It’s done with one function call

test/DeploymentStorage.t.sol
```diff
@@ -0,0 +1,1180 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import "forge-std/Script.sol";
+import "forge-std/Test.sol";
+import "../src/application/AppManager.sol";
+import "./RuleStorageDiamondTestUtil.sol";
+import "../src/application/AppManager.sol";
+import {TaggedRuleDataFacet as TaggedRuleDataFacet} from "../src/economic/ruleStorage/TaggedRuleDataFacet.sol";
```
ShaneDuncan602, line 8: Many of these imports are already defined in TestCommon

test/DeploymentStorage.t.sol
```diff
@@ -0,0 +1,1180 @@
...
+        } else {
+            /// deploy a fresh diamond for testing without --fork-url testing 
+            vm.startPrank(superAdmin);
+            ruleStorageDiamond = getRuleStorageDiamond();
+            console.log("localStorageDiamond ", address(ruleStorageDiamond));
+            forkTest = false; 
+            vm.stopPrank();
+        }
+        vm.startPrank(superAdmin);
+        // Deploy app manager
```
ShaneDuncan602: See the comments from the other test

ShaneDuncan602 (changes requested):
It is very close. Everything is there, but the test need to be modernized to use the new way. We do the test. Great job though. I would recommend that you look at some of the other tests and see how the set up function is and then that way you can see how simplified it is now

ShaneDuncan602 (approved):
Looks awesome. Thanks for all the hard work in this.

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        accs[0] = bytes32("Oscar");
+        accs[1] = bytes32("Tayler");
+        accs[2] = bytes32("Shane");
+        min[0] = uint256(10);
+        min[1] = uint256(20);
+        min[2] = uint256(30);
+        max[0] = uint256(10000000000000000000000000);
+        max[1] = uint256(10000000000000000000000000000);
+        max[2] = uint256(1000000000000000000000000000000);
+        // add empty rule at ruleId 0
```
oscarsernarosero: I think this comment is a fossil from millions of years ago since we don't do empty rules at position 0 since late Precambrian age.

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        vm.startPrank(appAdministrator);
+        applicationAppManager.addGeneralTag(superAdmin, "Oscar"); //add tag
+        assertTrue(applicationAppManager.hasTag(superAdmin, "Oscar"));
+        vm.stopPrank();
+        vm.startPrank(superAdmin);
+        applicationCoin.mint(superAdmin, 10000);
+        uint256 amount = 10;
+        bytes32[] memory tags = applicationAppManager.getAllTags(superAdmin);
+
+        ERC20TaggedRuleProcessorFacet(address(ruleProcessor)).minAccountBalanceCheck(applicationCoin.balanceOf(superAdmin), tags, amount, ruleId);
```
oscarsernarosero: is superAdmin a good test actor? Im not sure if the superAdmin will bypass the rules like previous appAdmin.

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        accs[0] = bytes32("Oscar");
+        accs[1] = bytes32("Tayler");
+        accs[2] = bytes32("Shane");
+        min[0] = uint256(10);
+        min[1] = uint256(20);
+        min[2] = uint256(30);
+        max[0] = uint256(10000000000000000000000000);
+        max[1] = uint256(10000000000000000000000000000);
+        max[2] = uint256(1000000000000000000000000000000);
+        // add empty rule at ruleId 0
```
oscarsernarosero: fossil

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        TaggedRuleDataFacet(address(ruleStorageDiamond)).addBalanceLimitRules(address(applicationAppManager), accs, min, max);
+        uint32 ruleId = TaggedRuleDataFacet(address(ruleStorageDiamond)).addBalanceLimitRules(address(applicationAppManager), accs, min, max);
+        vm.stopPrank();
+        vm.startPrank(appAdministrator);
+        for (uint i = 1; i < 11; i++) {
+            applicationAppManager.addGeneralTag(superAdmin, bytes32(i)); //add tag
```
oscarsernarosero: same questions about superAdmin being able to bypass rules.

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        vm.startPrank(appAdministrator);
+        applicationAppManager.addGeneralTag(superAdmin, "Oscar"); //add tag
+        assertTrue(applicationAppManager.hasTag(superAdmin, "Oscar"));
+        vm.stopPrank();
+        vm.startPrank(superAdmin);
+        uint256 amount = 10000000000000000000000;
+        assertEq(applicationCoin.balanceOf(superAdmin), 10000000000000000000000);
+        bytes32[] memory tags = applicationAppManager.getAllTags(superAdmin);
+
+        ERC20TaggedRuleProcessorFacet(address(ruleProcessor)).minAccountBalanceCheck(applicationCoin.balanceOf(superAdmin), tags, amount, ruleId);
```
oscarsernarosero: I guess it does fail even with superAdmin since the test name indicates this call will revert, but could you test this with an "expectRevert" instead so we know for sure it is failing as expected?

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        uint32 ruleId = TaggedRuleDataFacet(address(ruleStorageDiamond)).addBalanceLimitRules(address(applicationAppManager), accs, min, max);
+        vm.stopPrank();
+        vm.startPrank(appAdministrator);
+        applicationAppManager.addGeneralTag(superAdmin, "Oscar"); //add tag
+        assertTrue(applicationAppManager.hasTag(superAdmin, "Oscar"));
+        vm.stopPrank();
+        vm.startPrank(superAdmin);
+        uint256 amount = 999;
+        bytes32[] memory tags = applicationAppManager.getAllTags(superAdmin);
+        ERC20TaggedRuleProcessorFacet(address(ruleProcessor)).maxAccountBalanceCheck(applicationCoin.balanceOf(superAdmin), tags, amount, ruleId);
```
oscarsernarosero: I'll just keep highlighting superAdmins just in case, but feel free to ignore these if superAdmin turns out to be ok for tests

test/DeploymentProcessor.t.sol
```diff
@@ -0,0 +1,570 @@
...
+        uint32 ruleId = TaggedRuleDataFacet(address(ruleStorageDiamond)).addBalanceLimitRules(address(applicationAppManager), accs, min, max);
+        vm.stopPrank();
+        vm.startPrank(appAdministrator);
+        applicationAppManager.addGeneralTag(superAdmin, "Oscar"); //add tag
+        assertTrue(applicationAppManager.hasTag(superAdmin, "Oscar"));
+        vm.stopPrank();
+        vm.startPrank(superAdmin);
+        uint256 amount = 10000000000000000000000000;
+        assertEq(applicationCoin.balanceOf(superAdmin), 10000000000000000000000);
+        bytes32[] memory tags = applicationAppManager.getAllTags(superAdmin);
```
oscarsernarosero: superAdmin

oscarsernarosero (approved):
Looks really good overall Tayler. Im just not sure if the superAdmin is a good address to test against, so Im gonna hold on approving until we are sure. The failing test should use the expectRevert so we are sure about them failing. If the failing tests do fail with the superAdmin then we are good.

oscarsernarosero (changes requested):
Sorry I meant to request changes

test/DeploymentProcessor.t.sol
TJ-Everett: I have swapped out the super admins for users and we are getting failures when we expect them. The super admins were technically fine since the rule was being checked directly and bypassing the admin checks in the handlers.

test/DeploymentProcessor.t.sol
TJ-Everett: The expect revert doesn't work here since the expected reversion (for foundry) tries to fail at the balanceOf() function. So marking the testFails in the title works the best. I did remove the fail and ensure that the expectedly passing test fails with the BalBelowMin error we want though.

oscarsernarosero (approved):
Great job!

PR #149, opened 2023-09-07
Made conditional changes for code cleanup

ShaneDuncan602:


src/economic/ruleProcessor/ApplicationAccessLevelProcessorFacet.sol
```diff
@@ -43,14 +43,11 @@ contract ApplicationAccessLevelProcessorFacet is IRuleProcessorErrors, IAccessLe
      */
     function checkwithdrawalLimitsByAccessLevel(uint32 _ruleId, uint8 _accessLevel, uint128 _usdWithdrawalTotal, uint128 _usdAmountTransferring) external view returns (uint128) {
         AppRuleDataFacet data = AppRuleDataFacet(actionDiamond.ruleDataStorage().rules);
-        if (data.getTotalAccessLevelWithdrawalRules() < _ruleId) revert RuleDoesNotExist();
```
oscarsernarosero, line 46: why can we get rid of this?

src/token/ProtocolERC20.sol
```diff
@@ -165,7 +165,7 @@ contract ProtocolERC20 is ERC20, ERC165, ERC20Burnable, ERC20FlashMint, Pausable
         if (!appManager.isAppAdministrator(msg.sender) && !appManager.isRegisteredAMM(msg.sender)) {
             revert CallerNotAuthorizedToMint();
         }
-        if (MAX_SUPPLY > 0 && totalSupply() + amount > MAX_SUPPLY) {
+        if (totalSupply() + amount > MAX_SUPPLY) {
```
oscarsernarosero: I thought we had stablished that MAX_SUPPLY = 0 means infinite supply

oscarsernarosero (changes requested):
I just have a couple questions. Also, it looks like we are not using isValue anymore from the fees. If this is the case, then we should remove the field from the rule struct.

ShaneDuncan602:
> I just have a couple questions. Also, it looks like we are not using isValue anymore from the fees. If this is the case, then we should remove the field from the rule struct.

It is removed from the struct. That struct is defined at the top of Fees.sol

src/economic/ruleProcessor/ApplicationAccessLevelProcessorFacet.sol
ShaneDuncan602: The same check is happening in getAccessLevelWithdrawalRule. ToB said it was redundant.

src/token/ProtocolERC20.sol
ShaneDuncan602: I was torn on this one. I assume this is the standard way that the MAX_SUPPLY is used and that's why ToB thought is was unnecessary. I am good with putting it back.

TJ-Everett: I agree with Oscar, otherwise we would have to set it at maxUint() or something similar for unlimited minting.

TJ-Everett (changes requested):
This looks great Shane, we just need to flip that MAX_SUPPLY back to >0. My thinking is that since we are focused on gaming ecosystems and hgih volume economies we should not assume that these collections will have max caps or insanely high max caps. If not already notated in the contact, can we add a comment line?

src/token/ProtocolERC20.sol
oscarsernarosero: I guess this really a design question. Do we want to force devs to set a max supply without having a default? is the infinite supply default value a desirable value? do we want to set this variable at deployment time?

src/economic/ruleProcessor/ApplicationAccessLevelProcessorFacet.sol
oscarsernarosero: ok I doubled checked and you're spot on

oscarsernarosero (approved):
Looks good Shane!

TJ-Everett (approved):
Approved for me. Oscar I agree with you and think this is worth revisiting in the next design meeting / see if we need to get product feedback. 
For now, since we are all in agreement we can merge this thing since Shane is going to be out for a bit.

PR #151, opened 2023-09-13
Events

oscarsernarosero:
This work picks up lost PR for branch https://github.com/thrackle-io/tron/tree/ui_event_refactor only for external branch.

- Add/Remove events were unified into a single event so the historical indexer can work properly. These unified events have an extra boolean field which indicates if it is adding (true) or removing (false).
- AppManagerDeployed and AppManagerDeployedForUpgrade were modified to better serve the purpose. They used to log their own address, but this datum is available in the transaction receipt which makes it duplicated data. On the other hand, we don't know who the superAdmin and appAdmin are, or even the name of the app when these contracts are deployed even though this information is setup at construction time. Therefore, these events were modified to log the root address and the appName instead of the appManager's own address.


TJ-Everett (approved):
Looks good Oscar

TJ-Everett (approved):
Looks great Oscar.

PR #153, opened 2023-09-19
External test

bfcoursewool:
Some tweaks needed to be made in Tron to end up with DOOM being properly compatible again after all the audit changes and whatnot... these are those changes.


TJ-Everett (approved):
quick and easy pr here. Looks great Rob, from the Tron side. I don't know much about the dockerfile changes.

oscarsernarosero (approved):

PR #154, opened 2023-09-19
Se 171 code quality comments and function parameters external

oscarsernarosero:


src/economic/ruleProcessor/ApplicationRiskProcessorFacet.sol
```diff
@@ -53,7 +55,7 @@ contract ApplicationRiskProcessorFacet is IRuleProcessorErrors, IRiskErrors {
                         revert BalanceExceedsRiskScoreLimit();
                     } else {
                         ///Jump out of loop once risk score is matched to array index
-                        break;
+                        return;
```
ShaneDuncan602, line 58: Doesn't it still need to hit lines 65-70 once it is matched? For the situation where the score matches the highest value? If not, we should get rid of those lines.

src/economic/ruleProcessor/ApplicationRiskProcessorFacet.sol
```diff
@@ -101,14 +103,10 @@ contract ApplicationRiskProcessorFacet is IRuleProcessorErrors, IRiskErrors {
...
         ApplicationRuleStorage.TxSizePerPeriodToRiskRule memory rule = data.getMaxTxSizePerPeriodRule(ruleId);
-        /// reseting the "tradesWithinPeriod", unless...
-        uint128 amountTransactedInPeriod = amount;
-        /// if (we have been in current period for longer than the last update)...
-        if (((block.timestamp - rule.startingTime) % (uint256(rule.period) * 1 hours)) >= (block.timestamp - lastTxDate)) {
-            /// This means that the last trades "tradesWithinPeriod" were inside current period,
-            /// and we need to acumulate this trade to the those ones
-            amountTransactedInPeriod = amount + _usdValueTransactedInPeriod;
-        }
+        /// reseting the "tradesWithinPeriod", unless we have been in current period for longer than the last update
```
ShaneDuncan602: resetting

src/economic/ruleProcessor/ApplicationRiskProcessorFacet.sol
```diff
@@ -101,14 +103,10 @@ contract ApplicationRiskProcessorFacet is IRuleProcessorErrors, IRiskErrors {
...
+        uint128 amountTransactedInPeriod = rule.startingTime.isWithinPeriod(rule.period, lastTxDate) ? 
+        amount + _usdValueTransactedInPeriod: amount;
+
```
ShaneDuncan602, line 109: Great use of the ternary here

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -72,14 +72,10 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, ITagRuleErrors,
...
-        for (uint i = 0; i < toTags.length; ) {
+        for (uint i; i < toTags.length; ) {
             uint256 max = data.getBalanceLimitRule(ruleId, toTags[i]).maximum;
             /// if a max is 0 it means it is an empty-rule/no-rule. a max should be greater than 0
-            if (max > 0) {
-                if (balanceTo + amount > max) {
-                    revert MaxBalanceExceeded();
-                }
-            }
+             if (max > 0 && balanceTo + amount > max) revert MaxBalanceExceeded();
```
ShaneDuncan602, line 78: nice

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -144,6 +136,7 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, ITagRuleErrors,
         toTags.checkMaxTags();
         TaggedRuleDataFacet data = TaggedRuleDataFacet(Diamond.ruleDataStorage().rules);
         uint totalRules = data.getTotalMinBalByDateRule();
+        uint finalBalance = balance - amount;
```
ShaneDuncan602: Is creating this variable in memory cheaper than the single operation? I could see if the expression was evaluated multiple times but I'm not sure it would be more efficient for single use.

src/economic/ruleProcessor/ERC721RuleProcessorFacet.sol
```diff
@@ -38,14 +38,9 @@ contract ERC721RuleProcessorFacet is IERC721Errors, IRuleProcessorErrors, IMaxTa
...
                     // if within time period, add to cumulative
-                    if (rule.startTs.isWithinPeriod(period, lastTransferTime)) {
-                        cumulativeTotal = transfersWithinPeriod + 1;
-                    } else {
-                        cumulativeTotal = 1;
-                    }
-                    if (cumulativeTotal > tradesAllowedPerDay) {
-                        revert MaxNFTTransferReached();
-                    }
+                    cumulativeTotal = rule.startTs.isWithinPeriod(period, lastTransferTime) ?
```
ShaneDuncan602, line 41: Are these ternaries more efficient than standard if-else blocks? IMO they are harder to read and easier to overlook.

src/economic/ruleProcessor/RiskTaggedRuleProcessorFacet.sol
```diff
@@ -45,7 +45,7 @@ contract RiskTaggedRuleProcessorFacet is IRuleProcessorErrors, IRiskErrors {
                     revert TransactionExceedsRiskScoreLimit();
                 } else {
                     ///Jump out of loop once risk score is matched to array index
-                    break;
+                    return;
```
ShaneDuncan602, line 48: Same question as above.

src/example/ERC721/not-upgradeable/ApplicationERC721MintForAFee.sol
```diff
@@ -0,0 +1,134 @@
...
+     * @dev Function mints a new token to caller at mintPrice with tokenId incremented by 1 from previous minted token.
+     * @notice This function assumes the mintPrice is in Chain Native Token and in WEI units
+     * @param to Address of recipient
+     */
+    function safeMint(address to) public payable override whenNotPaused {
+        if(mintPrice == 0) revert PriceNotSet();
+        if (msg.value < mintPrice) revert MintFeeNotReached();
+        uint256 tokenId = _tokenIdCounter.current();
+        _tokenIdCounter.increment();
+        // withdrawMintFees(msg.value); // uncomment this is automatic withdrawal is desired. withdrawMintFees should be marked as private
```
ShaneDuncan602: uncomment this is automatic withdrawal if desired

src/example/ERC721/upgradeable/ApplicationERC721UpgAdminMint.sol
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @title ApplicationERC721UpgAdminOrOwnerMint
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett
+ * @dev This is an upgradeable example implementation of the protocol ERC721 where minting is only available for app administrators.
+ * During deployment, this contract should be deployed first, then initialize should be invoked, then ApplicationERC721UProxy should be deployed and pointed at * this contract. Any special or additional initializations can be done by overriding initialize but all initializations performed in ProtocolERC721U
+ * must be performed
+ */
+
+contract ApplicationERC721Upgradeable is ProtocolERC721U {
```
ShaneDuncan602, line 14: I think ProtocolERC721U.sol safeMint is missing the appAdministratorOrOwnerOnly modifier

src/example/script/ApplicationERC721.s.sol
ShaneDuncan602, line 1: I think we will need matching deploy scripts for each of the newly created example 721's. It will be confusing for users to try and match the example to the generic script and change it.

test/ERC721Examples.t.sol
ShaneDuncan602, line 1: Love this

ShaneDuncan602 (changes requested):
Really good stuff. I just have a couple of questions and comments. Also, it looks like the tests are failing.

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
TJ-Everett: also for code consistency we should use uint256 (even though it is the same thing)...

src/example/ERC721/not-upgradeable/ApplicationERC721MintForAFee.sol
TJ-Everett: if? Not is?

src/example/ERC721/not-upgradeable/ApplicationERc721AdminOrOwnerMint.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: UNLICENSED
```
TJ-Everett: the C in the name of this file should be capitalized

TJ-Everett (changes requested):
Added my thoughts to the current PR. Shane's questions and then a few typos.

src/economic/ruleProcessor/ApplicationRiskProcessorFacet.sol
oscarsernarosero: not if it is inside this block because this means we found the risk score range we are looking for (the conditional of line 52) so there is no need to check for the last range. It will get to lines 65-70 if the risk score was not less than the any of the rules risk-score limits which means it is in the highest range.

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
oscarsernarosero: It is not created in memory, but in the stack, and yes, it is cheaper than calculating the result once per tag. Although you got me thinking, so I went ahead and tested it:

```
contract StackA{

    uint constant x = 10;
    /// gas: 1660 with a=7, b=4, c=10
    function test(uint a, uint b, uint c) external pure{
        uint d = a + b;
        for(uint i; i < c;){
            bool r = d < x;
            unchecked{
                ++i;
            }
        }
    }
}

contract StackB{

    uint constant x = 10;
    /// gas: 3469 with a=7, b=4, c=10
    function test(uint a, uint b, uint c) external pure{
        
        for(uint i; i < c;){
            uint d = a + b;
            bool r = d < x;
            unchecked{
                ++i;
            }
        }
    }
}
```

And the test shows that it is way cheaper to do the math once since it reduces the operation gas cost by 52%.

src/economic/ruleProcessor/ERC721RuleProcessorFacet.sol
oscarsernarosero: You got me thinking again, so I went ahead and tested the gas efficiency:
```
contract WithTernary{
    /// gas: 614 with a=5; gas: 604 with a =12
    function test(uint a) external pure returns (uint r){
        r = a > 10 ? 10 : a;
    }
}

contract WithoutTernary{
    /// gas: 607 with a=5; gas: 617 with a =12
    function test(uint a) external pure returns (uint r){
        if (a > 10){
            r = 10;
        }else{
            r = a;
        }
    }
}
``` 

So it looks like the gas efficiency is the same in average since the ternary is slightly more efficient in the negative cases, but the if-else is slightly more efficient in the positive cases. Although the gap in efficiency is greater when the ternary beats the if-else, the efficiency difference is too small to consider it a factor. So I personally think that the ternary assignment saves us a lot of code lines, and it makes the code more readable imo.

src/example/ERC721/not-upgradeable/ApplicationERC721MintForAFee.sol
oscarsernarosero: yeah it was supposed to be "if". Should I leave it? Would that be confusing for devs?

src/example/ERC721/upgradeable/ApplicationERC721UpgAdminMint.sol
oscarsernarosero: Great catch!

oscarsernarosero: Oh! actually it is there.

test/ERC721Examples.t.sol
oscarsernarosero: Thank you!

src/example/ERC721/not-upgradeable/ApplicationERc721AdminOrOwnerMint.sol
oscarsernarosero: Great catch!

src/example/ERC721/not-upgradeable/ApplicationERC721MintForAFee.sol
ShaneDuncan602: Ha, yeah. That's what I meant. I changed the wrong is!

src/example/ERC721/upgradeable/ApplicationERC721UpgAdminMint.sol
ShaneDuncan602: Yep, sorry. My external branch was behind the last couple of merges.

src/example/ERC721/not-upgradeable/ApplicationERc721AdminOrOwnerMint.sol
oscarsernarosero: This is what was causing the build to fail in GitHub. Thanks for saving me a ton of frustration.

ShaneDuncan602 (approved):
Looks really good.

TJ-Everett (approved):
Looks great Oscar!

PR #155, opened 2023-09-26
External docker fixes

bfcoursewool:
Some things in the docker stack broke due to new versions of stuff, and these are the fixes. Foundry must have updated, requiring a higher version of GLIBC than was provided by the "bullseye" release of Debian (which was the base image for the tron docker container). Upgrading from "bullseye" to "bookworm" introduced some issues with python... apparently python 3.11 requires the use of virtual environments, so I added that. Additionally, our one python dependency in the requirements.txt file, eth-abi, was pretty old, and that was also causing issues with python3.11 so I updated that as well. After lots of retries, it seems to be finally working again, huzzah!


Dockerfile.tron
```diff
@@ -1,7 +1,7 @@
-FROM debian:bullseye
+FROM debian:bookworm
 
 RUN apt update
-RUN apt install -y curl git make procps python3 python3-pip
+RUN apt install -y curl git make procps python3 python3-pip python3.11-venv
 
 WORKDIR /usr/local/src/delylah
```
TJ-Everett: Can we rename this to tron from delylah?

TJ-Everett (changes requested):
This is failing the GH tests - then also found a reference to Delylah

bfcoursewool:
> This is failing the GH tests - then also found a reference to Delylah

Ahh, so the same problem is happening on the GH workflow server... it's using Ubuntu 20.04 and it looks like the GLIBC version on that OS is also behind what we need. Gonna try to update the workflow to use a newer Ubuntu...

Dockerfile.tron
bfcoursewool: Done

TJ-Everett (approved):
Looks good!

oscarsernarosero (approved):
Mental note: now we are using eth-abi 4.2.1 so expect some things to crash.

PR #156, opened 2023-10-06
Se 342 set up diamond test through echidna

ShaneDuncan602:
This is just a simple example of creating the RuleStorageDiamond and property testing the version facet.


TJ-Everett (approved):
Somehow the two files changed for the PR doesn't seem to reflect the amount of work/frustration I know when in to get Echidna working.

test/echidna/helpers/TestCommonEchidna.sol
```diff
@@ -1,31 +1,253 @@
...
+        // deploy the facet
+        assembly {
+            _facetAddress := create(0, add(_bytecode, 0x20), mload(_bytecode))
+        }
+
+        // Get the facet selectors.
+        inputs[2] = _facetName;
+        bytes memory res = vm.ffi(inputs);
+        _selectors = abi.decode(res, (bytes4[]));
+        return (_facetAddress, _selectors);
```
oscarsernarosero, line 119: So the fix was just to add the `end=""` ?

test/echidna/property/RuleStorageDiamond.e.sol
```diff
@@ -0,0 +1,25 @@
...
+
+    constructor() {
+        ruleStorageDiamond = _createRuleStorageDiamond();
+        VersionFacet(address(ruleStorageDiamond)).updateVersion("1,0,0");
+    }
+
+    /* ------------------------------ INVARIANTS -------------------------------- */
+    /// Test the version
+    function echidna_version_notblank() public returns (bool) {
+        return (bytes(VersionFacet(address(ruleStorageDiamond)).version()).length != 0);
```
oscarsernarosero: Should we test more functionality in the diamond? like upgrades, adding a rule and stuff like that?

oscarsernarosero (approved):
Looks really good Shane. Great job. Im just curious if you think the test on the diamond is enough as it is.

ShaneDuncan602:
> Looks really good Shane. Great job. Im just curious if you think the test on the diamond is enough as it is.

This was just to give an example of how we can build a diamond and test. We still have to get together and decide the best way to do it.

test/echidna/property/RuleStorageDiamond.e.sol
ShaneDuncan602: This is invariant testing. It only tests the properties. We could add assertion testing in the future.

PR #158, opened 2023-10-10
Se 286 break the application deployment into mutliple scripts

TJ-Everett:


src/example/script/Application_Deploy_01_AppMangerAndAssets.s.sol
```diff
@@ -0,0 +1,57 @@
...
+
+        /// create ERC20 token 2
+        ApplicationERC20 coin2 = new ApplicationERC20("Dracula Coin", "DRAC", address(applicationAppManager));
+        applicationCoinHandler2 = new ApplicationERC20Handler(vm.envAddress("RULE_PROCESSOR_DIAMOND"), address(applicationAppManager), address(coin2), false);
+        coin2.connectHandlerToToken(address(applicationCoinHandler2));
+
+        /// create NFT
+        ApplicationERC721 nft1 = new ApplicationERC721("Clyde", "CLYDEPIC", address(applicationAppManager), vm.envString("APPLICATION_ERC721_URI_1"));
+        applicationNFTHandler = new ApplicationERC721Handler(vm.envAddress("RULE_PROCESSOR_DIAMOND"), address(applicationAppManager), address(nft1), false);
+        nft1.connectHandlerToToken(address(applicationNFTHandler));
```
ShaneDuncan602: These assets need to be registered with the AppManager or none of the rules will work.

src/example/script/Application_Deploy_02_OracleAndPricing.s.sol
```diff
@@ -0,0 +1,68 @@
...
+        ApplicationERC721Handler applicationNFTHandler = ApplicationERC721Handler(vm.envAddress("APPLICATION_ERC721_HANDLER"));
+        
+        /// Deploy Oracle Contracts 
+        new OracleAllowed();
+        new OracleRestricted();
+
+        /// Register the tokens with the application's app manager
+        applicationAppManager.registerToken("Frankenstein Coin", vm.envAddress("APPLICATION_ERC20_ADDRESS"));
+        applicationAppManager.registerToken("Dracula Coin", vm.envAddress("APPLICATION_ERC20_ADDRESS_2"));
+        applicationAppManager.registerToken("Frankenstein Picture", vm.envAddress("APPLICATION_ERC721_ADDRESS_1"));
```
ShaneDuncan602: Nevermind! It's here. Seems like maybe it should be in the other script though.

ShaneDuncan602 (approved):
One suggestion but it's not a big deal and will work either way. Also, when this goes to internal, it will be a difficult merge because staking and AMM's will need to be added.

src/example/script/Application_Deploy_01_AppMangerAndAssets.s.sol
```diff
@@ -0,0 +1,65 @@
...
+import "../application/ApplicationHandler.sol";
+import "../ApplicationERC20Handler.sol";
+import "../ApplicationERC20.sol";
+import {ApplicationERC721Handler} from "../ApplicationERC721Handler.sol";
+import {ApplicationERC721} from "../ERC721/not-upgradeable/ApplicationERC721AdminOrOwnerMint.sol";
+import {ApplicationAppManager} from "../ApplicationAppManager.sol";
+
+
+/**
+ * @title Application Deploy App Manager Script
```
oscarsernarosero: add 01 here somewhere

src/example/script/Application_Deploy_02_OracleAndPricing.s.sol
```diff
@@ -0,0 +1,68 @@
...
+import {ApplicationERC721Handler} from "../ApplicationERC721Handler.sol";
+import {ApplicationERC721} from "../ERC721/not-upgradeable/ApplicationERC721AdminOrOwnerMint.sol";
+import {ApplicationAppManager} from "../ApplicationAppManager.sol";
+import "../OracleRestricted.sol";
+import "../OracleAllowed.sol";
+import "../pricing/ApplicationERC20Pricing.sol";
+import "../pricing/ApplicationERC721Pricing.sol";
+
+/**
+ * @title Application Deploy All Script
```
oscarsernarosero: add 02 here somewhere

src/example/script/Application_Deploy_01_AppMangerAndAssets.s.sol
```diff
@@ -0,0 +1,65 @@
...
+
+        /// Create NFT
+        ApplicationERC721 nft1 = new ApplicationERC721("Clyde", "CLYDEPIC", address(applicationAppManager), vm.envString("APPLICATION_ERC721_URI_1"));
+        applicationNFTHandler = new ApplicationERC721Handler(vm.envAddress("RULE_PROCESSOR_DIAMOND"), address(applicationAppManager), address(nft1), false);
+        nft1.connectHandlerToToken(address(applicationNFTHandler));
+
+        /// Register the tokens with the application's app manager
+        applicationAppManager.registerToken("Frankenstein Coin", address(coin1));
+        applicationAppManager.registerToken("Dracula Coin", address(coin2));
+        applicationAppManager.registerToken("Clyde Picture", address(nft1));
```
oscarsernarosero: I only have an issue with these 3 tokens being deployed all at the same time since from previous experience this seems too much for the blockchain to handle all at once. I would suggest going as granular as deploying token by token. Might seem too conservative, but I think that way we can secure a good result. 

Another option here, and this might be something you want to play with, is introducing some inactive periods of time between the deployments of the tokens using [vm.sleep()](https://book.getfoundry.sh/cheatcodes/sleep). I have never used it, and I don't know if that would work, but I think it's worth the shot if you don't think splitting this in 4 pieces is a good idea.

oscarsernarosero (changes requested):
I think splitting the script_01 is necessary to make sure it would be a successful and clean deployment on mainnet (testnet might behave much better so we can't really assume that mainnet will be just as smooth), Or! you could experiment with vm.sleep() cheatcode from Foundry to basically achieve the same result as splitting the script. If you go with the vm.sleep route, you could even rearrange  the order of deployment to better use the inactive (like deploy first all the token contracts, then wait, then deploy all the token handlers, then wait, then connect them all, etc.)

src/example/script/Application_Deploy_01_AppMangerAndAssets.s.sol
TJ-Everett: Given that this is most likely going to be used on mainnet next, I don't want to try anything too risky / new. I can happily break these up into per token if you think that is necessary, however my concern with a per token script is that you might as well use the individual forge create commands at that point instead of a script. 

In the user guides we tell devs to go this route as these scripts would be just for internal use.

ShaneDuncan602:
> I think splitting the script_01 is necessary to make sure it would be a successful and clean deployment on mainnet (testnet might behave much better so we can't really assume that mainnet will be just as smooth), Or! you could experiment with vm.sleep() cheatcode from Foundry to basically achieve the same result as splitting the script. If you go with the vm.sleep route, you could even rearrange the order of deployment to better use the inactive (like deploy first all the token contracts, then wait, then deploy all the token handlers, then wait, then connect them all, etc.)

Hmm, this is very interesting. First, I agree that it might be a good idea to even split the first script up more. It may be a little overkill but I'd rather do too much than too little. However, the vm.sleep seems like a route we should investigate. I didn't even know it existed. Shouldn't be too hard to put it into the UI script and test it.

TJ-Everett:
Further broke out the scripts so that each token is deployed via its own script as well as broke up the oracle and pricing scripts to ensure they are as light as possible. 

When trying to use vm.sleep() in the ApplicationDeployAll.s.sol file it looks like we do not have that function in the VM.sol list in the lib. This needs to be added to the lib > forge-std > src > VM.sol functions list.

TJ-Everett:
Updating to the newest release of Foundry will also solve this issue and be significantly less "hacky".

oscarsernarosero (approved):
Looks great Tayler

PR #159, opened 2023-10-11
Se 291 create foss style oracle and rule structure

oscarsernarosero:
### Don't merge this branch!

A list of assumptions and design guidelines alongside a diagram of the interaction of the different parts can be found [here](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/195362829/FOSS+Style+Oracle). 

Besides the doc, the unit test is very telling.

I understand now the necessity for 2 separate oracles since there are 2 very different processes (validation of address and staking process), and separating them might even simplify a bit the current oracle contract. For right now tho, this is a good proof of concept.


src/Oracle/SoftStakingOracle.sol
ShaneDuncan602, line 1: Probably doesn't matter for this POC, but we will need to change the name of this in the future. It will need to be more generic.

src/Oracle/SoftStakingOracle.sol
```diff
@@ -0,0 +1,141 @@
...
+            if (msg.value > 0){
+                (bool sent, bytes memory data) = payable(account).call{value: msg.value}("");
+                if(!sent) revert TransferFailed(data);
+            }
+        }
+    }
+
+    function _getStatusPerAccount(address account) internal view returns(uint8 _status){
+        unchecked{
+            _status = uint8(statusPerAccount[account]) - 1;
```
ShaneDuncan602: What if the status is 0 here?

src/Oracle/SoftStakingOracle.sol
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    function completeRequest(uint128 _requestId, bool isApproved, uint256 gasUsed) external onlyOracle{
+        Request memory _req = requestById[_requestId];
+        uint256 balance = _req.balance;
+        /// oracle should always check this before sending tx to avoid a malicious attack
+        if(balance < gasUsed) revert NotEnoughDeposit(gasUsed);
```
ShaneDuncan602: Would we really want to revert this if the client didn't send enough gas to begin with? They sent the minimum amount so that is really on the dev. If this reverts, the user would have no way of knowing their request is in limbo. Also, I wonder if the offchain would send a single check at a time or if it would send a batch of them.

src/Oracle/SoftStakingOracle.sol
```diff
@@ -0,0 +1,141 @@
...
+
+    function updateOracleOrigin(address newOracleOrigin) external onlyOwner{
+        oracleOrigin = newOracleOrigin;
+    }
+
+    /**
+     * @dev Function to withdraw a specific amount from this contract to oracleOrigin address.
+     * @param _amount the amount to withdraw (WEIs)
+     */
+    function withdrawAmount(uint256 _amount) external onlyOwner(){
```
ShaneDuncan602: Why is this modifier different than the other onlyOwners? It has onlyOwner() where the others are just onlyOwner with no parenthesis.

src/Oracle/SoftStakingOracle.sol
```diff
@@ -0,0 +1,141 @@
...
+        if (oracleOrigin == address(0x00)) revert OracleOriginNotSet();
```
ShaneDuncan602: So they could only withdraw the funds to by transferring them to the token contract?

src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
```diff
@@ -70,6 +70,25 @@ contract ERC20RuleProcessorFacet is IRuleProcessorErrors, IERC20Errors {
...
+        RuleDataFacet data = RuleDataFacet(Diamond.ruleDataStorage().rules);
+        if ( data.getTotalStatusOracleRules() <= _ruleId ) revert RuleDoesNotExist();
+        NonTaggedRules.StatusOracleRule memory statusOracleRule = data.getStatusOracleRule(_ruleId);
+        address oracleAddress = statusOracleRule.oracleAddress;
+        /// bytes4(keccak256(bytes('requestStatus(address,address)'))) = 0x7c44325c
+        (bool success, bytes memory res) = oracleAddress.call{value: msg.value}(abi.encodeWithSelector(0x7c44325c, account, tokenAddress));
+        if(! success) revert OracleCheckFailed(res);
+        /// improve following line with Yul
+        (status, _requestId) = abi.decode(res, (uint8, uint128));
+        if (status == 0) revert OracleCheckFailed("ACCOUNT DENIED");
```
ShaneDuncan602, line 89: We wouldn't want this to revert here. This oracle check needs to only send back the result unless it blows up. If they are denied, the handler or token could decide to revert, but I think it's best to leave that decision to that point. That would probably be better than making them use a try/catch because they may need to change some state. The scenario I'm thinking of is if they are already staked and this is a recheck.

src/example/ApplicationERC721HandlerWStatusOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+        address _assetAddress,
+        bool _upgradeMode
+    ) ProtocolERC721Handler(_ruleProcessorProxyAddress, _appManagerAddress, _assetAddress, _upgradeMode) {}
+
+
+    function checkStatusOracle(address account) external payable onlyOwner returns (uint8 status, uint128 _requestId){
+        if (statusOracleRuleActive){ 
+            // this should return either 1: approved or 2: pending. A 0 would cause a revert in
+            // the transaction which means 0 will never make it until here.
+            (status, _requestId) = ruleProcessor.checkStatusOraclePasses{value: msg.value}(statusOracleRuleId, account, msg.sender);
```
ShaneDuncan602, line 38: Does it save gas to not have the return statements? I see a lot of people omitting these so it makes me wonder.

src/token/ProtocolERC721WSoftStaking.sol
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * @dev this function returns the handler address
+     * @return handlerAddress
+     */
+    function getHandlerAddress() external view override returns (address) {
+        return handlerAddress;
+    }
+}
```
ShaneDuncan602: I don't think we would need this contract. This would be a minimal implementation that would only need the interface. We currently only have the IProtocolERC721UMin.sol which is for upgradeable but it could be easily modified to work by changing the import. The way you have this works but we will eventually use this to get more accurate gas assessments so we probably want to keep the protocol footprint as small as possible.

test/ApplicationERC721.t.sol
```diff
@@ -266,6 +271,143 @@ contract ApplicationERC721Test is TestCommon {
...
+        softStakingNft.safeMint(user1);
+
+        /// now let's get this party started
+        vm.stopPrank();
+        vm.startPrank(user1);
+        uint256 balanceBefUser = user1.balance;
+        /// let's make sure first that the staking nft is in the UNLOCKED status (0)
+        assertEq(uint8(softStakingNft.stakingStatusPerNFT(1)), 0);
+        /// we make sure that the staking can't be started if we don't pay the min gas deposit
+        vm.expectRevert(abi.encodeWithSignature("HandlerFailed(bytes)", abi.encodeWithSignature("OracleCheckFailed(bytes)", abi.encodeWithSignature("NotEnoughDeposit(uint256)",MIN_GAS_DEPOSIT ))));
```
ShaneDuncan602, line 316: this is pretty cool

test/ApplicationERC721.t.sol
```diff
@@ -266,6 +271,143 @@ contract ApplicationERC721Test is TestCommon {
...
+        softStakingNft.stakeNFT{value: MIN_GAS_DEPOSIT - 1}(1);
+        /// now we send the absolute min amount and we should be good
+        softStakingNft.stakeNFT{value: MIN_GAS_DEPOSIT}(1);
```
ShaneDuncan602, line 322: We should add a check to make sure the event it emitted since it's so important to the whole process.

ShaneDuncan602 (changes requested):
Great job, dude. I have a few suggestions and quite a few questions, but you killed this. The test is very thorough too.

TJ-Everett (commented):
I don't have any additional changes/questions other than what Shane has already asked.

src/Oracle/SoftStakingOracle.sol
oscarsernarosero: it would return a 255, but it never actually gets there because that would trigger an offchain check which then would result in returning a PENDING status or "2".

src/Oracle/SoftStakingOracle.sol
oscarsernarosero: I see it as a malicious/extremeGasPrice security measure. We can build a function to deposit more gas to a Request in case this happens (would also need a frontend to notify about this). I didn't want to spend more time on it since JB has made it clear this is just a POC. Good to through if we decide to take this on tho.

src/Oracle/SoftStakingOracle.sol
oscarsernarosero: ah! That was just me. It is the same since the parenthesis has no effect. Im removing them.

src/Oracle/SoftStakingOracle.sol
oscarsernarosero: They are sending it to the offchain signer of the oracle since that is the contract that is actually paying for the gas. I named that account as OracleOrigin although maybe a better name would be DataProvider. You would prob come up with a better name, so Im opened to suggestions there.

src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
oscarsernarosero: Ok. I disagree here but let's take this to @jjbailey-io. I personally think a try catch is a good idea since that way we don't break the pattern of the other rules and we have effectively the same functionality. Also, a _recheck_ wouldn't cause a revert since the address would have an APPROVED/DENIED state.

src/example/ApplicationERC721HandlerWStatusOracle.sol
oscarsernarosero: I don't think it saves gas. To me it's just less lines of code. It has the return statement and the variable creation implicit in the `returns (uint8 status, uint128 _requestId)`

src/token/ProtocolERC721WSoftStaking.sol
oscarsernarosero: ok. So you're saying I should delete some of these functions? maybe the pause functions?

test/ApplicationERC721.t.sol
oscarsernarosero: I did try this but foundry didn't recognize the vm.expectEmit cheatcode. I'll try again, although I did check in the logs that it was emitting the event.

src/Oracle/SoftStakingOracle.sol
ShaneDuncan602: Ok, I was thinking oracleOrigin was the asset.

src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
ShaneDuncan602: It was decided to keep the revert in order to not stray too far from the current rule logic.

src/token/ProtocolERC721WSoftStaking.sol
```diff
@@ -18,7 +14,7 @@ import "../token/IProtocolERC721Handler.sol";
  * @notice This is the base contract for all protocol ERC721s
  */
 
-contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon, AppAdministratorOrOwnerOnly {
+contract ProtocolERC721 is IProtocolERC721Min, ERC721Enumerable, ProtocolTokenCommon, Ownable {
```
ShaneDuncan602, line 17: I was thinking that the IProtocolERC721Min would be implemented directly by the Application example but I guess it really isn't much different than this. This should provide comparable gas tests. Good job.

test/ApplicationERC721.t.sol
ShaneDuncan602: It seems strange that Foundry makes you put line 320 in there. That's how it works but it seems odd. I think Foundry needs to put some extra work into it.

ShaneDuncan602 (approved):
Looks really good. Great job on this.

TJ-Everett (approved):
Looks Great Oscar, very well done.

PR #162, opened 2023-10-16
Se 287 rbac restructure

oscarsernarosero:
Implemented the migration for superAdmin role and the at-least-one-admin policy for all the other admins.

Highlights:

- Only-one-super-admin-per-app rule was implemented. `grantRole` function had to be overridden to revert when invoked to effectively force all role grants to be done through the appropriate functions in the appManager.
- A new internal role was created to make the transition of the super-admin role possible: PROPOSED_SUPER_ADMIN. This role doesn't emit any events because of its internal nature.
- The super-admin role can be migrated to another address through a 2-step propose-confirm mechanism for security reasons.
- The functions `revokeRole` and `renounceRole` were overridden to enforce the at-least-one-admin policy.
- The docs were updated and a table with the role hashes was added.

# Update

After some iterations with the team, it was decided that only the SupeAdmins would have the restriction of at least 1 superAdmin account. All other roles can have 0 members.


src/application/AppManager.sol
```diff
@@ -97,6 +100,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+    function renounceRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
+        /// enforcing the min-1-admin requirement. Only PROPOSED_SUPER_ADMIN_ROLE should be able to bypass this rule
+        if(role != PROPOSED_SUPER_ADMIN_ROLE && getRoleMemberCount(role) < 2) revert CannotRenounceIfOnlyOneAdmin();
+        AccessControl.renounceRole(role, account);
+    }
+
+    /**
+     * @dev This function overrides the parent's grantRole function. Its purpose is to enforce an at-least-one-admin policy.
+     * @param role the role to revoke.
+     * @param account address losing to the role.
```
TJ-Everett: address of revoked role*

src/application/AppManager.sol
```diff
@@ -97,6 +100,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+     * @notice you can't revoke a role if that means that nobody would have that role. At least one account has to have it except
+     * for PROPOSED_SUPER_ADMIN_ROLEs which is a transitionary one. Make sure there will be at least one account for all 
+     * the other roles.
+     */
+    function revokeRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
+        /// enforcing the min-1-admin requirement.
+        if(role != PROPOSED_SUPER_ADMIN_ROLE && getRoleMemberCount(role) < 2) revert CannotRenounceIfOnlyOneAdmin();
```
TJ-Everett: This should not be a magic number and in my opinion we (SUPER_ADMINs)  should be able to set this threshold for their ecosystem.

src/interfaces/IErrors.sol
```diff
@@ -97,6 +97,7 @@ interface IPermissionModifierErrors {
     error NotAppAdministratorOrOwner();
     error NotSuperAdmin(address);
     error NotRuleAdministrator();
+    error CannotRenounceIfOnlyOneAdmin();
 }
```
TJ-Everett, line 101: If we change to a threshold variable that is settable, we may want to change this to CannotRenounceIfBelowMinAdminThreshold() or something similar.

test/AppManager.t.sol
```diff
@@ -36,9 +37,45 @@ contract AppManagerBaseTest is TestCommon {
...
     }
 
-    function testRenounceSuperAdmin() public {
+    function testMigratingSuperAdmin() public {
+        address newSuperAdmin = address(0xACE);
+        switchToRiskAdmin();
+        /// first let's check that a non superAdmin can't propose a newSuperAdmin
+        vm.expectRevert();
+        applicationAppManager.proposeNewSuperAdmin(newSuperAdmin);
+        /// now let's porpose some superAdmins to make sure that only one will ever be in the app
```
TJ-Everett: propose*

test/AppManager.t.sol
```diff
@@ -36,9 +37,45 @@ contract AppManagerBaseTest is TestCommon {
...
         switchToSuperAdmin();
-        applicationAppManager.renounceRole(SUPER_ADMIN_ROLE, superAdmin);
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 0);
+        applicationAppManager.proposeNewSuperAdmin(address(0x666));
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        applicationAppManager.proposeNewSuperAdmin(address(0x420));
```
TJ-Everett: We shouldn't use meme addresses to this level, in my opinion (We should assume everything we add to the repo is a public reflection of Thrackle. Sorry to be fun police here)

test/AppManager.t.sol
```diff
@@ -36,9 +37,45 @@ contract AppManagerBaseTest is TestCommon {
...
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        applicationAppManager.proposeNewSuperAdmin(newSuperAdmin);
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        /// now let's confirm it, but let's make sure only the porposed
```
TJ-Everett: proposed

test/ApplicationAppManager.t.sol
```diff
@@ -74,9 +75,42 @@ contract ApplicationAppManagerTest is TestCommon {
...
     }
 
-    function testRenounceSuperAdmin() public {
+    function testMigratingSuperAdmin() public {
+       address newSuperAdmin = address(0xACE);
+        switchToRiskAdmin();
+        /// first let's check that a non superAdmin can't propose a newSuperAdmin
+        vm.expectRevert();
+        applicationAppManager.proposeNewSuperAdmin(newSuperAdmin);
+        /// now let's porpose some superAdmins to make sure that only one will ever be in the app
```
TJ-Everett: propose

test/ApplicationAppManager.t.sol
```diff
@@ -74,9 +75,42 @@ contract ApplicationAppManagerTest is TestCommon {
...
         switchToSuperAdmin();
-        applicationAppManager.renounceRole(SUPER_ADMIN_ROLE, superAdmin);
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 0);
+        applicationAppManager.proposeNewSuperAdmin(address(0x666));
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        applicationAppManager.proposeNewSuperAdmin(address(0x420));
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        applicationAppManager.proposeNewSuperAdmin(newSuperAdmin);
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        /// now let's confirm it, but let's make sure only the porposed
```
TJ-Everett: Proposed

test/ApplicationAppManager.t.sol
```diff
@@ -74,9 +75,42 @@ contract ApplicationAppManagerTest is TestCommon {
...
+        assertTrue(applicationAppManager.isSuperAdmin(newSuperAdmin));
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 0);
+
+        vm.expectRevert("Function disabled");
+        applicationAppManager.grantRole("Oscar", address(0x123));
+
+        // let's check that newSuperAdmin can in fact do superAdmin stuff
+        applicationAppManager.addAppAdministrator(address(0xB0b));
+        applicationAppManager.revokeRole(APP_ADMIN_ROLE,address(0xB0b));
     }
```
TJ-Everett: Is the reason to have this test duplicated in two test files in order too ensure the Application contract inherits the RBAC structure correctly?

TJ-Everett (changes requested):
This is really good Oscar. I have a suggestion on the threshold amount for admins and a few typos but others great work. This is exactly what the roles needed.

src/application/AppManager.sol
```diff
@@ -97,6 +100,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+     * @notice this is purposely going to fail every time it will be invoked in order to force users to only use the appropiate 
+     * channels to grant roles, and therefore enforce the special rules in an app.
+     */
+     function grantRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
+        /// this is done to funnel all the role granting functions through the app manager functions since
+        /// the superAdmins could add other superAdmins through this back door
+        role;
+        account;
+        revert("Function disabled");
+    }
```
ShaneDuncan602, line 116: This is good. We need to control access to these roles to only be our allowed channels.

src/application/AppManager.sol
ShaneDuncan602: I agree about the magic number. It should be a constant. As for if they can set it, it was decided that there should only be 1. That was to appease ToB.

oscarsernarosero: That makes sense. The check is for any admin tho (except for PROPOSED_SUPER_ADMIN_ROLE).

src/application/AppManager.sol
```diff
@@ -107,6 +152,40 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+     * @dev Propose a new super admin. Restricted to super admins.
+     * @param account address to be added
+     */
+    function proposeNewSuperAdmin(address account) external onlyRole(SUPER_ADMIN_ROLE) {
+        if(account == address(0)) revert ZeroAddress();
+        /// we should only have 1 proposed superAdmin. If there is one already in this role, we should remove it to replace it.
+        if(getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE) > 0){
+            revokeRole(PROPOSED_SUPER_ADMIN_ROLE, getRoleMember(PROPOSED_SUPER_ADMIN_ROLE, 0));
+        }
+        super.grantRole(PROPOSED_SUPER_ADMIN_ROLE, account);
```
ShaneDuncan602, line 163: This was a good way to fix this. The PROPOSED_SUPER_ADMIN_ROLE is an elegant solution.

src/application/AppManager.sol
```diff
@@ -107,6 +152,40 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
+    /**
+     * @dev Propose a new super admin. Restricted to super admins.
+     * @param account address to be added
+     */
+    function proposeNewSuperAdmin(address account) external onlyRole(SUPER_ADMIN_ROLE) {
+        if(account == address(0)) revert ZeroAddress();
+        /// we should only have 1 proposed superAdmin. If there is one already in this role, we should remove it to replace it.
+        if(getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE) > 0){
+            revokeRole(PROPOSED_SUPER_ADMIN_ROLE, getRoleMember(PROPOSED_SUPER_ADMIN_ROLE, 0));
+        }
```
ShaneDuncan602, line 162: This is better than reverting and forcing them to revoke the proposed super. Good idea.

test/AppManager.t.sol
oscarsernarosero: you just have to suck the fun out everything, don't you. Jk. You're right

src/application/AppManager.sol
oscarsernarosero: Thank you Shane

test/AppManager.t.sol
TJ-Everett: 🚔

src/interfaces/IErrors.sol
ShaneDuncan602: Agreed. Something generic would probably be better.

test/AppManager.t.sol
ShaneDuncan602: Agreed. Probably better to use less risque numbers.

test/ApplicationAppManager.t.sol
```diff
@@ -74,9 +75,42 @@ contract ApplicationAppManagerTest is TestCommon {
...
+        /// first let's check that a non superAdmin can't propose a newSuperAdmin
+        vm.expectRevert();
+        applicationAppManager.proposeNewSuperAdmin(newSuperAdmin);
+        /// now let's porpose some superAdmins to make sure that only one will ever be in the app
         switchToSuperAdmin();
-        applicationAppManager.renounceRole(SUPER_ADMIN_ROLE, superAdmin);
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 0);
+        applicationAppManager.proposeNewSuperAdmin(address(0x666));
+        assertEq(applicationAppManager.getRoleMemberCount(PROPOSED_SUPER_ADMIN_ROLE), 1);
+        applicationAppManager.proposeNewSuperAdmin(address(0x420));
```
ShaneDuncan602: Same as above

test/ApplicationAppManager.t.sol
```diff
@@ -74,9 +75,42 @@ contract ApplicationAppManagerTest is TestCommon {
         assertEq(applicationAppManager.isAppAdministrator(superAdmin), true);
     }
 
-    function testRenounceSuperAdmin() public {
+    function testMigratingSuperAdmin() public {
+       address newSuperAdmin = address(0xACE);
+        switchToRiskAdmin();
+        /// first let's check that a non superAdmin can't propose a newSuperAdmin
+        vm.expectRevert();
```
ShaneDuncan602: We need to add the error to this. ToB flagged non-specific expectReverts as an issue.

test/ApplicationAppManager.t.sol
ShaneDuncan602: That was the original intention behind not getting rid of AppManager.t.sol. I don't think it's necessary to have this test in both but I don't think it will hurt either.

test/ApplicationAppManager.t.sol
```diff
@@ -127,13 +167,20 @@ contract ApplicationAppManagerTest is TestCommon {
 
         vm.stopPrank(); //stop interacting as the default admin
         vm.startPrank(user); //interact as a user
-
+        vm.expectRevert();
```
ShaneDuncan602: Can we add the error here?

test/ApplicationAppManager.t.sol
```diff
@@ -127,13 +167,20 @@ contract ApplicationAppManagerTest is TestCommon {
...
         applicationAppManager.revokeRole(APP_ADMIN_ROLE, address(77)); // try to revoke other app administrator
     }
 
     /// Test renounce Application Administrators role
     function testRenounceAppAdministrator() public {
-        switchToAppAdministrator(); // create a app administrator and make it the sender.
+        switchToSuperAdmin(); 
+        applicationAppManager.revokeRole(APP_ADMIN_ROLE,superAdmin);
+        switchToAppAdministrator(); 
+        vm.expectRevert();
```
ShaneDuncan602: Specific error?

test/ApplicationAppManager.t.sol
```diff
@@ -270,25 +317,36 @@ contract ApplicationAppManagerTest is TestCommon {
         assertEq(applicationAppManager.isRiskAdmin(address(88)), false);
         vm.stopPrank(); //stop interacting as the app administrator
         vm.startPrank(riskAdmin); //interact as the created risk admin
+        vm.expectRevert(abi.encodeWithSignature("CannotRenounceIfOnlyOneAdmin()"));
```
ShaneDuncan602: I'm not sure we wanted to restrict them from having 0 RiskAdmins and AccessTierAdmins. Wasn't it just for SuperAdmins and AppAdmins? Maybe we should get clarification on this from @jjbailey-io

ShaneDuncan602 (changes requested):
Just a few things and a question for JB.

test/ApplicationAppManager.t.sol
jjbailey-io: I agree, we shouldn't enforce >0 Risk/AccessTier admins. I think if they aren't using that feature they shouldn't be required to populate it. We wanted this implemented for Super Admins because it's detrimental and there is no way to recover the system if that happens. Risk/AccessTier admins could be re-added by the AppAdmins if they lose access.

test/ApplicationAppManager.t.sol
oscarsernarosero: Yeah I just did it in both tests, but Im not sure why we have two tests for the appManager

test/ApplicationAppManager.t.sol
oscarsernarosero: ok. So the restriction is only for super admins and app admins? The document related to the ticket does mention all admin roles, but Im ok to limit the check to those admins. Just to make sure we are making an informed decision here, this doesn't mean necessarily that risk or tier admins have to have at least 1 admin since the requirement is only enforced when renouncing/revoking which means that if you never set an admin then this will never apply to your app. It only becomes relevant when you have already added accounts to these roles and then you want to renounce/revoke.

TJ-Everett: I too want to clarify. So we would have:
Minimum of 1 in these roles always: 
SUPER_ADMIN
APP_ADMIN

No minimum number of admins: 
ACCESS_TIER_ADMIN
RISK_ADMIN
RULE_ADMIN
?

test/ApplicationAppManager.t.sol
oscarsernarosero: added

test/ApplicationAppManager.t.sol
oscarsernarosero: added

test/ApplicationAppManager.t.sol
oscarsernarosero: added

test/ApplicationAppManager.t.sol
jjbailey-io: I don't think we should enforce a minimum for any admin accounts except super admin. The minimum check on super admins is to ensure we never get into a state where we can't add/remove new admins. This is still an issue if we lose access to keys, but we're assuming we're not losing access to accounts for this. So I'm proposing:
- Minimum of 1 in these roles always:
    - SUPER_ADMIN
- No minimum number of admins:
    - APP_ADMIN
    - ACCESS_TIER_ADMIN
    - RISK_ADMIN
    - RULE_ADMIN

As long as the developer has access to SUPER_ADMIN they should be able to recover access to the other roles. Thoughts?

ShaneDuncan602: Technically, we wouldn't need to enforce a minimum on AppAdmin either. They wouldn't be able to really do anything in their assets but that's on them. They would be fine as long as the SuperAdmin exists and is usable. If they lose their SuperAdmin, they are in big trouble unless they have all the admins defined that they need.

oscarsernarosero: Ok. Only superAdmins have such restriction now.

ShaneDuncan602 (approved):
Looks great!

TJ-Everett (approved):
Approved. Looks great Oscar, well done.

PR #163, opened 2023-10-18
Refactor Risk Rules

TJ-Everett:
CheckAccBalanceByRisk
checkMaxTxSizePerPeriodByRisk
checkTransactionLimitByRiskScore
Rules accept match length array inputs for risk score & max size/balance Refactored associated tests


src/economic/ruleProcessor/RuleProcessorCommonLib.sol
```diff
@@ -50,4 +50,21 @@ library RuleProcessorCommonLib {
     function checkMaxTags(bytes32[] memory _tags) internal pure {
         if (_tags.length > MAX_TAGS) revert MaxTagLimitReached();
     }
+
+    function retrieveRiskScoreMaxSize(uint8 _riskScore, uint8[] memory _riskLevels, uint48[] memory _maxSizes) internal pure returns(uint256){
+        uint256 maxSize;
+        for (uint256 i; i < _riskLevels.length;) {
```
oscarsernarosero: _i_ should be initialized to 1 here since 0 is already checked in the caller functions.

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
```diff
@@ -50,4 +50,21 @@ library RuleProcessorCommonLib {
...
+            if (_riskScore < _riskLevels[i]) {
+                maxSize = uint(_maxSizes[i - 1]) * (10 ** 18);
```
oscarsernarosero, line 58: not sure why the `i - 1` here.

oscarsernarosero (changes requested):
Looks great Tayler. I only have one change request and a question.

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
ShaneDuncan602: agreed

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
ShaneDuncan602: Is that because the if statement is for < rather than <=?

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
TJ-Everett: i'll change - great catch!

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
TJ-Everett: This is because the index it checks if it is less than would then check that index 
i.e. 
0-24  NO Limit 
25 - 49 100 limit 
50 -74 50 limit 
75 - 99 10 limit 

If the risk score is 40 the "less than" index is 50 but they should not have a limit of just 50 they should have the 100 
so the index of their max value is i -1 

at least that is my reasoning behind and how the tests would match up

ShaneDuncan602: That makes sense to me.

ShaneDuncan602 (approved):
Good job.

src/economic/ruleProcessor/RuleProcessorCommonLib.sol
oscarsernarosero: Ok yeah. That makes sense.

oscarsernarosero (approved):
Looks great Tayler. Very clean job.

PR #165, opened 2023-10-18
Tron build fix

bfcoursewool:
Tron container was building correctly but failing with a familiar old error when the task was actually running against Anvil... trying to tweak the dockerfile / forge setup script here to see if I can fix it.


bfcoursewool:
this was a mistaken PR

PR #168, opened 2023-10-25
Se 352 foundry update for cheatcodes

oscarsernarosero:
### What was done

Although the ticket was about researching on upgrading to Foundry 2.0, it was really about being able to use the latest cheat codes offered by Foundry. Since it was found that we are already using Foundry 2.0, then it was discovered that the issue was that the forge-std library was really behind in its version which is where the cheat codes live. So the library was updated by doing the following steps from the root directory of Tron:

```
rm -rf lib/forge-std
git add -A
git commit -m "removed lib/forge-std"
forge install foundry-rs/forge-std
```

These steps are not necessary to do locally since it was already pushed to this branch.

A simple `vm.sleep(ms)` was added to the ApplicationDeployAll file to demonstrate it is now possible to use this cheat code.

### Important

It is necessary to build the project for this changes to take effect:

```
forge build
```


src/example/script/ApplicationDeployAll.s.sol
```diff
@@ -25,6 +25,7 @@ contract ApplicationDeployAllScript is Script {
     uint128[7] yieldPerTimeUnitArray = [1, 60, 3_600, 86_400, 604_800, 2_592_000, 31_536_000];
     uint128[7] yieldPerTimeUnitArray2 = [2, 120, 7_200, 172_800, 1_209_600, 5_184_000, 63_072_000];
     address[] applicationNFTAddresses;
+    uint256 PERIOD_BETWEEN_TX_BATCHES_MS = 36_000; // 36 seconds ~2.4 blocks
```
TJ-Everett, line 28: This is great and will be very helpful in deploying the diamond contracts if we change architecture.

TJ-Everett (approved):
Looks great Oscar, sorry about the confusion on the ticket description!

ShaneDuncan602 (approved):
Awesome

PR #169, opened 2023-10-25
Se 356 create a markdown for the minimum account balance by date rule

oscarsernarosero:


docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+This rule works at a token level which means that it has to be activated and configured in each token handler it is desired to enforce this rule.
+
+## Data Structure
+
+Due to the tagged nature of this rule, you can think of it as a collection of rules, where each "sub-rule" is totally independent from each other, and where each "sub-rule" is indexed by its tag. Therefore, a minumum-account-balance-by-date "sub-rule" is composed of 3 variables:
+
+- Hold amount (uint256).
+- hold period (uint16).
+- starting timestamp (uint64).
+
```
TJ-Everett, line 23: These should either all start with a capitol letter or none should. My vote is for capitols.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+    - This rule can only be updated in the asset handler by a **rule administrator**.
+
+## Rule Processing
+
+The rule will be evaluated in the following way:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags alongside the id of the minimum-balance-by-date rule set in the token handler. 
+3. The processor then will try to retrieve the sub-rule associated with each tag of the account.
+4. The processor will evaluate if such sub-rule's period is still active. If it is, the processor will then evaluate if the final balance of the account will be less than the minimum balance set in the sub-rule after the transaction is completed. If it will be in fact less than the minimum, then the transaction will revert.
```
TJ-Everett: If the final balance after transaction is less than the minimum, the transaction will revert. *

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+This rule doesn't require of any data to be recorded.
+
+### Events
+
+No events are emitted in this rule.
+
+### Dependencies
+
+This rule doesn't have any dependency.
```
TJ-Everett: Would the rule be dependent on the account being tagged as well as the dates for the rule? IDK how we want to define dependencies for rules vs other sections of the protocol.

TJ-Everett (commented):
Just a few spelling suggestions and a question. Otherwise I like this a lot Oscar.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
+# Minimum Account Balance By Date Rule
+
+## Purpose
+
+The purpose of this rule is to avoid ERC20 token holders to flood the market with their new tokens since this could effectively drive the supply up dramatically and, with it, cause a token price crash. In other words, this rule attempts to cool down traders by making them wait for some periods of time depending on their tags before they can sell their tokens. Due to the tagged nature of this rule, accounts have to wait for different periods of time depending on their tags, or even no time at all.
```
anne-thrackle:
```suggestion
The purpose of the minimum-balance-by-date rule is to prevent ERC20 token holders from rapidly flooding the market with newly acquired tokens. A dramatic increase in supply over a short time frame can cause a token price crash; this rule mitigates this scenario. This rule attempts to cool down traders by making them wait some period of time before they can sell their new tokens. The length of time depends on the account's tags. Different accounts may need to wait different periods of time depending on their tags, or even no time at all.
```
What are tags? 
I'm not sure what is meant by "cool down".

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Tokens Supported
+
+- ERC20
+
+## Scope 
+
+This rule works at a token level which means that it has to be activated and configured in each token handler it is desired to enforce this rule.
```
anne-thrackle:
```suggestion
This rule works at a token level. It must be activated and configured for each token desired in the corresponding token handler.  
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Data Structure
+
+Due to the tagged nature of this rule, you can think of it as a collection of rules, where each "sub-rule" is totally independent from each other, and where each "sub-rule" is indexed by its tag. Therefore, a minumum-account-balance-by-date "sub-rule" is composed of 3 variables:
```
anne-thrackle: I might be getting this wrong since I'm guessing what "tags" are. 
```suggestion
As this is a tag-based rule, you can think of it as a collection of rules, where all "sub-rules" are independent from each other, and where each "sub-rule" is indexed by its tag. A minumum-balance-by-date "sub-rule" is specified by 3 variables:
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Hold amount (uint256).
```
anne-thrackle:
```suggestion
- hold amount (uint256).
```

anne-thrackle: might be useful here to briefly say what these variables are for. Particularly for "hold amount", it isn't obvious to me what it means. From reading the "Rule Processing" section at the end I understand it's the minimum balance the account has to maintain.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+    /// ******** Minimum Balance By Date ********
+    struct MinBalByDateRuleS {
+        /// ruleIndex => userTag => rules
+        mapping(uint32 => mapping(bytes32 => ITaggedRules.MinBalByDateRule)) minBalByDateRulesPerUser;
+        uint32 minBalByDateRulesIndex; /// increments every time someone adds a rule
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+There is no limit in the amount of sub-rules that a rule can have other than at least one sub-rule.
```
anne-thrackle:
```suggestion
A minimum-balance-by-date rule must have at least one sub-rule. There is no maximum number of sub-rules. 
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Rule Registering
```
anne-thrackle:
```suggestion
### Registering the Rule
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+function addMinBalByDateRule(
+        address _appManagerAddr,
+        bytes32[] calldata _accountTags,
+        uint256[] calldata _holdAmounts,
+        uint16[] calldata _holdPeriods,
+        uint64[] calldata _startTimestamps
+    ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+
+The registering function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges.
```
anne-thrackle: is this true across all rules?

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+        bytes32[] calldata _accountTags,
+        uint256[] calldata _holdAmounts,
+        uint16[] calldata _holdPeriods,
+        uint64[] calldata _startTimestamps
+    ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+
+The registering function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges. 
+
+The function will return the protocol id of the rule.
```
anne-thrackle:
```suggestion
The registering function will return the protocol id of the rule.
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+    - This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+    - This rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
+
+- **Configuration and Enabling/disabling**:
+    - This rule can only be configured in the protocol by a **rule administrator**.
+    - This rule can only be set in the asset handler by a **rule administrator**.
+    - This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+    - This rule can only be updated in the asset handler by a **rule administrator**.
+
+## Rule Processing
```
anne-thrackle: I wonder if this section should be higher up in the doc, like maybe after the Data Structure section? I think I'd want to understand this before understanding how to register it. 

Also debating whether I like the wording, vs. something like "Rule Logic" or "What this Rule Does"

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+
+The rule will be evaluated in the following way:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags alongside the id of the minimum-balance-by-date rule set in the token handler.
```
anne-thrackle:
```suggestion
2. The processor will receive these tags along with the ID of the minimum-balance-by-date rule set in the token handler. 
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+    - This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+    - This rule can only be updated in the asset handler by a **rule administrator**.
+
+## Rule Processing
+
+The rule will be evaluated in the following way:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags alongside the id of the minimum-balance-by-date rule set in the token handler. 
+3. The processor then will try to retrieve the sub-rule associated with each tag of the account.
```
anne-thrackle:
```suggestion
3. The processor will then try to retrieve the sub-rule associated with each tag.
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
anne-thrackle:
```suggestion
4. The processor will evaluate whether each sub-rule's hold period is still active (if the current time is within `hold period` from the `starting timestamp`). If it is, the processor will then evaluate if the final balance of the account would be less than the `hold amount` if the transaction completed. If yes, then the transaction will revert.
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,114 @@
...
+4. The processor will evaluate if such sub-rule's period is still active. If it is, the processor will then evaluate if the final balance of the account will be less than the minimum balance set in the sub-rule after the transaction is completed. If it will be in fact less than the minimum, then the transaction will revert.
+5. The step 4 is repeated for each tag of the account.
```
anne-thrackle:
```suggestion
5. Step 4 is repeated for each of the account's tags. 
```

anne-thrackle (commented):
Overall this was quite good, I left some phrasing suggestions for clarity.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
anne-thrackle: just noticed in https://github.com/thrackle-io/tron/blob/internal/docs/userGuides/rules/APP-RULE-DIRECTORY.md we refer to this as "Create Function" (I think it's the same thing?). Perhaps we should decide whether to use the words create/creation or register and be consistent (especially because register <> activate, right?)

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: Ok. I mostly accepted your suggestion. I deleted the "cool down" part (it means to calm down someone that is in a excited or frantic stage). Interesting to see you using the ";". I replaced it with a simple "." tho. I think it is not used anymore here. Don't ask me why.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: Yes. Great catch!

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: it is.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+function addMinBalByDateRule(
+        address _appManagerAddr,
+        bytes32[] calldata _accountTags,
+        uint256[] calldata _holdAmounts,
+        uint16[] calldata _holdPeriods,
+        uint64[] calldata _startTimestamps
+    ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+
+The registering function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges.
```
anne-thrackle:
```suggestion
The create function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges. 
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+        bytes32[] calldata _accountTags,
+        uint256[] calldata _holdAmounts,
+        uint16[] calldata _holdPeriods,
+        uint64[] calldata _startTimestamps
+    ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+
+The registering function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges. 
+
+The registering function will return the protocol ID of the rule.
```
anne-thrackle:
```suggestion
The create function will return the protocol ID of the rule.
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+3. The processor will then try to retrieve the sub-rule associated with each tag.
+4. The processor will evaluate whether each sub-rule's hold period is still active (if the current time is within `hold period` from the `starting timestamp`). If it is, the processor will then evaluate if the final balance of the account would be less than the `hold amount` in the case of the transaction succeeding. If yes, then the transaction will revert.
+5. Step 4 is repeated for each of the account's tags. 
+
+###### *see [IRuleStorage](../../../src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol) -> checkMinBalByDatePasses*
+
+
+## Create Function
+
+Registering a minimum-balance-by-date rule is done through the function:
```
anne-thrackle:
```suggestion
Adding a minimum-balance-by-date rule is done through the function:
```
or "creating"

anne-thrackle:
```suggestion
Adding a minimum-balance-by-date rule is done through the function:
```
or "creating"

TJ-Everett (approved):
Looks great Oscar

anne-thrackle (approved):
looks good to me at far as content, clarity, organization. Not being that familiar with the logic itself I'm not confirming the correctness here but someone else can do that

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
+# Minimum Account Balance By Date Rule
+
+## Purpose
+
+The purpose of the minimum-balance-by-date rule is to prevent ERC20 token holders from rapidly flooding the market with newly acquired tokens since a dramatic increase in supply over a short time frame can cause a token price crash. This rule attempts to mitigates this scenario by making holders wait some period of time before they can sell their tokens. The length of time depends on the account's tags. Different accounts may need to wait different periods of time depending on their tags, or even no time at all.
```
ShaneDuncan602: This can also be applied to 721's. The intention there would be to force holders to keep the asset for a specified minimum amount of time. This could be for earned assets or achievements so they aren't immediately traded.

"mitigates" should be "mitigate". 

We should probably say, "before they can transfer their tokens" instead of "sell their tokens"

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Tokens Supported
+
+- ERC20
```
ShaneDuncan602, line 9: Also ERC721

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+As this is a [tag](../GLOSSARY.md)-based rule, you can think of it as a collection of rules, where all "sub-rules" are independent from each other, and where each "sub-rule" is indexed by its tag. A minumum-balance-by-date "sub-rule" is specified by 3 variables:
+
```
ShaneDuncan602, line 19: Shouldn't the tags parameter be here as well?

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+- **Hold period** (uint16): The amount of hours to hold the tokens.
+- **Starting timestamp** (uint64): The timestamp of the date when the *hold period* starts counting.
+
+```c
+/// ******** Minimum Balance By Date Rules ********
+    struct MinBalByDateRule {
+        uint256 holdAmount; /// token units
+        uint16 holdPeriod; /// hours
+        uint256 startTimeStamp; /// start
+    }
```
ShaneDuncan602, line 30: We need to be careful about how we define this for clients. This struct is on a per-tag basis so we should make that clear or it might get confusing.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+```
+###### *see [RuleDataInterfaces](../../../src/economic/ruleStorage/RuleDataInterfaces.sol)*
+
+Additionally, each one of these data structures will be under a tag (bytes32):
```
ShaneDuncan602, line 34: Ok, I see that distinction here. I'm not sure that it's clear though. We need to set a good example for this now because it will repeat quite a bit.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+A minimum-balance-by-date rule must have at least one sub-rule. There is no maximum number of sub-rules.
+
+## Role Applicability
+
+- **Evaluation Exceptions**: 
+    - This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+    - This rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
```
ShaneDuncan602: Only for ERC20 transfers

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+        uint256 startTimeStamp; /// start
+    }
+```
+###### *see [RuleDataInterfaces](../../../src/economic/ruleStorage/RuleDataInterfaces.sol)*
+
+Additionally, each one of these data structures will be under a tag (bytes32):
+
+ tag -> sub-rule.
+
+ ```javascript
```
ShaneDuncan602: Is there a reason we are labelling some code snippets as javascript or C?

ShaneDuncan602 (changes requested):
Some small suggestions and additions. Each one of these keeps getting better than the last.

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags along with the ID of the minimum-balance-by-date rule set in the token handler. 
+3. The processor will then try to retrieve the sub-rule associated with each tag.
+4. The processor will evaluate whether each sub-rule's hold period is still active (if the current time is within `hold period` from the `starting timestamp`). If it is, the processor will then evaluate if the final balance of the account would be less than the `hold amount` in the case of the transaction succeeding. If yes, then the transaction will revert.
+5. Step 4 is repeated for each of the account's tags.
```
anne-thrackle, line 79: You can take or leave these comments. 

I was a little bothered by the phrasing "the account will pass the tags to the protocol" which made it seem like the account could send false information about its tags (which I'm sure is not the case in practice). 

I also wondered what the point of the processor receiving the ID of the rule was, since it wasn't mentioned as being used in the rest of the steps. Should it be that it retrieves the sub-rule for each tag _for that rule ID_? 

When you wrote "the process will try to retrieve..." - is it important to include here what happens if it fails to retrieve them? Or can we just say "it retrieves" and the failure case is obvious/implied (i.e. the transaction fails). 

```suggestion
1. The processor receives the following information: 
    - for the account being evaluated, the tags registered to its address in the application manager. 
    - the ID of the minimum-balance-by-date rule set in the token handler. 
2. The processor retrieves the sub-rule associated with each tag.
3. The processor evaluates whether each sub-rule's hold period is still active (if the current time is within `hold period` from the `starting timestamp`). If it is, the processor then evaluates whether the final balance of the account would be less than the `hold amount` in the case of the transaction succeeding. If yes, then the transaction will revert.
4. Step 3 is repeated for each of the account's tags. 
```

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: yes. It's because it doesn't recognize solidity, so I try to play with it to have the most amount of colors

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: ah ok. I got confused by the rule being processed in the ERC20TaggedRules

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: that's in the Create Function section

docs/userGuides/rules/MINIMUM-ACCOUNT-BALANCE-BY-DATE.md
oscarsernarosero: I honestly think I made it very clear since I spend the rest of the section explaining just that.

ShaneDuncan602 (approved):
Looks great!

jjbailey-io (approved):

PR #170, opened 2023-10-26
Se 381 research ERC721C and implement an example integration in tron

TJ-Everett:
DO NOT MERGE BRANCH 

This branch serves as a POC for the ERC721C and ERC721CW integration into the protocol. This is a rough first pass to ensure this was even possible. Many more iterations and improvements should be made prior to full integration into the external branch. 
Additional gas testing should also be conducted to fully understand the impacts of this contract suite compared to the current versions of ERC721 the protocol is utilizing

DO NOT MERGE BRANCH


src/example/ERC721/NonProtocolERC721.sol
```diff
@@ -0,0 +1,132 @@
...
+    /// Base Contract URI
+    string public baseUri;
+
+    /// Treasury Address
+    address private proposedTreasury;
+    address payable private treasury;
+
+    /// errors
+    error MintFeeNotReached();
+    error PriceNotSet();
```
oscarsernarosero: These errors suggest you need to buy the NFTs, but that's not the case here. It's kinda confusing.

src/example/ERC721/NonProtocolERC721.sol
```diff
@@ -0,0 +1,132 @@
...
+    error CannotWithdrawZero();
+    error TreasuryAddressCannotBeTokenContract();
+    error TreasuryAddressNotSet();
```
oscarsernarosero: Same with these and others down below. It suggests you have withdrawal functions and a treasury, but there is non of these.

src/example/StakingWrapper.sol
```diff
@@ -0,0 +1,109 @@
...
+     * @param from sender address
+     * @param to recipient address
+     * @param tokenId Id of token to be transferred
+     * @param batchSize the amount of NFTs to mint in batch. If a value greater than 1 is given, tokenId will
+     * represent the first id to start the batch.
+     */
+    function _beforeTokenTransfer(address from, address to, uint256 tokenId, uint256 batchSize) internal override {
+        // Rule Processor Module Check
+        if(to != address(this) ||  from != address(this)) {
+        require(handler.checkAllRules(from == address(0) ? 0 : balanceOf(from), to == address(0) ? 0 : balanceOf(to), from, to, batchSize, tokenId, ActionTypes.TRADE));
```
oscarsernarosero, line 96: Not relevant for this PR, but Im just noticing we don't have this in the ERC20. Might be worth it revisiting this later.

src/token/ERC721/ProtocolERC721.sol
```diff
@@ -11,14 +11,15 @@ import "../ProtocolTokenCommon.sol";
...
+import "@limitbreak/creator-token-contracts/contracts/erc721c/ERC721C.sol";
 
 /**
  * @title ERC721 Base Contract
  * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett
  * @notice This is the base contract for all protocol ERC721s
  */
 
-contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon, AppAdministratorOrOwnerOnly {
+contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon, AppAdministratorOrOwnerOnly, ERC721C {
```
oscarsernarosero: So we are using now ERC721C as our default and official ERC721 token? I think we should have ProtocolERC721 and ProtocolERC721C since the C is heavier in gas consumption.

test/ERC721/ApplicationERC721.t.sol
oscarsernarosero, line 1: I would say, you need some dedicated tests for the ERC721C in this file to test the functionalities.

test/ERC721/ERC721CGasTest.t.sol
oscarsernarosero, line 1: For what I know about the ERC721C, you can set some operator whitelists and you can have some security configurations (restrict NFT receivers to be only EOAs or contracts with no code, or senders cannot be contracts, etc). I think the gas should be analyzed with those configurations on since that's most likely how this is going to be used.

src/token/ERC721/ProtocolERC721.sol
oscarsernarosero: ok never mind. We are not merging this branch so you're good.

oscarsernarosero (commented):
This is really good in general. Im not sure if this within the scope of the ticket, but since you are already doing some tests in here, I think some tests on the functionality of the ERC721C should be added. You can demonstrate how to configure the operator whitelists and some of the security configurations, for example. I think a gas analysis with those configurations on would be much more realistic than without them.

test/ERC721/ApplicationERC721.t.sol
TJ-Everett: I agree that we should test out functionality more than this, however I think that should be an additional ticket. Since this was just to ensure protocol integration. They have some really interesting royalty payment splitting going on also that we should further test if we are going to fully integrate this version of ERC721.

src/example/ERC721/NonProtocolERC721.sol
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * @dev Constructor sets the name, symbol and base URI of NFT along with the App Manager and Handler Address
+     * @param _name Name of NFT
+     * @param _symbol Symbol for the NFT
+     * @param _baseUri URI for the base token
+     */
+    constructor(string memory _name, string memory _symbol, string memory _baseUri) ERC721(_name, _symbol) {}
+
+    /**
+     * @dev Function mints a new token to anybody. Don't enabled this function if you are not sure about what you're doing.
```
ShaneDuncan602: The second statement needs reworked

test/ERC721/ApplicationERC721.t.sol
ShaneDuncan602: I'm not sure if it's a full POC if the royalties aren't in here. We have proven a rule can be applied to ERC721C, but I think we also need to know for certain that a 2981 royalty can be enforced on a transfer. Would it be too much to add one of LimitBreak's simple royalty interfaces to this or would it increase the scope of this ticket too much?

test/ERC721/ApplicationERC721.t.sol
```diff
@@ -668,7 +668,7 @@ contract ApplicationERC721Test is TestCommonFoundry {
     function testTransferVolumeRuleNFT() public {
         /// set the rule for 40% in 2 hours, starting at midnight
         switchToRuleAdmin();
-        uint32 _index = RuleDataFacet(address(ruleStorageDiamond)).addTransferVolumeRule(address(applicationAppManager), 2000, 2, Blocktime, 0);
+        uint32 _index = RuleDataFacet(address(ruleStorageDiamond)).addTransferVolumeRule(address(applicationAppManager), 2000, 2, Blocktime, 10);
```
ShaneDuncan602: This was purposely set at 0 to test if the transfer volume rule could be used to turn a NFT into a soul bound token.

test/ERC721/ApplicationERC721.t.sol
ShaneDuncan602: Also, I think all tests related to 721C should be in separate test contracts. We would want to keep the current tests the same.

src/example/StakingWrapper.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+import ".././token/ERC721/ProtocolERC721.sol";
+import ".././token/ProtocolTokenCommon.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@limitbreak/creator-token-contracts/contracts/erc721c/extensions/AdventureERC721CW.sol";
```
ShaneDuncan602: Is there a reason we're using their AdventureERC721CW instead of the ERC721CW template?

ShaneDuncan602 (changes requested):
Really cool stuff. Just a few questions and suggestions.

src/example/ERC721/NonProtocolERC721.sol
TJ-Everett: This is from the free mint contract so I will also rework the statement in my documents ticket I am working so it gets merged to external and not lost. Great catch.

src/example/ERC721/NonProtocolERC721.sol
TJ-Everett: Yeah, I was using the mintFee at first and then switched to a regular free mint so I can clean these up.

test/ERC721/ApplicationERC721.t.sol
TJ-Everett: Gotchya, I will ensure this still works as intended or document why it doesn't in the write up.

src/example/StakingWrapper.sol
TJ-Everett: No great catch, I can inherit from the regular in the testing rework.

test/ERC721/ApplicationERC721.t.sol
TJ-Everett: I will rework the tests so that that we have a ProtocolERC721C and test suite associated to it. Since this will take some thought anyway, I can try and work in some of the royalty payment tests at the same time. I am going to time box myself to 1 working day and if I cannot get it done will spawn another ticket from this one to continue it.

oscarsernarosero (approved):
Looks good to me Tayler. Good job.

src/example/ERC721/NonProtocolERC721.sol
ShaneDuncan602: I think this was missed. It was the " Don't enabled this function if you are not sure about what you're doing."

TJ-Everett: I corrected this in the Docs branch I was working on - since this wasn't a merge branch that way it was corrected in the external branch. I will correct here as well to "Function mints a new token freely without restriction. This function should only be enabled if you are certain you want a free mint with no restrictions."

ShaneDuncan602 (approved):
Great job and a huge amount of effort on this. It looks really good.

TJ-Everett:
Closing PR as this was not a merge branch.

PR #171, opened 2023-10-27
first draft of the pause rule

oscarsernarosero:


docs/userGuides/rules/PAUSE-RULE.md
```diff
@@ -0,0 +1,94 @@
...
+    uint256 pauseStart;
+    uint256 pauseStop;
+}
+```
+###### *see [PauseRule](../../../src/data/PauseRule.sol)*
+
+## Role Applicability
+
+- **Evaluation Exceptions**: 
+    - This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
```
ShaneDuncan602, line 36: The second statement seems contradictory to the first.

docs/userGuides/rules/PAUSE-RULE.md
```diff
@@ -0,0 +1,94 @@
...
+    - This rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
```
ShaneDuncan602: I'm not sure this is the case when a 721 is transferred. Probably need to mention that it's only in the case of 20's.

docs/userGuides/rules/PAUSE-RULE.md
```diff
@@ -0,0 +1,94 @@
...
+This create function will also delete automatically any pause rule where the `pauseStop` timestamp is less than current timestamp  since this means that the rule has *expired*. 
+
+There is a default limit of 15 pause rules per application to avoid too much gas consumption during transactions.
+
+```
+uint8 constant MAX_RULES = 15;
+```
+###### *see [PauseRules](../../../src/data/PauseRules.sol)*
+
+It is worth noting that this rule is special in the sense that it is not stored in the protocol but in the AppManager data contracts. Therefore, this rule doesn't have an ID like the other rules.
```
ShaneDuncan602: We might want to mention that this fact makes it unshareable with other applications.

ShaneDuncan602 (changes requested):
Just a few clarifying things. I love this format.

docs/userGuides/rules/PAUSE-RULE.md
```diff
@@ -0,0 +1,94 @@
...
+
+This rule doesn't return any data.
+
+### Data Recorded
+
+This rule doesn't require of any data to be recorded.
+
+### Events
+
+- **PauseRuleEvent(uint256 indexed pauseStart, uint256 indexed pauseStop, bool indexed add)**: emitted when:
```
TJ-Everett, line 139: This raises a question: 
In the MinBalanceByDate rule doc we said there was no event emitted for the rule. 
There is for the creation but not for when the rule is checked. 
Do we want to include rule creation events in these docs? (my vote is yes)

TJ-Everett (commented):
Content and format are looking great, Just one question for me.

ShaneDuncan602 (approved):
Looks good.

PR #174, opened 2023-10-30
Se 360 markdown for token supply volatility rule

oscarsernarosero:
Introduced `Parameter` section in this document and also added it in the past docs:

PUASE RULES
NFT TRANSFER COUNTER
MIN BALANCE BY DATE


docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
+# Total Supply Volatility Rule
+
+## Purpose
+
+The prupose of this rule is to control a sudden increase or decrease in the supply of a token. This can help to prevent a sudden crash or spike in the price of a token due to sharp changes in the supply side, which can lead to severe damages to the health of the economy in the long run.
```
ShaneDuncan602: "prupose"

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+- **_totalSupply** (uint256): (optiona) if not 0, then this means that this value is going to be the locked value for totalSupply instead of the live token's totalSupply value at rule processing time.
+
+The create function will return the protocol ID of the rule.
+
+### Parameter Optionality:
+
+The parameters where developers have options are:
+
+- **_totalSupply**: developers can pass a non-zero value here which will mean that the rule will always check against this fix total supply value. If this value is set as zero, it will mean that the processor will expect a dynamic totalSupply value from the asset handler.
+
```
ShaneDuncan602: This could be simplified as "For volatility calculations, when this is zero, the token's totalSupply is used. Otherwise, this value is used as the total supply.

ShaneDuncan602 (changes requested):
One typo and a minor suggestion. These docs keep getting cleaner each time. Great job.

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+A total-supply-volatility rule is composed of 4 variables:
+
+- **Max Change** (uint16): The maximum percentual change allowed in the supply during a *period* of time (expressed in basis points).
```
anne-thrackle:
```suggestion
- **Max Change** (uint16): The maximum percent change allowed in the supply during a *period* of time (expressed in basis points).
```

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+- **Period** (uint16): The amount of hours that defines a period.
```
anne-thrackle: general question: why do we make our periods of times in hours? What if people wanted to use 30 min, they wouldn't be able to?

docs/userGuides/rules/SUPPLY-VOLATILITY.md
anne-thrackle:
```suggestion
The purpose of this rule is to prevent a sudden increase or decrease in the supply of a token. This can help prevent a sudden crash or spike in the price of a token due to sharp changes in the supply side. 
```
I suggest removing the last part of the sentence, sticking more to simple and unopinionated facts in this rule-level documentation. People might ask "why wouldn't I want a spike in price?" and I feel like that discussion belongs somewhere else... in another part of our product, where we're actually advising people on what a long-term healthy economy looks like. 

But you can keep it if you want, it's a subjective opinion here

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+- **Starting timestamp** (uint64): The timestamp of the date when the rule starts the first *period*.
+- **Total Supply** (uint256): The token's total supply.
```
anne-thrackle: total initial supply? targeted supply? In the context of this rule (which is about supply changing) I'm not sure what this is, why it isn't automatically known (i.e. supply of the token at Starting timestamp)

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+    - This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The asset handler keeps track of the total supply per period, the net supply increase/decrease per period, and the date of the last mint/burn transaction.
+2. The asset handler will first check if the transaction is a *minting* or *burning* operation. If it is, then it sends the aforementioned data to the rule processor with the rule Id, and the amount of tokens being minted/burned in the transaction.
+3. The rule processor will evaluate if the current transaction is in a new period or part of the same period as the last burning/minting transactions. If it is a new period, the net supply change for the period will be equal to the amount of tokens minted/burned in current transaction. Also, the totalSupply for the current period will be set to current value. If it is not a new period, then the net supply change for the period will accumulate the amount of tokens minted/burned in current transaction. The totalSupply for the current period will remain fixed.
```
anne-thrackle:
```suggestion
3. The rule processor will determine whether the current transaction is in a new period or part of the same period as the last burning/minting transaction and correspondingly calculate the totalSupply and net supply change. 
    - If it is in a new period, the net supply change for the period will be equal to the amount of tokens minted/burned in current transaction. Also, the totalSupply for the current period will be set to the token's current totalSupply value. 
    - If it is not a new period, then the net supply change for the period will accumulate the amount of tokens minted/burned in current transaction. The totalSupply for the current period will remain fixed. 
```
I think for readability of these paragraphs with a branching condition, it might be nice to split it out with bullet points or something.

anne-thrackle: not sure I did markdown correctly though

docs/userGuides/rules/SUPPLY-VOLATILITY.md
anne-thrackle: I read some more below and I do find it confusing that we're using the same term "total supply" to mean two things. I think perhaps we could consider a rename here

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+        uint64 _startTimestamp,
+        uint256 _totalSupply
+    ) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+###### *see [RuleDataFacet](../../../src/economic/ruleStorage/RuleDataFacet.sol)*
+
+### Parameters:
+
+- **_appManagerAddr** (address): the address of the application manager to verify that the caller has Rule administrator privileges.
+- **_maxVolumePercentage** (uint16): Maximum percentual change of supply allowed expressed in basis points (1 -> 0.01%; 100 -> 1.0%).
```
anne-thrackle:
```suggestion
- **_maxVolumePercentage** (uint16): Maximum percent change of supply allowed expressed in basis points (1 -> 0.01%; 100 -> 1.0%). 
```

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+```
+###### *see [RuleDataFacet](../../../src/economic/ruleStorage/RuleDataFacet.sol)*
+
+### Parameters:
+
+- **_appManagerAddr** (address): the address of the application manager to verify that the caller has Rule administrator privileges.
+- **_maxVolumePercentage** (uint16): Maximum percentual change of supply allowed expressed in basis points (1 -> 0.01%; 100 -> 1.0%). 
+- **_period** (uint16): amount of hours that defines a period.
+- **_startTimestamp** (uint64): Unix timestamp for the *_period*s to start counting.
+- **_totalSupply** (uint256): (optiona) if not 0, then this means that this value is going to be the locked value for totalSupply instead of the live token's totalSupply value at rule processing time.
```
anne-thrackle:
```suggestion
- **_totalSupply** (uint256): (optional) if not 0, then this is the value used for totalSupply instead of the live token's totalSupply value at rule processing time.
```

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+
+The create function will return the protocol ID of the rule.
+
+### Parameter Optionality:
```
anne-thrackle, line 108: I don't quite understand the purpose of this section. The developers can set the percent change and timestamp and period as well, so why is this parameter singled out as having more "optionality"?

docs/userGuides/rules/SUPPLY-VOLATILITY.md
```diff
@@ -0,0 +1,201 @@
...
+```c
+int256 private volumeTotalForPeriod;
+uint256 private totalSupplyForPeriod;
+```
+
+*see [ERC721Handler](../../../src/token/ERC721/ProtocolERC721Handler.sol)/[ERC20Handler](../../../src/token/ERC20/ProtocolERC20Handler.sol)*
+
+## Data Recorded
+
+This rule requires the record of the following information in the asset handler:
```
anne-thrackle: is the meaning "This rule requires recording the following information in the asset handler:"
or  "This rule records the following information in the asset handler:"

anne-thrackle (commented):
looks good. My main source of confusion was the totalSupply meaning.

docs/userGuides/rules/SUPPLY-VOLATILITY.md
TJ-Everett: optional*

docs/userGuides/rules/TRANSFER-COUNTER.md
```diff
@@ -102,6 +102,13 @@ The function will return the protocol id of the rule.
 
 ###### *see [RuleDataFacet](../../../src/economic/ruleStorage/RuleDataFacet.sol)*
 
+### Parameters:
+
+- **_appManagerAddr** (address): The address of the application manager to verify that the caller has Rule administrator privileges.
+- **_nftTypes** (bytes32[]): array of types (tags) that the NFT contract has registered in the application manager.
+- **_tradesAllowed** (uint8[]): the amount of trades allowed in a day.
+- **_startTs** (uint64): starting timestamp of the rule. This timestamp will determin the time that a day starts and ends for the rule processing. For example, *the amount of trades will reset to 0 very day at 2:30 pm.*
```
TJ-Everett: determine*

TJ-Everett (commented):
Looks great overall Oscar, just a few typos I found.

TJ-Everett:
While working on my current rule doc I noticed that this rule (supply vol) does not have the mapping from IRulesStorage.sol like the other tagged rules have. Was this intentional since this isn't a double mapping like the tagged rules are?

docs/userGuides/rules/SUPPLY-VOLATILITY.md
oscarsernarosero: love it

docs/userGuides/rules/SUPPLY-VOLATILITY.md
oscarsernarosero: I guess we thought it would be the best time unit to achieve the goal in the simplest way. You are correct. It is impossible to do 30-minute periods with current design. In our defense, volume and price are tracked usually in a 24-hour period, and in 1 hour in extreme conditions. I think going to 30 minutes would be something very rarely used.

docs/userGuides/rules/SUPPLY-VOLATILITY.md
oscarsernarosero: I see your point. I still think it is a good idea to leave the comment since this is the best place to communicate what this rule does and why you might want to use it. Rules are not mandatory to use, so devs that don't see value in controlling the supply can simply ignore it completely, but at least they know why they might want to enable it in their assets.

docs/userGuides/rules/SUPPLY-VOLATILITY.md
oscarsernarosero: ok. I changed it to:
`if not zero, this will represent always the token's total supply.`. I agree with you though. This rule is very confusing on its implementation, but here is an explenation. The token supply has 2 possible sources: in the rule itself (this above), and the actual token's totalSupply. If the rule has 0 as total supply, then it will require the totalSupply to be passed to the protocol, in which case the total-supply value needs to be recorded at the start of the period to be able to accurately calculate the change. Otherwise, we are going to always compare the supply change between current and last transaction.

docs/userGuides/rules/SUPPLY-VOLATILITY.md
oscarsernarosero: Having totalSupply here as 0 will be interpreted different and will trigger a different rule evaluation process than if it is not zero. Therefore, it is considered as an optioned parameter. When the devs pass any kind of values in the other parameters, they don't really trigger different actions. They are always interpreted in the same way.

anne-thrackle: Ah I see, that makes sense.

ShaneDuncan602:
> While working on my current rule doc I noticed that this rule (supply vol) does not have the mapping from IRulesStorage.sol like the other tagged rules have. Was this intentional since this isn't a double mapping like the tagged rules are?

I was thinking it's there. VolatilityRuleS?

ShaneDuncan602 (approved):
Looks really good.

anne-thrackle (approved):

PR #176, opened 2023-10-31
TOKEN-TRANSFER-VOLUME.md

TJ-Everett:


docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+A token-transfer-volume rule is composed of 4 components:
+
+- **Max Volume** (uint24): The maximum percent of total supply to be traded during the *period*.
```
oscarsernarosero: in percent basis units?

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule's period is still active (if the current time is within `period` from the `starting timestamp`).
```
oscarsernarosero: I think you're mixing 2 concepts here. This sounds more like a *rule-by-date* type of thing where you have an active period and then nothing. I think maybe something more like "The processor will evaluate whether the transaction is the first one in current period"... or something of the kind.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+    - **If it is not a new period**, the processor will then evaluate the `max volume`. 
+    - **If it is a new period**, the processor will return the `transfer volume` value increased from current trade for the handler to store. 
+5. The processor will then evaluate if the final volume percentage would be greater than the `max volume` in the case of the transaction succeeding. 
+    - If yes, then the transaction will revert. 
+    - If no, the processor will return the `transfer volume` for the current `period`.
+
+###### *see [IRuleStorage](../../../src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol) -> checkTokenTransferVolumePasses*
```
oscarsernarosero: see [ERC20RuleProcessorFacet]*

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
oscarsernarosero: I think I know what you're trying to do here. It does have a startTimestamp, so you do have to say this, but that's a different step before you start talking about transactions in a new period or not.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule's period is still active (if the current time is within `period` from the `starting timestamp`).
+    - **If it is not a new period**, the processor will then evaluate the `max volume`. 
+    - **If it is a new period**, the processor will return the `transfer volume` value increased from current trade for the handler to store.
```
oscarsernarosero: This is not very clear. I think you are missing steps in the "not-a-new-period" part.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+
+
+The create function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges. 
+
+The create function will return the protocol ID of the rule.
+
+### Parameters:
+
+- **_appManagerAddr** (address): The address of the application manager to verify that the caller has Rule administrator privileges.
+- **_maxVolumePercentage** (uint24): .
```
oscarsernarosero: Missing this one

oscarsernarosero (changes requested):
Looks really good, but some parts need more work.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
TJ-Everett: Hmmm, I disagree. 
The rule has to check the rule period and last transfer time to ensure it is within the current period or not. 
If it is in a new period then this transaction is simply added to the volume tracker 
if it is not a new period then the you move to the next step of volume calculations 

I don't think we should be too different with rule evaluation descriptions that are all using the same logic 
We even have common functions for isWithinPeriod() so to me those should all be nearly identical in the evaluation step.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
+# Token Transfer Volume Rule
+
+## Purpose
+
+The purpose of the token-transfer-volume rule is to reduce high trading volatility periods by allowing developers to set a volume percentage of the token's total supply that can be traded within a period. When the total traded percentage limit is reached, transfers are suspended until the next period begins.
```
anne-thrackle: phrasing suggestion, take it or leave it
```suggestion
The purpose of the token-transfer-volume rule is to reduce high trading volatility periods by allowing developers to set a maximum volume (as a percentage of the token's total supply) that can be traded within a period of time. When the trading volume maximum is reached, transfers are suspended until the next period begins. 
```

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+A token-transfer-volume rule is composed of 4 components:
+
+- **Max Volume** (uint24): The maximum percent in basis uints of total supply to be traded during the *period*.
+- **Period** (uint16): The amount of hours that defines a period..
```
anne-thrackle:
```suggestion
- **Period** (uint16): The amount of hours that defines a period.
```

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+- **Starting Timestamp** (uint64): The timestamp of the date when the *period* starts counting.
+- **Total Supply** (uint256): if not zero, this value will always be used as the token's total supply for rule evaluation. This can be used when the amount of circulating supply is much smaller than the amount of the token totalSupply due to some tokens being locked in a dev account or a vesting contract, etc. Only use this value in such cases.
+
+```c
+/// ******** Token Transfer Volume ********
+    struct TokenTransferVolumeRule {
+        uint24 maxVolume; // this is a percentage with 2 decimals of precision(2500 = 25%)
```
anne-thrackle:
```suggestion
        uint24 maxVolume; // this is a percentage with 2 decimals of precision (2500 = 25%)
```

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule's period is still active (if the current time is within `period` from the `starting timestamp`).
+    - **If it is not a new period**, the processor will then evaluate the `max volume`.
```
anne-thrackle: I find myself confused (throughout this doc) with what are percentages and what are straight up values. The paramter "Max Volume" defined on line 22 is a percentage in basis points. The "max volume" here is the actual maximum volume in token units? 

Maybe there's a different name we could use like "max volume percent" that would make this clearer. But I know it's wordier. (though this is what it's called in the creation function)

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,226 @@
...
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule's period is still active (if the current time is within `period` from the `starting timestamp`).
+    - **If it is not a new period**, the processor will then evaluate the `max volume`. 
+    - **If it is a new period**, the processor will return the `transfer volume` value increased from current trade for the handler to store.
```
anne-thrackle:
```suggestion
    - **If it is a new period**, the processor will return the `transfer volume` value, set to the current trade's size for the handler to store. 
```
is my understanding correct?

anne-thrackle (commented):
just one source of confusion for me, when things are percentages vs. raw values. Otherwise looks pretty good.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
TJ-Everett: I will definitely try to clarify throughout the doc where possible 
The max volume variable is the percent basis points and stays that way. 
transfer volume passed from the handler is a true value. 

The rule will take the true value of transfer volume (reset after each period with the first transaction of the new period) and calculate that against the totalSupply (or fixed supply provided to the rule) 
Then it takes that calculation and compares that it is not greater than the max volume percent in basis units

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
ShaneDuncan602: I would also consider this a rule by date. It has a period within the rule and the volatility is determined by checking the supply at the beginning of the period. 

I think we should also mention the total supply "snapshot" for when they use the token's totalSupply and not their own. It's unique enough that I always have to go look at the code to remember it.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,227 @@
...
+ ```c
+/// ******** Token Transfer Volume ********
+    struct TransferVolRuleS {
+        mapping(uint32 => INonTaggedRules.TokenTransferVolumeRule) transferVolumeRules;
+        uint32 transferVolRuleIndex;
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Role Applicability
```
ShaneDuncan602: I am questioning this "Role" here. For instance, line 50 doesn't pertain to a defined role. It's a registered account but not a role.

ShaneDuncan602 (changes requested):
Minor tweaks

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
TJ-Everett: Shane:
Point 1 - is that an ask for a change in the phrasing or an agreement in current phrasing?
Point 2 - is step 3 enough clarification or would you prefer it be more clear? 
I don't want to miss any needed clarity in these PRs as I have multiple at once

oscarsernarosero: yeah. It does kinda combine the rule by date and the rule per period logic. But that's not clear in your phrases. It seems like you start talking about one thing, but end talking about another one. I would split step 3 in 2: one where it checks if it is in an active stage (post startTimestamp) and then another one to check if the transaction is in a new period or not.

TJ-Everett: I promise I am not trying to be difficult here, but I do not understand how this is different than the min-max-bal-by-date logic for this specific step 
This phrasing was pulled directly from that approved doc and used here. Since this is the same date and period validation used, why would it need to be explained differently? 

i.e. 
4. The processor will evaluate whether each sub-rule's hold period is still active (if the current time is within `hold period` from the `starting timestamp`). If it is, the processor will then evaluate if the final balance of the account would be less than the `hold amount` in the case of the transaction succeeding. If yes, then the transaction will revert.

oscarsernarosero: It is not the same because it is not the same rule. min-max-bal-by-date rule doesn't have periods (plural). It only has one period.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
TJ-Everett: I have this section broken out now in the most recent push - if we like this, I will update all the currently merged rules 
(balance-by-risk  should be merged by that time also)

anne-thrackle: Perhaps "Evaluation Exceptions" should live beneath the "Rule Evaluation" section now.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time`, `amount` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule is active base on `starting timestamp`.
```
ShaneDuncan602: This should be based instead of base

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time`, `amount` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule is active base on `starting timestamp`. 
+5. The processor will evaluate whether the current time is within the `period`.
```
oscarsernarosero: I would say "The processor will evaluate whether the current time is within a new `period` (first transaction in current period)"

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+    - **If it is a new period**, the processor will return the `transfer volume` value from the current transaction for the handler to store.
```
oscarsernarosero: if it is in a new period, it won't evaluate as if it is not?

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+    - **If it is not a new period**, the processor will then evaluate the `transfer volume` (tokens transferred) against the rule's `max volume` (percent in basis units of total supply).
```
oscarsernarosero: Maybe what throws me off is that you mention here the evaluation of the rule, but I think this happens after the volume has been defined which is what this fifth step is about (from what I am understanding).

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time`, `amount` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler. 
+4. The processor will evaluate whether the rule is active base on `starting timestamp`. 
+5. The processor will evaluate whether the current time is within the `period`.
+    - **If it is a new period**, the processor will return the `transfer volume` value from the current transaction for the handler to store.
+    - **If it is not a new period**, the processor will then evaluate the `transfer volume` (tokens transferred) against the rule's `max volume` (percent in basis units of total supply).  
+6. The processor will then calculate the final volume percentage, in basis units, of the total supply by adding the `transfer volume` (tokens transferred) for the period and the `amount` to be transferred in the transaction then dividing against the total supply set in step 3.
```
oscarsernarosero: Like you can say here "[...] by adding the `transfer volume` (tokens transferred) for the period defined in the step 5 [...]"

oscarsernarosero (changes requested):
Much better. I still think it needs some final tweaks.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
ShaneDuncan602: > Shane: Point 1 - is that an ask for a change in the phrasing or an agreement in current phrasing? Point 2 - is step 3 enough clarification or would you prefer it be more clear? I don't want to miss any needed clarity in these PRs as I have multiple at once

1. I think the phrasing needs to be changed a little because the period is on the rule and not the subrule. We might also need to clarify exactly what the term, transfer volume, means. It might help to put that in the purpose section as well. 
2. I think step 3 should mention that, when the token's total supply is used, the value it uses is the one from the beginning of the period.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the token-transfer-volume rule set in the token handler. 
+2. The processor will receive the current `transfer volume`, `last transfer time`, `amount` and token's total supply from the handler.
+3. The processor will evaluate whether the rule has a set total supply or use the token's total supply provided by the handler set at the beginning of every new `period`. 
+4. The processor will evaluate whether the rule is active based on the `starting timestamp`.
```
oscarsernarosero: I would just add here at the end: "If it is not active, the rule evaluation will skip the next steps, and will simply return the `transfer volume` value.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
```diff
@@ -0,0 +1,223 @@
...
+4. The processor will evaluate whether the rule is active based on the `starting timestamp`. 
+5. The processor will evaluate whether the current time is within a new `period`.
+    - **If it is a new period**, the processor will set the `amount` value from the current transaction as the `_volume` value for the volume percent of total supply calculation.
+    - **If it is not a new period**, the processor will then accumulate the `transfer volume` (tokens transferred) and the `amount` of tokens to be transferred as the `_volume` value for the volume percent of total supply calculation. 
+6. The processor will then calculate the final volume percentage, in basis units, from `_volume` and the total supply set in step 3. 
+7. The processor will then evaluate if the final volume percentage of total supply would be greater than the `max volume` in the case of the transaction succeeding. 
+    - If yes, then the transaction will revert. 
+    - If no, the processor will return the `_volume` value for the current `period` to the handler.
```
oscarsernarosero, line 68: This looks amazing now. Great job.

oscarsernarosero (approved):
I love it Tayler. I like how you reorganized the Role Applicability section and reworded the rule evaluation section. I only have one small suggestion, but Im approving rn.

docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md
TJ-Everett: Good call! Added that in there.

ShaneDuncan602 (approved):
Looks great

PR #177, opened 2023-11-01
added MAX-BALANCE-BY-ACCESS-LEVEL doc. Also fixed a couple other files.

oscarsernarosero:


docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
+# Maximum Account Balance By Access Level Rule
+
+## Purpose
+
+The purpose of this rule is to provide control over the balance limits for accounts at the application level. The limits will depend on the *access level* of the accounts. This control is mostly thought as a way of compliancy with possible government regulatory laws around crypto and KYC. For example, KYC could be used as the access level indicator where KYC 0 means no KYC at all, which means the account is only allowed to manage -for instance- 10 dollars worth of application assets, or even no assets at all, whereas KYC 4 means a very thorough KYC process carried out and therefore a trusted party which is allowed to manage millions of dollars within the application economy assets.
```
jjbailey-io: We have to be careful about using verbiage that says "we're making a compliance solution for the government to satisfy regulatory requirements." KYC shouldn't be used at all, that's why we renamed it to "Access level". I recommend using the something more along the lines of:

> The purpose of this rule is to provide balance limits for accounts at the application level based on an application defined segment of users. The segments are defined as the access levels of the accounts. This rule may be used to provided gated accumulation of assets to ensure accounts cannot accumulate more assets without performing other actions defined by the application. For example, the application may decide users may not accumulate any assets without performing specific onboarding activities. The application developer may set a maximum balance of $0 for the default access level and $1000 for the next access level. As accounts are introduced to the ecosystem, they may not receive any assets until the application changes their access level to a higher value.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+    - parameters: 
+        - ruleType: "BALANCE_BY_ACCESSLEVEL".
+        - ruleId: the ruleId set for this rule in the handler.
+
+
+## Dependencies
+
+This rule depends on:
+
+- **Pricing contracts**: pricing contracts for ERC20s and ERC721s need to be setup in the token handlers in order for this rule to work.
```
jjbailey-io, line 181: It would be nice to link the documentation/information on how to set pricing here.

jjbailey-io (changes requested):

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+The purpose of this rule is to provide control over the balance limits for accounts at the application level. The limits will depend on the *access level* of the accounts. This control is mostly thought as a way of compliancy with possible government regulatory laws around crypto and KYC. For example, KYC could be used as the access level indicator where KYC 0 means no KYC at all, which means the account is only allowed to manage -for instance- 10 dollars worth of application assets, or even no assets at all, whereas KYC 4 means a very thorough KYC process carried out and therefore a trusted party which is allowed to manage millions of dollars within the application economy assets. 
+
+## Tokens Supported
+
+- ERC20
+- ERC721
+
+## Scope 
+
+This rule works at the application level which means that all tokens in the app will have no choise but to comply with this rule when active.
```
TJ-Everett: choice*

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+    - This rule can only be activated/deactivated in the application handler by a **rule administrator**.
+    - This rule can only be updated in the application handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The application manager will send to the protocol's rule processor the sum of the balances of every application asset of the account, the access level of the account, the ruleId, and the dollar amount to be transferred in the transaction.
+2. Then, the rule processor will retrieve the maximum balance allowed for the rule with the ruleId passed, and for the access level of the account. If the balance will exceed the maximum allowed by the rule in the case of a successful  transactions, then the transaction reverts.
```
TJ-Everett: there are two spaces after successful

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+        returns (uint32);
+```
+###### *see [AppRuleDataFacet](../../../src/economic/ruleStorage/AppRuleDataFacet.sol)*
+
+### Parameters:
+
+- **_appManagerAddr** (address): the address of the application manager to verify that the caller has rule administrator privileges.
+- **_balanceAmounts** (uint48[]): array of balance limits for each 5 levels (levels 0 to 4) in whole USD amounts (1 -> 1 USD; 1000 -> 1000 USD). Note that the position within the array matters. Posotion 0 represents access level 0, and position 4 represents level 4.
+
+The create function will return the protocol ID of the rule.
```
TJ-Everett: This line should be above ###Parameters:

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
TJ-Everett: I believe (and by that I mean I looked through the docs and couldn't find) we only have the deployment directions for the pricer - we do not currently have documentation outlining the pricer contracts. 
I imagine this will happen a good bit as we build out the docs, when we need to point to something that doesn't exist yet. 

Where is the best place to consolidate these kinds of issues? Should we create a shared doc somewhere and try to collect these as they pop up so when we are done with documentation we can review and make sure these little issues are corrected? @jjbailey-io

TJ-Everett (commented):
Looks good overall Oscar, just a few changes and typos for me. I agree with JB's purpose description.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
jjbailey-io: Good point. @TJ-Everett  Can you generate a ticket for it? We can ignore this for now then.

TJ-Everett: SE 401 created 
Created a collection doc for doc issues [here](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/213778530/Tron+Documentation+TODO+List)

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
+# Maximum Account Balance By Access Level Rule
+
+## Purpose
+
+The purpose of this rule is to provide control over the balance limits for accounts at the application level. The limits will depend on the *access level* of the accounts. This control is mostly thought as a way of compliancy with possible government regulatory laws around crypto and KYC. For example, KYC could be used as the access level indicator where KYC 0 means no KYC at all, which means the account is only allowed to manage -for instance- 10 dollars worth of application assets, or even no assets at all, whereas KYC 4 means a very thorough KYC process carried out and therefore a trusted party which is allowed to manage millions of dollars within the application economy assets. 
+
+## Tokens Supported
+
+- ERC20
+- ERC721
```
ShaneDuncan602, line 10: Since this is application level, should we even have this section? It might be better to remove it for application level rules or to have a blanket statement about applying to all protocol supported assets within the application.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Role Applicability
+
+- **Evaluation Exceptions**: 
+    - This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+    - In the case of the ERC20s, this rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
+
```
ShaneDuncan602: We need to include that it does not apply to burns. That is specifically code into the ProtocolApplicationHandler

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,181 @@
...
+        function checkAccBalanceByAccessLevel(
+                    uint32 _ruleId, 
+                    uint8 _accessLevel, 
+                    uint128 _balance, 
+                    uint128 _amountToTransfer
+                ) 
+                external 
+                view;
+        ```
+- In the [App Manager Handler](../../../src/application/ProtocolApplicationHandler.sol):
```
ShaneDuncan602: I don't think "App Manager Handler" is the correct term. I think it is just ApplicationHandler

ShaneDuncan602 (changes requested):
Just a couple of things.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
oscarsernarosero: This sounds great Johnathan. Thank you.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
oscarsernarosero: I think Markdown still displays one, but I'll delete the extra one anyways. Great catch man.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
oscarsernarosero: That's a @jjbailey-io question

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,182 @@
...
+The purpose of this rule is to provide balance limits for accounts at the application level based on an application defined segment of users. The segments are defined as the access levels of the accounts. This rule may be used to provided gated accumulation of assets to ensure accounts cannot accumulate more assets without performing other actions defined by the application. For example, the application may decide users may not accumulate any assets without performing specific onboarding activities. The application developer may set a maximum balance of $0 for the default access level and $1000 for the next access level. As accounts are introduced to the ecosystem, they may not receive any assets until the application changes their access level to a higher value.
+
+## Tokens Supported
+
+- ERC20
+- ERC721
+
+## Scope 
+
+This rule works at the application level which means that all tokens in the app will have no choice but to comply with this rule when active.
```
anne-thrackle, line 14:
```suggestion
This rule works at the application level which means that all tokens in the app will comply with this rule when the rule is active.
```

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Data Structure
+
+A max-balance-by-access-level rule is composed of a single variable:
+
+- **maxBalance** (mapping(uint8 =>uint48)): the maximum USD worth of application assets that accounts can have per acces level.
```
anne-thrackle:
```suggestion
- **maxBalance** (mapping(uint8 =>uint48)): the maximum USD worth of application assets that accounts can have per access level.
```

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,182 @@
...
+    - This rule can only be set in the application handler by a **rule administrator**.
+    - This rule can only be activated/deactivated in the application handler by a **rule administrator**.
+    - This rule can only be updated in the application handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The application manager will send to the protocol's rule processor the sum of the balances of every application asset of the account, the access level of the account, the ruleId, and the dollar amount to be transferred in the transaction.
```
anne-thrackle:
```suggestion
1. The application manager will send to the protocol's rule processor the dollar value sum of all application assets the account holds, the access level of the account, the ruleId, and the dollar amount to be transferred in the transaction.
```

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
```diff
@@ -0,0 +1,182 @@
...
+2. Then, the rule processor will retrieve the maximum balance allowed for the rule with the ruleId passed, and for the access level of the account. If the balance will exceed the maximum allowed by the rule in the case of a successful transactions, then the transaction reverts.
```
anne-thrackle, line 62: What if I acquired a bunch of in-game gold when it was worth $0.10. Now it's worth $10 and I've gone above my limit based on my access level (through no action of my own, just the value going up independently of me). 
Now I'm trying to sell some gold. But the result after selling it will still be above my limit. Will the transaction succeed?

anne-thrackle (commented):
Looks good. Just a couple minor comments.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
TJ-Everett: This is what I have for the risk rule I am working but applies here if yall like it
"This rule works at the application level. It must be activated and configured in the application handler. Each token in the application ecosystem will be valued and contribute to the rule balance calculation. This rule, when active, will apply to each token within the application."

I don't think we should get rid of the section for standardization reasons.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
TJ-Everett: The transaction would revert, Anne. This rule captures the current price (or at least the last price provided to the pricing contract, so we would assume the contract knows the price has increased for this example). 
You would need to either increase your access level or wait for the price to go back down to transact. @anne-thrackle

TJ-Everett (approved):
Approved for me Oscar, this looks great. 

I know there is an outstanding question for the scope section, but I left my vote in the in line comments.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
oscarsernarosero: That's correct.

docs/userGuides/rules/MAX-BALANCE-BY-ACCESS-LEVEL.md
ShaneDuncan602: Please capture this decision in the ticket @oscarsernarosero

ShaneDuncan602 (approved):
Looks good.

PR #178, opened 2023-11-02
Se 363 create a markdown file for the implementation of the account balance controller by risk score rule

TJ-Everett:


docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,202 @@
+# Account Balance By Risk Rule
+
+## Purpose
+
+The account-balance-by-risk rule enforces accumulated balance limits in $USD for user accounts based on a protocol assigned risk score to that account via the application manager. Risk scores are ranged between 0-99. Balance limits are set by range based on the risk scores given at rule creation. For example, if risk scores given in the array are: 25,50,75 and balance limits are: 500,250,100. The balance limit ranges are as follows: 0-24 = NO LIMIT, 25-49 = 500, 50-74 = 250, 75-99 = 100.
```
ShaneDuncan602: I recommend changing $USD to U.S. Dollars just to cut down on confusion. USD, USDC, etc. It can get confusing.

Also, "The balance limit ranges are as follows" should be, "The balance limit ranges would be as follows"

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,202 @@
...
+ ```c
+    /// ******** Account Balance Rules ********
+    struct AccountBalanceToRiskRuleS {
+        mapping(uint32 => IApplicationRules.AccountBalanceToRiskRule) balanceToRiskRule;
+        uint32 balanceToRiskRuleIndex;
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Role Applicability
```
ShaneDuncan602, line 51: Shouldn't this be Rule instead of Role?

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,202 @@
...
+
+- **Evaluation Exceptions**:
```
ShaneDuncan602, line 53: Also does not apply to burns.

ShaneDuncan602 (changes requested):
Couple of minor things.

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
TJ-Everett: This section calls out the role based exceptions and roles that can enable/disable the rule, so I don't think so.

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
TJ-Everett: While this is technically true since you cannot add a risk score to the address(0) and that all zero scored addresses would be allowed a no limit - that isn't an exemption in the code like it is for the other app level rules. 
I am happy to add that in just clarifying that we don't check for to being address(0) for this one.  

Should this be addressed in a ticket to add this check and confirm?

ShaneDuncan602: The exemption is in ApplicationRiskProcessorFacet. Line 50.

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,203 @@
...
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Role Applicability
+
+- **Evaluation Exceptions**: 
+    - This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+    - In the case of ERC20s, this rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
+    - When the transaction is **burning** of tokens.
```
oscarsernarosero: "When the transaction is **burning** tokens." -of

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,203 @@
...
+    - This rule can only be updated in the application handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the account-balance-by-risk rule set in the application handler. 
+2. The processor will receive the risk score of the user set in the app manager.
+3. The processor will receive the U.S. Dollars value of all protocol supported tokens owned by the to address and the U.S. Dollars value of the transaction.
```
oscarsernarosero: "U.S. dollar value". `dollar` doesn't have to be capitalized and it should be singular

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor will receive the ID of the account-balance-by-risk rule set in the application handler. 
+2. The processor will receive the risk score of the user set in the app manager.
+3. The processor will receive the U.S. Dollars value of all protocol supported tokens owned by the to address and the U.S. Dollars value of the transaction. 
+4. The processor will loop through the risk scores within the rule ID provided to find the range that the user is within. The processor loops until a risk score within the rule is greater than the risk score of the to address and uses the previous `max balance` value.
```
oscarsernarosero: This might be too much technical detail. You can simply state that the protocol will find the balance limit for the risk score of the user.

ShaneDuncan602 (approved):
Looks great

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -0,0 +1,203 @@
...
+function addAccountBalanceByRiskScore(
+    address _appManagerAddr, 
+    uint8[] calldata _riskScores, 
+    uint48[] calldata _balanceLimits
+) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+###### *see [AppRuleDataFacet](../../../src/economic/ruleStorage/AppRuleDataFacet.sol)*
+
+
+The create function in the protocol needs to receive the appManager address of the application in order to verify that the caller has Rule administrator privileges.
```
oscarsernarosero: I stopped putting this line here because we state the same in the Role Applicability section and also in the parameter section, but not sure about it. What do you think?

oscarsernarosero (approved):
Looks great Tayler. I just requested some minor changes, but Im approving now.

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
TJ-Everett: I am fine with removing this - we just have to remove from all the currently merged docs - I will do so in this ticket.

PR #179, opened 2023-11-03
did Admin withdrawal rule and fixed a couple small things in other docs

oscarsernarosero:
I think I accidentally changed the `scope` for all the rules to application level rules in a past ticket, so I fixed it in this one. Also fixed a couple other minor typos.

### DIFF
--- a/docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
+++ b/docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
@@ -47,7 +47,7 @@ These rules are stored in a mapping indexed by ruleId(uint32) in order of creati
 - This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
 - This rule can only be updated in the asset handler by a **rule administrator**.
 
-In order to prevent the malfunction of this rule, there are some special restrictions:
+Since this rule is intended to apply specifically to app administrators, there are some special restrictions:
 
 - This rule can only be deactivated if current rule is outside its active period (post `releaseDate`).
 - This rule prevents app administrators from renouncing their roles when the rule is in its active period (pre `releaseDate`).
--- a/docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
+++ b/docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
@@ -1,0 +1,186 @@
+# Admin Withdrawal Rule
+
+## Purpose
+
+The purpose of the admin-withdrawal rule is to allow developers to show proof to their community that they will hold a certain amount of tokens for certain period of time. This is usually done to prove that a sudden increase of supply flooding the market won't happen as a malicious attempt of the developers to rug pull their community.
+
+## Tokens Supported
+
+- ERC20
+- ERC721
+
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+An admin-withdrawal rule is composed of 2 variables:
+
+- **Amount** (uint256): The minimum amount of tokens to be held by the admin until the *releaseDate* (value in weis).
+- **Release Date** (uint256): The Unix timestamp of the date after which the administrator is free to transfer the tokens.
+
+```c
+/// ******** Admin Withdrawal Rules ********
+    struct AdminWithdrawalRule {
+        uint256 amount;
+        uint256 releaseDate; /// timestamp
+    }
+```
+###### *see [RuleDataInterfaces](../../../src/economic/ruleStorage/RuleDataInterfaces.sol)*
+
+These rules are stored in a mapping indexed by ruleId(uint32) in order of creation:
+
+ ```c
+    /// ******** Admin Withdrawal Rules ********
+    struct AdminWithdrawalRuleS {
+        mapping(uint32 => ITaggedRules.AdminWithdrawalRule) adminWithdrawalRulesPerToken;
+        uint32 adminWithdrawalRulesIndex;
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Configuration and Enabling/Disabling
+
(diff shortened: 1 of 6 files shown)


docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
```diff
@@ -0,0 +1,186 @@
+# Admin Withdrawal Rule
+
+## Purpose
+
+The purpose of the admin-withdrawal rule is to allow developers to show proof to their community that they will hold a certain amount of tokens for certain period of time. This is usually done to prove that a sudden increase of supply flooding the market won't happen as a malicious attempt of the developers to rug pull their community.
```
anne-thrackle:
```suggestion
The purpose of the admin-withdrawal rule is to allow developers to prove to their community that they will hold a certain amount of tokens for a certain period of time. Adding this rule prevents developers from flooding the market with their supply and effectively "rug pulling" their community. 
```

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
```diff
@@ -0,0 +1,186 @@
...
+    }
+```
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Configuration and Enabling/Disabling
+
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
```
anne-thrackle, line 48: if the rule can be deactivated or updated by a rule admin (presumably one of the devs), then it's not much a guarantee, is it? Maybe the idea is that the rule admin would be governance-based, or NOT one of the devs somehow?

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The asset handler will check if the transfer of tokens will be from an app administrator account. If it is not, the rule evaluation will be skipped.
+2. Then, the handler will send the amount of tokens being transferred along with the current balance of the app administrator account and the ruleId to the protocol's rule processor.
+3. The rule processor will then compare the final balance of the administrator account if the transaction succeeds. If the final balance will be less than the minimum balance specified in the rule, the transaction will revert.
```
anne-thrackle: I suggested this in a previous ticket, but I'll suggest it again... I think using present tense would make these sections easier to parse. 

```suggestion
1. The asset handler checks if the transfer of tokens is from an app administrator account. If it is not, the rule evaluation is skipped.
2. The handler sends the amount of tokens being transferred, the current balance of the app administrator account, and the ruleId to the protocol's rule processor.
3. The rule processor calculates what the final balance of the administrator account would be if the transaction succeeds. If the final balance calculated is less than the minimum balance specified in the rule, the transaction reverts.
```

anne-thrackle (commented):
looks good, this one was quite clear to me. Just a couple phrasing suggestions.

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Scope 
+
+This rule works at a token level. It must be activated and configured for each desired token in the corresponding token handler.
+
+## Data Structure
+
+An admin-withdrawal rule is composed of 2 variables:
+
+- **Amount** (uint256): The minimum amount of tokens to be held by the admin until the *releaseDate* (value in weis).
```
TJ-Everett: Just adding a comment here to un-pluralize wei here, as spoken about in the slack thread.

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
TJ-Everett: Anne, the rule has logic that will prevent an active rule from being disabled until the rule period is over. So, while this isn't a perfect guarantee, once the rule is started all admin role accounts would fall under the minimum balance until the rule ends. 

Oscar, we should call this out somewhere that the rule cannot be deactivated while it is within the rule period.

TJ-Everett (commented):
Just the Weis=>wei on line 20 and the call out for deactivation before releaseDate. Otherweis (haha) this looks great!

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
ShaneDuncan602: I agree. This is a very important feature of this rule. The admins can't deactivate until the period has passed.

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
ShaneDuncan602: I agree

ShaneDuncan602 (approved):
My only concerns were the same as already commented by the other reviewers. Once those are addressed, I approve.

docs/userGuides/rules/ADMIN-WITHDRAWAL-RULE.md
```diff
@@ -47,15 +47,20 @@ These rules are stored in a mapping indexed by ruleId(uint32) in order of creati
 - This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
 - This rule can only be updated in the asset handler by a **rule administrator**.
 
+In order to prevent the malfunction of this rule, there are some special restrictions:
```
anne-thrackle:
```suggestion
Since this rule is intended to apply specifically to project administrators, there are some special restrictions:
```

oscarsernarosero: I like it. Changed.

TJ-Everett (approved):
Looks great Oscar

PR #180, opened 2023-11-03
Se 362 create a markdown file for the implementation of the token percentage sell rule

TJ-Everett:


docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
+# Token Transfer Volume Rule
+
+## Purpose
+
+The token-percentage-sell rule enforces a limit of the sale of tokens during a period. This rule sets a percentage of the token's total supply that is able to be sold per period. For this rule, a sell is considered a swap of an application token for an non application or chain native token. This can also be interpreted as an account leaving the application's ecosystem via AMM an swap.
```
oscarsernarosero: "via AMM an swap"?

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.  
+4. The processor evaluates whether the current time is within a new `period`.
+    - **If it is a new period**, the processor will use the `_token_amount_0` value from the current transaction as the `percentOfTotalSupply` value for the sell percent of total supply calculation.
+    - **If it is not a new period**, the processor will then accumulate the `totalSoldWithinPeriod` (tokens transferred) and the `_token_amount_0` of tokens to be transferred as the `percentOfTotalSupply` value for the sell percent of total supply calculation. 
+5. The processor calculates the final sell percentage, in basis units, from `percentOfTotalSupply` and the total supply set in step 3. 
+6. The processor evaluates if the final sell percentage of total supply would be greater than the `tokenPercentage` in the case of the transaction succeeding. 
+    - If yes, then the transaction will revert. 
+    - If no, the processor will return the `tokenPercentage` value for the current `period` to the handler.
```
oscarsernarosero: Great job here

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+) external ruleAdministratorOnly(_appManagerAddr) returns (uint32);
+```
+###### *see [RuleDataFacet](../../../src/economic/ruleStorage/RuleDataFacet.sol)* 
+
+The create function will return the protocol ID of the rule.
+
+### Parameters:
+
+- **_appManagerAddr** (address): The address of the application manager to verify that the caller has Rule administrator privileges.
+- **_tokenPercentage** (uint16): maximum allowable basis unit percentage of trading volume per period.
```
oscarsernarosero: "maximum allowable percentage (in basis unit) of trading volume per period."

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- **_sellPeriod** (uint16): the amount of hours per period.
+- **_startTimestamp** (uint64): starting timestamp of the rule. This timestamp will determine the time that a day starts and ends for the rule processing. For example, *the amount of trades will reset to 0 every day at 2:30 pm.*
```
oscarsernarosero: Unix timestamp

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- **_totalSupply** (uint256): (optional) if not 0, then this is the value used for totalSupply instead of the live token's totalSupply value at rule processing time.
+
+
+### Parameter Optionality:
+
+The parameters where developers have the options are:
+- **_totalSupply**: For volatility calculations, when this is zero, the token's totalSupply is used. Otherwise, this value is used as the total supply.
```
oscarsernarosero: "For sell volume over supply calculations, when [...]"

oscarsernarosero (approved):
Very minor typos. Great job in general tho.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
+# Token Transfer Volume Rule
+
+## Purpose
+
+The token-percentage-sell rule enforces a limit of the sale of tokens during a period. This rule sets a percentage of the token's total supply that is able to be sold per period. For this rule, a sell is considered a swap of an application token for an non application or chain native token. This can also be interpreted as an account leaving the application's ecosystem via an AMM swap.
```
anne-thrackle:
```suggestion
The token-percentage-sell rule enforces a limit on the sale of tokens during a certain time period. This rule sets the limit as a percentage of the token's total supply. For this rule, a "sell" is a swap of an application token for a token that is not part of the application (e.g. the chain native token, or a token from another application). 
```
Not sure that "non application token" was clear so I suggested a rephrase, but maybe it's ok. 

I removed the last sentence, I think it caused a bit of confusion for me (maybe the "leaving the ecosystem" which I don't think is related to this rule) - but I'm not against adding something like "for example, an account using an AMM to swap their application token for chain native token".

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Scope 
+
+This rule works at an asset level. It must be activated and configured for each desired AMM in the corresponding asset handler.
+
+## Data Structure
+
+A token-percentage-sell rule is composed of 4 components:
+
+- **Token Percentage** (uint16): The maximum percent in basis units of total supply to be sold during the *period*.
```
anne-thrackle:
```suggestion
- **Token Percentage** (uint16): The maximum percent in basis units of total supply able to be sold during the *period*.
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- **Sell  Period** (uint16): The amount of hours that defines a period.
```
anne-thrackle:
```suggestion
- **Sell  Period** (uint16): The length of time for which the rule will apply, in hours.
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+###### *see [IRuleStorage](../../../src/economic/ruleStorage/IRuleStorage.sol)*
+
+## Configuration and Enabling/Disabling
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
```
anne-thrackle, line 53: I find this section confusing again. Trying to figure out why, I think it's: 
- variables being introduced without explaining what they are, and it not being obvious from the name. Especially `_token_amount_0` is a very nondescript name. 
- some variables being camel case, some snake case, some just words with spaces. Makes it harder to parse and digest. 
- statements like "the processor will use the `_token_amount_0` value from the current transaction as the `percentOfTotalSupply` value for the sell percent of total supply calculation." What does it mean to use one value as another, for two variables that I don't know what they are? 

I wonder if these evaluation sections would be better served by actual pseudocode. I think at least more accurate and descriptive/more consistent variable names would help.

anne-thrackle (changes requested):
Good overall, I think the Evaluation section needs a bit of work.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
TJ-Everett: - Added clarity for new variable intro'ed via the handler params 
- changed all variables to be plain english words with spaces 
- The rule will either take the combined total of the token amount 0 and the percent of total supply from the handler if within an existing period, or take just the token amount 0 for the current transaction if in a new period to find the transaction percentage of total supply value. 

I do believe that some of this will become much more clear as we expand the docs to the handlers and assets and not just the rules. Creating a clearer picture, however yes we need a better handle on standardized variable names and to do a once through the protocol do catch instances of naming conventions not making sense.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
+# Token Transfer Volume Rule
+
+## Purpose
+
+The token-percentage-sell rule enforces a limit on the sale of tokens during a certain time period. This rule sets the limit as a percentage of the token's total supply. For this rule, a "sell" is a swap of an application token for a token that is not part of the application (e.g. the chain native token, or a token from another application).  
+
```
ShaneDuncan602: I know this is the intention of the rule, but they could also create an AMM for two application tokens and still use it.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+## Tokens Supported
+
+- ERC20
```
ShaneDuncan602, line 9: Shouldn't this be AMM? I know it's not a token but this rule is only supported within the AMM, right?

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Scope 
+
+This rule works at an asset level. It must be activated and configured for each desired AMM in the corresponding asset handler.
+
+## Data Structure
+
+A token-percentage-sell rule is composed of 4 components:
+
+- **Token Percentage** (uint16): The maximum percent in basis units of total supply able to be sold during the *period*.
```
ShaneDuncan602: it might be helpful to give an example of the numbers since it is shifted to account for lack of decimals.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- **Sell  Period** (uint16): The length of time for which the rule will apply, in hours.
+- **Starting Timestamp** (uint64): The timestamp of the date when the *period* starts counting.
+- **Total Supply** (uint256): if not zero, this value will always be used as the token's total supply for rule evaluation. This can be used when the amount of circulating supply is much smaller than the amount of the token totalSupply due to some tokens being locked in a dev account or a vesting contract, etc. Only use this value in such cases.
```
ShaneDuncan602, line 22: Shouldn't we also include the validations for these? For instance, that period cannot be zero

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
ShaneDuncan602: I agree. I think internally labeling the tokens and token0 and token1 is fine in the code, but it's not how most people view things. It's our nerdy engineering jargon. For the sake of documentation, token_A and token_B or some other more natural label might be better.

ShaneDuncan602 (changes requested):
Just minor things. I'm not sure where it would fit best, but I think it needs to be clear that this rule is only applicable to AMM's.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor receives the ID of the token-percentage-sell rule set in the asset handler. 
+2. The processor receives the current `total sold within period`, `token amount 0` (total amount of token 0 being transferred), `previous sell time`, and token's total supply from the handler.
```
anne-thrackle:
```suggestion
2. The processor receives the current `total sold within period`, `token amount 0` (total amount of token 0 being transferred in the current transaction), `previous sell time`, and token's total supply from the handler.
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.  
+4. The processor evaluates whether the current time is within a new `period`.
+    - **If it is a new period**, the processor will use the `token amount 0` value from the current transaction as the `percent of total supply` value for the sell percent of total supply calculation.
```
anne-thrackle:
```suggestion
    - **If it is a new period**, the processor sets the `percent of total supply` to the value of `token amount 0`. 
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+    - **If it is not a new period**, the processor will then accumulate the `total sold within period` (tokens transferred) and the `token amount 0` of tokens to be transferred as the `percent of total supply` value for the sell percent of total supply calculation.
```
anne-thrackle:
```suggestion
    - **If it is not a new period**, the processor sets `percent of total supply` to the sum of `total sold within period` and `token amount 0`.   
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+
+1. The processor receives the ID of the token-percentage-sell rule set in the asset handler. 
+2. The processor receives the current `total sold within period`, `token amount 0` (total amount of token 0 being transferred), `previous sell time`, and token's total supply from the handler.
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.  
+4. The processor evaluates whether the current time is within a new `period`.
+    - **If it is a new period**, the processor will use the `token amount 0` value from the current transaction as the `percent of total supply` value for the sell percent of total supply calculation.
+    - **If it is not a new period**, the processor will then accumulate the `total sold within period` (tokens transferred) and the `token amount 0` of tokens to be transferred as the `percent of total supply` value for the sell percent of total supply calculation. 
+5. The processor calculates the final sell percentage, in basis units, from `percent of total supply` and the total supply set in step 3.
```
anne-thrackle:
```suggestion
5. The processor calculates the final sell percentage, in basis units, using the `percent of total supply` calculated in step 4 and the total supply set in step 3. 
```

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.  
+4. The processor evaluates whether the current time is within a new `period`.
+    - **If it is a new period**, the processor will use the `token amount 0` value from the current transaction as the `percent of total supply` value for the sell percent of total supply calculation.
+    - **If it is not a new period**, the processor will then accumulate the `total sold within period` (tokens transferred) and the `token amount 0` of tokens to be transferred as the `percent of total supply` value for the sell percent of total supply calculation. 
+5. The processor calculates the final sell percentage, in basis units, from `percent of total supply` and the total supply set in step 3. 
+6. The processor evaluates if the final sell percentage of total supply would be greater than the `token percentage` in the case of the transaction succeeding. 
+    - If yes, then the transaction will revert. 
+    - If no, the processor will return the `tokenPercentage` value for the current `period` to the handler.
```
anne-thrackle:
```suggestion
    - If no, the processor returns the `tokenPercentage` value for the current `period` to the handler.
```

anne-thrackle (approved):
discussed this last set of suggestions with Tyler live. We may also want to have a broader conversation with the team about the approach here, and variable names.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
TJ-Everett: I will expand the scope section a bit to be more explicitly clear that this is an AMM applied rule. 
I think the token section should remain as is and then add in 721 and 1155 as we build those AMM's

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
TJ-Everett: We have the parameter validation steps on line 117

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
oscarsernarosero: This is a great point. Maybe renaming this to "Protocol-Element Supported" or something of the kind

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+
+The token-percentage-sell rule enforces a limit on the sale of tokens during a certain time period. This rule sets the limit as a percentage of the token's total supply. For this rule, a "sell" is a swap of an application token for a token that is not part of the application (e.g. the chain native token, or a token from another application). This rule can also be applied to AMMs created for two application tokens. In these cases the tokens are defined as token 0 and token 1 within the AMM. This rule will apply to token 0 in the AMM swaps. 
+
+## Tokens Supported
+
+- ERC20
+
+## Scope 
+
+This rule works at the AMM level. It must be activated and configured for each desired AMM in the corresponding asset handler. This rule will not be applied at the token level and will only be checked through the AMM swap function.
```
oscarsernarosero: "AMM handler"? I don't think an AMM can be considered an asset.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The processor receives the ID of the token-percentage-sell rule set in the asset handler. 
+2. The processor receives the current `total sold within period`, `token amount 0` (total amount of token 0 being transferred in the current transaction), `previous sell time`, and token's total supply from the handler.
```
oscarsernarosero: This looks good, but looking into Shane's and Anne's comments, I would just avoid putting variable descriptions in "code blocks". So for instance just " total sold within period " instead of " `total sold within period` ". I think that would make much clearer the statements. Also, echoing Shane's point about us being nerds (lol), I would rename token 0 to token A, and instead of "token amount A" (assuming you are accepting my naming suggestion) maybe say "token A amount" since the former could be interpreted as "amount A of tokens" (where A is a number) instead of the "the amount of tokens for token A"

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.
```
oscarsernarosero, line 59: for instance `period` here is well used because it has been already defined in the data structure section. So I think this is good.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+
+1. The processor receives the ID of the token-percentage-sell rule set in the asset handler. 
+2. The processor receives the current `total sold within period`, `token amount 0` (total amount of token 0 being transferred in the current transaction), `previous sell time`, and token's total supply from the handler.
+3. The processor evaluates whether the rule has a set total supply or uses the token's total supply provided by the handler set at the beginning of every new `period`.  
+4. The processor evaluates whether the current time is within a new `period`.
+    - **If it is a new period**, the processor sets the `percent of total supply` to the value of `token amount 0`.
+    - **If it is not a new period**, the processor sets `percent of total supply` to the sum of `total sold within period` and `token amount 0`. 
+5. The processor calculates the final sell percentage, in basis units, using the `percent of total supply` calculated in step 4 and the total supply set in step 3.  
+6. The processor evaluates if the final sell percentage of total supply would be greater than the `token percentage` in the case of the transaction succeeding.
```
oscarsernarosero: same suggestions as step 2

oscarsernarosero: Also, in line 61, you could use a more intuitive and less technical name for the token like "amount of tokens being sold in this transaction", or something like that.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+5. The processor calculates the final sell percentage, in basis units, using the `percent of total supply` calculated in step 4 and the total supply set in step 3.  
+6. The processor evaluates if the final sell percentage of total supply would be greater than the `token percentage` in the case of the transaction succeeding. 
+    - If yes, then the transaction will revert. 
+    - If no, the processor returns the `tokenPercentage` value for the current `period` to the handler.
+
+###### *see [ERC20RuleProcessorFacet](../../../src/economic/ruleProcessor/ERC20RuleProcessorFacet.sol) -> checkSellPercentagePasses*
+
+## Evaluation Exceptions 
+- This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+- In the case of ERC20s, this rule doesn't apply when a **registered treasury** address is in the *to* side of the transaction.
```
oscarsernarosero, line 72: This is something that is in the code so we should keep it, but Im wondering now why an AMM would send tokens to a treasury address.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,222 @@
...
+###### *see [RuleDataFacet](../../../src/economic/ruleStorage/RuleDataFacet.sol)* 
+
+The create function will return the protocol ID of the rule.
+
+### Parameters:
+
+- **_appManagerAddr** (address): The address of the application manager to verify that the caller has Rule administrator privileges.
+- **_tokenPercentage** (uint16): maximum allowable percentage (in basis unit) of trading volume per period.
+- **_sellPeriod** (uint16): the amount of hours per period.
+- **_startTimestamp** (uint64): starting unix timestamp of the rule. This timestamp will determine the time that a day starts and ends for the rule processing. For example, *the amount of trades will reset to 0 every day at 2:30 pm.*
```
oscarsernarosero: In this case, you should mention that this example is only valid if `_sellPeriod` is 24 hours (1 day). Otherwise that example doesn't really apply because you start getting an accumulated shifting of hours per period (unless `_sellPeriod` is a factor of 24).

oscarsernarosero (commented):
Looking into Shane's and Anne's comments which I think are valid, I made some more suggestions.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
anne-thrackle: +1 to "token A"

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
TJ-Everett: I believe this is for Fees being deposited to a Treasury account and not wanting additional rules on that transfer.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
TJ-Everett: In previous conversations and docs we have called the handlers asset handlers and that continued to include AMM's but I have clarified here all the same that this is applied in the AMM handler.

TJ-Everett: To the Tokens supported section, I think this is valuable to keep for both consistency in the docs as well as when we do add support for additional tokens.

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
oscarsernarosero: awesome! Thank you

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
ShaneDuncan602: I don't think we want to define the sell as between chain native and application tokens and then tell them they can also apply to any AMM swap. It would probably be clearer to just say that it only applies to AMM swaps

ShaneDuncan602 (changes requested):
A little change to the purpose statement

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -2,7 +2,7 @@
 
 ## Purpose
 
-The token-percentage-sell rule enforces a limit on the sale of tokens during a certain time period. This rule sets the limit as a percentage of the token's total supply. For this rule, a "sell" is a swap of an application token for a token that is not part of the application (e.g. the chain native token, or a token from another application). This rule can also be applied to AMMs created for two application tokens. In this case the tokens are defined as token 0 and token 1 within the AMM. This rule will apply to token 0 in the AMM swaps. 
+The token-percentage-sell rule enforces a limit on the sale of tokens during a certain time period. This rule sets the limit as a percentage of the token's total supply. The rule will apply to the token designated token 0 in the AMM at construction and apply to all AMM swaps of token 0 to token 1 while this rule is active.
```
ShaneDuncan602, line 5: This applies to both sides of the swap. It also checks it on token 1 to token 0 swaps.

TJ-Everett: The swap for 1 to 0 passes in the purchase action type so it would not check the sell only the purchase, no?

TJ-Everett: That is why I was defining "sell" in the purpose before

ShaneDuncan602: Ok, now I remember. Sheesh, the application of AMM rules is very confusing. Only one side of the swap is a sell but we don't enforce anything when the tokens are set in the AMM. If they want this rule on both sides of the swap, they have to create its purchase counterpart and apply it even if the rule intention is the same. We definitely need to revisit this in the future. I'm going to add this comment to Jira so we have it.

ShaneDuncan602 (approved):
Looks good.

PR #181, opened 2023-11-07
Se 389 gas comparisons erc721 c to protocol

oscarsernarosero:
Report available [here](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/216825857/ERC721C+vs+Tron+Implementation+of+it+Gas+Comparison)


TJ-Everett (approved):
This is great work Oscar! Really interesting results here.

src/token/ERC721/ProtocolERC721.sol
```diff
@@ -108,7 +108,7 @@ contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, P
      */
     function _beforeTokenTransfer(address from, address to, uint256 tokenId, uint256 batchSize) internal override(ERC721, ERC721Enumerable) whenNotPaused {
         // Rule Processor Module Check
-        require(handler.checkAllRules(from == address(0) ? 0 : balanceOf(from), to == address(0) ? 0 : balanceOf(to), from, to, batchSize, tokenId, ActionTypes.TRADE));
+        require(handler.checkAllRules(msg.sender, from == address(0) ? 0 : balanceOf(from), to == address(0) ? 0 : balanceOf(to), from, to, batchSize, tokenId, ActionTypes.TRADE));
```
ShaneDuncan602, line 111: Can you explain why we need the msg.sender here?

src/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -24,19 +25,29 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, RuleAdministra
...
+
     address public erc721Address;
     /// RuleIds for implemented tagged rules of the ERC721
     uint32 private minMaxBalanceRuleId;
     uint32 private minBalByDateRuleId;
     uint32 private minAccountRuleId;
     uint32 private oracleRuleId;
+    uint32 private operatorOracleRuleId;
+    uint32 private senderOracleRuleId;
+    uint32 private recipientOracleRuleId;
```
ShaneDuncan602, line 39: Love this.

src/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -221,6 +234,20 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, RuleAdministra
         }
     }
 
+    function _checkRoyaltyEnforcementRules(address _sender, address _from, address _to) internal view {
+        if(senderOracleRuleActive && _sender != _from) ruleProcessor.checkOraclePasses(senderOracleRuleId, _sender);
+        if(recipientOracleRuleActive && _to.isContract()) ruleProcessor.checkOraclePasses(recipientOracleRuleId, _to);
+        if(restrictedToTradesOnlyRuleActive && _sender == _from) revert NoOTCTransfersAllowed();
+    }
```
ShaneDuncan602, line 241: Seems like there is some room for further optimization here. We could have one function that checks both as well. Not recommending we do it, just mentioning it for the future.

ShaneDuncan602 (approved):
I absolutely love this. You crushed a very difficult ticket and in a detailed and easy to follow way. Great job.

src/token/ERC721/ProtocolERC721.sol
oscarsernarosero: This is necessary because msg.sender could be the owner or an approved address. So, this is how you identify operators/marketplaces, and it also lets you identify trades from OTC transfers.

src/token/ERC721/ProtocolERC721Handler.sol
oscarsernarosero: You might be right Shane

PR #183, opened 2023-11-10
removed try catch clauses from processor facets

oscarsernarosero:


TJ-Everett (approved):
Looks great to me Oscar, Approving for internal here. I will be happy to PR for external if that is what the deep state decides in your slack thread.

ShaneDuncan602 (approved):
Good job.

PR #189, opened 2023-11-17
Se 424 restrict super admins from revoking app admins while admin withdrawal rule active

oscarsernarosero:
turned out to be smaller than what I thought.


TJ-Everett (approved):
Nice work! Great catch on finding this one Oscar.

ShaneDuncan602 (approved):
Looks good.

PR #190, opened 2023-11-17
SE-427-Utilize-factories-to-deploy-AMMs-in-Tron

ShaneDuncan602:
Added the AMM Factory.


TJ-Everett (approved):
This looks great Shane. 
I have some small questions: Do we want AMM type specific events? Instead of just a new AMM event with the address it is deployed too? Also if we do that should we include the token addresses that are added? Will Doom need that info at this step?

src/liquidity/ProtocolAMMFactory.sol
```diff
@@ -0,0 +1,86 @@
...
+     * @dev This creates a sample01 AMM and calculation module.
+     * @param _token0 valid ERC20 address
+     * @param _token1 valid ERC20 address
+     * @param _f_tracker f(x) tracker value
+     * @param _g_tracker g(x) tracker value
+     * @param _appManagerAddress address of the application's appManager
+     * @return _calculatorAddress
+     */
+    function createSample01AMM(address _token0, address _token1, int256 _f_tracker, int256 _g_tracker, address _appManagerAddress) external returns (address) {
+        return address(createAMM(_token0, _token1, _appManagerAddress, address(protocolAMMCalculatorFactory.createSample01(_f_tracker, _g_tracker, _appManagerAddress))));
```
oscarsernarosero, line 84: love it

oscarsernarosero (approved):
Really nice Shane.

ShaneDuncan602:
> This looks great Shane. I have some small questions: Do we want AMM type specific events? Instead of just a new AMM event with the address it is deployed too? Also if we do that should we include the token addresses that are added? Will Doom need that info at this step?

That is a really good question. I'm not sure if we will need them or not. Much of this is just a first run at it. I assume that we will need to tweak in the future as we implement into DOOM and think about the permissions of it. Currently, it is set where anyone can create these objects.

PR #191, opened 2023-11-20
Se 450 general documentation updates

anne-thrackle:
Reorganized the landing page README to be more approachable to an external user. Removed duplicate README content. 

Renamed "Developer Guide" in favor of "User Guide" and "Contributor Guide" to remove ambiguity about who the docs are for.


docs/README.md
```diff
@@ -1,27 +1,11 @@
...
-# [Architecture Overview][archOverview-url]
-# [Developer Guide][developer-url]
-# [Deployment Guide][deploymentGuide-url]
-# [Rule Guide][ruleGuide-url]
-# [Glossary][glossary-url]
-
-
-
----
 ## Documentation maintenance
```
jjbailey-io: This should be an H1 (remove a #)

docs/README.md
```diff
@@ -1,27 +1,11 @@
...
 
 This repository has two types of documentation:
 1. Manually generated - this is the documentation that is written by hand
 2. Auto-generated - this is the documentation that is generated by the code
 
-
 ### Manually Generated Documentation
```
jjbailey-io: Likewise, remove another # here to elevate to a level 2 heading.

docs/README.md
```diff
@@ -1,27 +1,11 @@
...
-The manually generated documentation root can be found [here](userGuides/DEVELOPER-GUIDE.md).
+The manually generated documentation root can be found [here](userGuides/USER-GUIDE.md).
 
 ### Auto-generated Documentation
```
jjbailey-io: H2 heading

docs/contributorGuides/CONTRIBUTOR-GUIDE.md
jjbailey-io, line 1: If we renamed this file to README.md, it would auto-display in Github when navigating to that folder.

docs/userGuides/USER-GUIDE.md
jjbailey-io, line 1: Similar to the previous comment. If we rename this to README.md, it'll autodisplay to users navigating to the folder in GitHub.

jjbailey-io (changes requested):
I commented on a few items. The only real change should probably be to make the headers consistent on the modified files. We could get some added benefit by renaming files but I realize that was not really in the scope of original changes.

oscarsernarosero (approved):
Looks good Anne. I have the same suggestions about the renaming for some docs to be README, so GH can automatically display them, but other than that, It looks great. So Im approving.

jjbailey-io (approved):

ShaneDuncan602 (approved):
Looks great.

PR #195, opened 2023-11-24
Se 442 nft amm calculation factory

oscarsernarosero:


script/python/linear_calculator.py
```diff
@@ -0,0 +1,27 @@
...
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    calculate_y(args)
+
+
+if __name__ == "__main__":
+    main()
```
ShaneDuncan602, line 27: This is sweet.

src/liquidity/ProtocolAMMCalculatorFactory.sol
```diff
@@ -34,6 +36,23 @@ contract ProtocolAMMCalculatorFactory is AppAdministratorOnly, IZeroAddressError
...
+     * @dev This creates a linear calculation module.
+     * @notice a LineInput has the shape {uint256 m; uint256 b}
+     *    *m* is the slope of the line expressed with 8 decimals of precision. Input of 100000001 means -> 1.00000001
+     *    *b* is the intersection of the line with the ordinate axis expressed in atto (18 decimals of precision). 1 ^ 10 ** 18 means -> 1
+     * @param buyCurve the definition of the buyCurve
+     * @param sellCurve the definition of the sellCurve
+     * @param _appManagerAddress address of the application's appManager
+     * @return _calculatorAddress
+     */
+    function createDualLinearNFT(LineInput memory buyCurve, LineInput memory sellCurve, address _appManagerAddress) external returns (address) {
```
ShaneDuncan602, line 49: This curve abstraction is great. I considered something similar for the other curves but my concern was for ease of invoking the functions from a command line through cast or something similar. We are getting away from that manual access model but is this easily called from DOOM?

src/liquidity/calculators/ProtocolAMMCalcLinear.sol
```diff
@@ -68,8 +68,7 @@ contract ProtocolAMMCalcLinear is IProtocolAMMFactoryCalculator {
      * @dev calculate the total digits in a number.
      * @param _number number to count digits for
      */
-    function _numDigits(uint256 _number) private pure returns (uint8) {
-        uint8 digits = 0;
+    function _numDigits(uint256 _number) private pure returns (uint8 digits) {
```
ShaneDuncan602, line 71: Whoops! Good catch.

src/liquidity/calculators/ProtocolNFTAMMCalcDualLinear.sol
```diff
@@ -0,0 +1,141 @@
...
+import {Line, LineInput, Curve} from "./libraries/Curve.sol";
+import {CurveErrors} from "../../interfaces/IErrors.sol";
+
+/**
+ * @title Automated Market Maker Swap Linear Calculator for NFT Pools
+ * @dev This is external and used by the ProtocolAMM. The intention is to be able to change the calculations
+ *      as needed. It contains an example linear. It is built through ProtocolAMMCalculationFactory
+ * @author @ShaneDuncan602 @oscarsernarosero @TJ-Everett
+ */
+contract ProtocolNFTAMMCalcDualLinear is IProtocolAMMFactoryCalculator, CurveErrors {
```
ShaneDuncan602, line 14: Since we will likely be adding the curve data structures to all the calculation module, we should probably create an interface(ICurve) or abstract contract for it. Each calculator would need the common dependencies and functions. Things like: _validateSingleCurve, Curve.sol import, etc. would be shared for all. I'm not sure that we would want all possible curves to be in a library. When I think about libraries, I think of them as Singletons for housing common logic that would be reused over and over again by many different external modules. But the Linear module would never use Sigmoid functions in the the library and vice versa. The consolidation of logic seems like it might be a little redundant since the calculators will get generated via the factory. Is the thought here to make the common curve logic available to devs who want to create their own calculation module? I think they would create their own calculator when they did not like our logic.

test/liquidity/ProtocolNFTAMMCalcFuzz.t.sol
```diff
@@ -0,0 +1,214 @@
...
+        if(isAscii){
+            _checkPriceAsAsciiWithinTolerance(price, res);
+        }
+        /// if the response was not an ascii.
+        else{
+            /// we decode the fake decimal and check if the difference is within tolerance
+            _checkPriceAsFakeDecimalWithinTolerance(price, res);
+        }
+        /// if the check above did not revert, then the test passed
+        emit Passed();
```
ShaneDuncan602: This is awesome. We need it in all the AMM tests. But could the _checkPrice functions return a bool that we use in an assert to maintain conformity with our other tests?

test/liquidity/ProtocolNFTAMMCalcFuzz.t.sol
```diff
@@ -0,0 +1,214 @@
...
+    error OutOfTolerance(uint256 _contractPrice, uint256 _pythonPrice); // 0x700e344e
+
+    using Strings for uint256;
+    uint256 constant PRECISION_DECIMALS = 8;
+    uint256 constant ATTO = 10 ** 18;
+    uint256 constant Y_MAX = 1_000_000_000_000_000_000_000_000 * ATTO;
+    uint256 constant M_MAX = 1_000_000_000_000_000_000_000_000 * 10 ** PRECISION_DECIMALS;
+    uint8 constant MAX_TOLERANCE = 5;
+    uint8 constant TOLERANCE_PRECISION = 11;
+    uint256 constant TOLERANCE_DEN = 10 ** TOLERANCE_PRECISION;
```
ShaneDuncan602, line 41: We probably need to get with Neel to determine our acceptable tolerances.

ShaneDuncan602 (changes requested):
This is fantastic work. I just have some questions about the libraries and how we might organize the curve data structures to work with all the calculators.

src/liquidity/ProtocolAMMCalculatorFactory.sol
oscarsernarosero: I had the exact same concern, but I [looked it up](https://viem.sh/docs/abi/encodeAbiParameters.html#more-examples) and it looks very easy.

src/liquidity/calculators/ProtocolNFTAMMCalcDualLinear.sol
oscarsernarosero: The problem I see with the ICurve interface is that these _validateSingleCurve functions will accept different types of parameters depending on the curve type of the AMM, and since Solidity is so strongly typed, I'm not sure how to achieve that. The idea of the library I created is to create a kind of a super type where all curves are sub types of it. So they all implement the getY() function to solve  ƒ(x). This is mostly to store math in a single file (Single Responsibility Principle) and to standardize the API for all curves. So all curves (Sigmoidal, quadratic, sinusoidal, etc) would have an overloaded version of the getY() function and of the fromInput() function (if needed). So they will "share" the same functions although not the logic.

test/liquidity/ProtocolNFTAMMCalcFuzz.t.sol
oscarsernarosero: Agree. I just went as low as I could. Our precision was +/-0.000000005% which I think is great.

test/liquidity/ProtocolNFTAMMCalcFuzz.t.sol
oscarsernarosero: ah ok. Yeah that's a great idea instead of the reversion I was doing.

src/liquidity/calculators/ProtocolNFTAMMCalcDualLinear.sol
ShaneDuncan602: I see. My concern was if the getY() functions had the same parameters you wouldn't be able to overload them and libraries can't have virtual functions so you couldn't override them. But you're planning on creating a separate struct for each type of curve regardless of their parameter similarities so this couldn't happen. Makes sense. What about adding new curve types post implementation? This would just require us to add the stuff to the libraries and redeploy the factory, right? As long as the factory is protocol owned we should be alright.

test/liquidity/ProtocolNFTAMMCalcFuzz.t.sol
ShaneDuncan602: Yeah, it's a little scary when the Kyber exploit was caused by a 10^(-13) difference.

test/helpers/Utils.sol
```diff
@@ -0,0 +1,73 @@
...
+    * @param y uint
+    * @return diff the absolute difference between *x* and *y*
+    */
+    function absoluteDiff(uint x, uint y) public pure returns (uint diff){
+        diff = x > y ? x - y : y - x;
+    }
+
+    /**
+    * @dev gets a bytes variable and checks if it is an ascii value or not.
+    * @notice this algorythm is 100% accurate in the negative case, but false
```
TJ-Everett: algorithm*

TJ-Everett (approved):
I agree with Shanes concern about the acceptable tolerances but otherwise this is incredibly well done. Great work and Approved for me.

ShaneDuncan602 (approved):
Great job, man. You knocked this out of the park.

src/liquidity/calculators/ProtocolNFTAMMCalcDualLinear.sol
oscarsernarosero: So in theory you can change what library to use, but doing it in practice has always been a mystery to me. I would say tho, it shouldn't be an issue since to add new functionality you would have to deploy a new calculator that supports it, and then you can deploy also your new library with it. So, it is not like we would depend on a library to be able to be upgraded.

PR #197, opened 2023-11-30
Add events to ProtocolApplicationHandler

TJ-Everett:
Add events to ProtocolApplicationHandler Rule activation functions

### DIFF
--- a/src/interfaces/IEvents.sol
+++ b/src/interfaces/IEvents.sol
@@ -71,6 +71,15 @@ interface IApplicationHandlerEvents {
     event ApplicationHandlerDeployed(address indexed deployedAddress, address indexed appManager);
     // Rule applied
     event ApplicationRuleApplied(bytes32 indexed ruleType, uint32 indexed ruleId);
+}
+
+/**
+ * @title Application Handler Events Interface
+ * @author @ShaneDuncan602 @oscarsernarosero @TJ-Everett
+ * @dev This library for all events in the Protocol module for the protocol. Each contract in the Protocol module should inherit this library for emitting events.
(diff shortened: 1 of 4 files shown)


src/interfaces/IEvents.sol
```diff
@@ -71,6 +71,10 @@ interface IApplicationHandlerEvents {
     event ApplicationHandlerDeployed(address indexed deployedAddress, address indexed appManager);
     // Rule applied
     event ApplicationRuleApplied(bytes32 indexed ruleType, uint32 indexed ruleId);
+    /// Rule deactivated
+    event ApplicationHandlerDeactivated(bytes32 indexed ruleType, address indexed handlerAddress);
+    /// Rule activated
+    event ApplicationHandlerActivated(bytes32 indexed ruleType, address indexed handlerAddress);
```
oscarsernarosero, line 86: Since we are duplicating these events, we should create an interface specifically for these and inherit them in both interfaces that use them. This way we only have one source of truth for these events (in the case we want to modify them later).

oscarsernarosero (approved):
This looks really good Tayler. I just have one suggestion, but since it is easy enough, I'll go ahead and approve so you don't have to wait for me tomorrow.

src/interfaces/IEvents.sol
ShaneDuncan602: Agreed. We want to avoid any event or error definition duplication.

ShaneDuncan602 (changes requested):
Just echoing Oscar's finding.

ShaneDuncan602 (approved):
Perfect

PR #198, opened 2023-12-01
Se 443 nft amm factory

oscarsernarosero:


src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -0,0 +1,427 @@
...
+        /// transfer NFTs from sender to the AMM
+        for(uint i; i < _tokenIds.length;){
+            _sendERC721WithConfirmation(_msgSender(), address(this), _tokenIds[i]);
+            emit AddLiquidity(address(ERC20Token), address (ERC721Token), 0, _tokenIds[i]);
+            unchecked{
+                ++i;
+            }
+        }
+        return true;
+    }
```
ShaneDuncan602, line 174: Great idea. This will be very useful.

src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -0,0 +1,427 @@
...
+    /**
+     * @dev Must provide the addresses for both tokens that will provide liquidity
+     * @param _ERC20Token valid ERC20 address
+     * @param _ERC721Token valid ERC721 address
+     * @param _appManagerAddress valid address of the corresponding app manager
+     * @param _calculatorAddress valid address of the corresponding calculator for the AMM
+     */
+    constructor(address _ERC20Token, address _ERC721Token, address _appManagerAddress, address _calculatorAddress) {
+        if(_ERC20Token == address(0) || _ERC721Token == address(0)) revert ZeroAddress();
+        if(!_isERC721Enumerable(_ERC721Token)) revert NotEnumerable();
```
ShaneDuncan602: I am not sure that they will always be enumerable. We have that requirement if the user wants total balance based rules but if they don't want to implement those rules then they don't have to have enumerable. This is likely a JB question. @jjbailey-io , should we have a restriction so that only ERC721Enumerables can be put into the AMM?

src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -0,0 +1,427 @@
...
+     * @return success pass/fail
+     */
+    function removeERC721(uint256 _tokenId) external appAdministratorOnly(appManagerAddress) returns (bool) {
+        /// we make sure we have the nft
+        _checkNFTOwnership(address(this), _tokenId);
+        /// transfer the tokens to the remover
+        _sendERC721WithConfirmation(address(this), _msgSender(), _tokenId);
+        emit RemoveLiquidity(address (ERC721Token), _tokenId);
+        return true;
+    }
```
ShaneDuncan602, line 204: If we keep the enumerable requirement, then we should add a function that will drain all the NFT's out of the AMM. Otherwise, if they want to shut down the AMM and retrieve the NFT's, they would have to know all the tokenId's and go one by one.

src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -0,0 +1,427 @@
...
+    function getBuyPrice() public view returns(uint256 price, uint256 fees){
+        price = calculator.simulateSwap(0, 0, 0, 1);
+        uint256 feesPct = handler.assessFees (ERC20Token.balanceOf(_msgSender()), ERC20Token.balanceOf( address(this)), _msgSender(), address(this), PCT_MULTIPLIER , ActionTypes.PURCHASE);
+        fees = _calculateBuyFeesFromPct(price, feesPct); 
+    }
+
+    /**
+    * @dev internal function to get the price for buying an NFT and also changing the state of the calculator
+    */
+    function _calculateBuyPrice() internal returns(uint256 price, uint256 fees){
```
ShaneDuncan602, line 267: Can this be a view function as well?

src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -0,0 +1,427 @@
...
+
+    /**
+    * @dev internal function to calculate the fees in a purchase by getting its price and fee percentage
+    * @param price the price for the NFT
+    * @param feesPct the percentage of fees to pay in a purchase
+    * @return fees to be paid at such price and at such percentage
+    */
+    function _calculateBuyFeesFromPct(uint256 price, uint256 feesPct) internal pure returns(uint256 fees){
+        fees = (price * PCT_MULTIPLIER) / (PCT_MULTIPLIER - feesPct ) - price;
+    }
```
ShaneDuncan602, line 298: Same, can these be view?

src/liquidity/calculators/ProtocolNFTAMMCalcDualLinear.sol
```diff
@@ -22,6 +22,8 @@ contract ProtocolNFTAMMCalcDualLinear is IProtocolAMMFactoryCalculator, CurveErr
     Line public buyCurve;
     Line public sellCurve;
 
+    uint256 public q;
```
ShaneDuncan602, line 31: Can we include a definition of what q is? It's apparent to us at the moment but might not be later or for outsiders looking at the code.

ShaneDuncan602 (changes requested):
I just have some super minor questions and suggestions. This is really, really good work. Excellent job, Oscar.

src/liquidity/ProtocolAMMFactory.sol
```diff
@@ -44,7 +60,20 @@ contract ProtocolAMMFactory is AppAdministratorOnly, IZeroAddressError, IAMMFact
...
+     * @dev This creates a linear AMM and calculation module.
+     * @param _ERC20Token valid ERC20 address
+     * @param _ERC721Token valid ERC721 address
+     * @param buyCurve LineInput for buy curve
+     * @param sellCurve LineInput for sell curve
+     * @param _appManagerAddress address of the application's appManager
+     * @return _calculatorAddress
+     */
+    function createDualLinearERC721AMM(address _ERC20Token, address _ERC721Token, LineInput memory buyCurve, LineInput memory sellCurve, address _appManagerAddress) external returns (address) {
+        return address(createERC721AMM(_ERC20Token, _ERC721Token, _appManagerAddress, address(protocolAMMCalculatorFactory.createDualLinearNFT(buyCurve, sellCurve, _appManagerAddress))));
```
TJ-Everett, line 76: I believe this is where the issue TronPy was having is originating, we should be firing an event here so TronPy (and I would assume DOOM eventually also) needs.
i.e. a ProtocolAMMCreated(address, token, token, app manager) 
This would go for lines 63 and 50 respectively.

TJ-Everett (changes requested):
This is amazing work Oscar! I have one question about events in the creation function for the new AMMs and then echoing Shane's Q for JB

src/liquidity/ProtocolERC721AMM.sol
oscarsernarosero: yeah. I only implemented that as I read it as part of the requirements. I don't mind taking it out. I'll just wait for JB's take on this.

src/liquidity/ProtocolERC721AMM.sol
oscarsernarosero: I see. Yeah. That's a good point, although the same algorithm that I would have to place inside the contract can be done offchain through a Python script or a Javascript script to save some gas, but would also be more inconvenient I guess. Maybe I could do a removeLiqInBatch?

src/liquidity/ProtocolERC721AMM.sol
oscarsernarosero: it can't because it calls a non view function `calculateSwap`. This is another reason we had to create a simulateSwap.

src/liquidity/ProtocolERC721AMM.sol
oscarsernarosero: same as above

src/liquidity/ProtocolERC721AMM.sol
jjbailey-io: I agree, unless the AMM has the explicit requirement of needing enumerable, I'd say we shouldn't have this check. If an AMM specific rule requires Enumerable, this check should occur at rule level (likely at enablement). We don't want to exclude a developer from using an NFT AMM because they didn't use enumerable when the AMM doesn't care.

src/liquidity/ProtocolAMMFactory.sol
oscarsernarosero: I agree, and I know this is a really easy fix. This, however, is out of scope, and I just don't want to mix tasks in this ticket which could lead to possibly spending more time than expected (which has already happened btw)  which is a blocker for other tickets.

TJ-Everett: Great point - I can gen a new ticket so we don't forget and this doesn't slow you down.

TJ-Everett: https://thracklelabs.atlassian.net/browse/SE-461

src/liquidity/ProtocolERC721AMM.sol
oscarsernarosero: Ok. Taking it out now. Maybe it would be a good idea to fix the requirements as well.

ShaneDuncan602 (approved):
Fantastic

TJ-Everett (approved):
Great work Oscar

PR #199, opened 2023-12-01
Se 386 create application configuration test script

ShaneDuncan602:


TJ-Everett (approved):
This is really cool Shane! I am approving but I do have a question, should we add a "how to run" section to the read me or somewhere? This is pretty new to me so if its there and I am a dumb dumb feel free to say so.

oscarsernarosero (changes requested):
Looks good Shane. Echoing Tayler's point, we should have a section in the README or at least the standard usage section in the header of each file. I would suggest having both.

ShaneDuncan602:
> This is really cool Shane! I am approving but I do have a question, should we add a "how to run" section to the read me or somewhere? This is pretty new to me so if its there and I am a dumb dumb feel free to say so.

I like this suggestion. My only concern is that putting references to it in the README will cause it to get pushed to Rules-Protocol. @jjbailey-io , do we want these scripts to be available all the way up the ladder or is this more for internal use? If it's internal, I can just add more comments to each script, otherwise, I can put them in the README.

jjbailey-io:
> > This is really cool Shane! I am approving but I do have a question, should we add a "how to run" section to the read me or somewhere? This is pretty new to me so if its there and I am a dumb dumb feel free to say so.
> 
> I like this suggestion. My only concern is that putting references to it in the README will cause it to get pushed to Rules-Protocol. @jjbailey-io , do we want these scripts to be available all the way up the ladder or is this more for internal use? If it's internal, I can just add more comments to each script, otherwise, I can put them in the README.

Everything in external goes directly into rules-protocol, so this is already going to be in rules-protocol so we should likely have it documented.

oscarsernarosero (approved):
Looks great Shane!

PR #205, opened 2023-12-05
SE-365-Create-a-Markdown-file-for-the-implementation-of-the-Account-Purchase-Controller-Rule

ShaneDuncan602:


docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,216 @@
...
+
+This rule works at the AMM level. It must be activated and configured for each desired AMM in the corresponding AMM handler. This rule will not be applied at the token level and will only be checked through the AMM swap function. 
+
+## Data Structure
+
+As this is a [tag](../GLOSSARY.md)-based rule, you can think of it as a collection of rules, where all "sub-rules" are independent from each other, and where each "sub-rule" is indexed by its tag. An account-purchase-controller "sub-rule" is specified by 3 variables:
+
+- **Purchase Amounts** (uint192): The maximum amount of tokens that may be purchased during the *purchase period*. 
+- **Purchase Periods** (uint16): The length of each time period for which the rule will apply, in hours.
+- **Starting Timestamps** (uint64): The timestamp of the date when the *period* starts counting.
```
oscarsernarosero: Unix timestamp*?

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,216 @@
...
+
+The rule will be evaluated with the following logic:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags along with the ID of the account-purchase-controller rule set in the token handler. 
+3. The processor will then try to retrieve the sub-rule associated with each tag.
+4. The processor will evaluate whether each sub-rule's period is active (if the current time is within `period` from the `starting timestamp`). If it is not within the period, it will set the cumulative purchases to the current purchase amount. If it is within the period, the processor will add the current purchase amount to the accrued purchase amount for the rule period.   
+5. The processor will then check if the cumulative purchases amount is greater than the `purchase amount` defined in the rule. If true, the transaction will revert. 
+6. Steps 4 and 5 are repeated for each of the account's tags.
+7. Return the cumulative purchases amount.
```
oscarsernarosero: Something I just remembered is that Anne did mentioned that this should always be in present tense. I forgot about the other doc. I think that's a good rule of thumb for this section in the docs.

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
+1. The account being evaluated will pass to the protocol all the tags it has registered to its address in the application manager.
+2. The processor will receive these tags along with the ID of the account-purchase-controller rule set in the token handler. 
+3. The processor will then try to retrieve the sub-rule associated with each tag.
+4. The processor will evaluate whether each sub-rule's period is active (if the current time is within `period` from the `starting timestamp`). If it is not within the period, it will set the cumulative purchases to the current purchase amount. If it is within the period, the processor will add the current purchase amount to the accrued purchase amount for the rule period.
```
oscarsernarosero: Im realizing that this statement has the same issue Tayler was having with another rule where there was basically a combination of 2 different checks. One is with the startTime and the other one is with the period. I would suggest to have [Tayler's work as reference](https://github.com/thrackle-io/tron/blob/internal/docs/userGuides/rules/TOKEN-TRANSFER-VOLUME.md#rule-evaluation)

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,216 @@
...
+3. The processor will then try to retrieve the sub-rule associated with each tag.
+4. The processor will evaluate whether each sub-rule's period is active (if the current time is within `period` from the `starting timestamp`). If it is not within the period, it will set the cumulative purchases to the current purchase amount. If it is within the period, the processor will add the current purchase amount to the accrued purchase amount for the rule period.   
+5. The processor will then check if the cumulative purchases amount is greater than the `purchase amount` defined in the rule. If true, the transaction will revert. 
+6. Steps 4 and 5 are repeated for each of the account's tags.
+7. Return the cumulative purchases amount.
+
+###### *see [ERC20TaggedRuleProcessorFacet](../../../src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol) -> checkPurchaseLimit*
+
+## Evaluation Exceptions 
+- This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction.
```
oscarsernarosero: Looking at the code, it seems that this rule doesn't apply for treasury, but it does apply for app administrators.

oscarsernarosero (changes requested):
I noticed a couple stuff that might also affect the SellPurchase-rule doc. Looks good in general.

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
TJ-Everett: +1 to this, it was a frustrating logic problem for me to unwrap so hopefully we don't have to go through that again

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,216 @@
...
+        - handlerAddress: the address of the asset handler where the rule has been applied.
+        - ruleId: the index of the rule created in the protocol by rule type.
+
+- **event ApplicationHandlerActivated(bytes32 indexed ruleType, address indexed handlerAddress)** emitted when an Account Purchase Controller rule has been activated in an asset handler:
+    - ruleType: "PURCHASE_LIMIT".
+    - handlerAddress: the address of the asset handler where the rule has been activated.
+
+## Dependencies
+
+- **Tags**: This rules relies on accounts having [tags](../GLOSSARY.md) registered in their [AppManager](../GLOSSARY.md), and they should match at least one of the tags in the rule for it to have any effect.
```
TJ-Everett: rule* 
I caught this in my doc and tried to correct all the ones in external that use tags.

TJ-Everett (changes requested):
Echoing Oscar for the eval steps and then a typo in the Dependencies section that seems to have copy pasta'ed its way around

TJ-Everett:
I also realized that we have not updated the RULES-GUIDE.md with the new rules. I am adding to min/max and Oracle branches as well

docs/userGuides/rules/RULE-GUIDE.md
```diff
@@ -16,20 +16,16 @@
 | Minimum/Maximum Account Balances | TODO |
```
TJ-Everett, line 16: I can't leave an inline comment on the lines but we need to add the Purchase and Sell links and purposes on line 9 and 10

TJ-Everett (changes requested):
Great work Shane. Just need to plug in the links in the Rule-Guide md

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
```diff
@@ -0,0 +1,219 @@
...
+
+1. The account is passed to the protocol with all the tags it has registered to its address in the application manager.
+2. The processor receives these tags along with the ID of the account-purchase-controller rule set in the token handler. 
+3. The processor retrieves the sub-rule associated with each tag.
+4. The processor evaluates whether the rule is active based on the `starting timestamp`. If it is not active, the rule aborts the next steps, and returns zero as the accrued cumulative purchases value.
+5. The processor evaluates whether the current time is within a new period.
+   -If it is a new period, the processor sets the cumulative purchases to the current purchase amount.
+   -If it is not a new period, the processor adds the current purchase amount to the accrued purchase amount for the rule period. 
+6. The processor checks if the cumulative purchases amount is greater than the `purchase amount` defined in the rule. If true, the transaction reverts. 
+8. Steps 4 and 5 are repeated for each of the account's tags.
```
oscarsernarosero: Should we mention that the default behavior will be to enforce the most restrictive sub rule in case of multiple tags matching?

oscarsernarosero (approved):
Looks good Shane. I only have one small question/suggestion, but it looks great, so Im approving.

docs/userGuides/rules/ACCOUNT-PURCHASE-CONTROLLER.md
oscarsernarosero: you also skipped step 7

TJ-Everett (approved):
Approved. Great work Shane.

PR #209, opened 2023-12-07
SE-404-Make-StartTimestamp-A-Rule-Level-Variable-In-Tagged-Rules

ShaneDuncan602:


TJ-Everett (approved):
Internal branch approved. Great work Shane and good job Oscar for identifying this.

src/economic/ruleProcessor/IRuleStorage.sol
```diff
@@ -17,13 +17,15 @@ interface IRuleStorage {
     struct PurchaseRuleS {
         /// ruleIndex => userType => rules
         mapping(uint32 => mapping(bytes32 => ITaggedRules.PurchaseRule)) purchaseRulesPerUser;
+        uint64 startTime; ///Time the rule is applied
```
oscarsernarosero: doesn't this make the startTime a protocol level variable?

src/economic/ruleProcessor/IRuleStorage.sol
```diff
@@ -17,13 +17,15 @@ interface IRuleStorage {
...
         uint32 purchaseRulesIndex; /// increments every time someone adds a rule
     }
 
     /// ******** Account Sell Rules ********
     struct SellRuleS {
         /// ruleIndex => userType => rules
         mapping(uint32 => mapping(bytes32 => ITaggedRules.SellRule)) sellRulesPerUser;
+        uint64 startTime; /// Time the rule is created
```
oscarsernarosero: same here

src/economic/ruleProcessor/TaggedRuleDataFacet.sol
```diff
@@ -29,43 +29,44 @@ contract TaggedRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents,
...
-            _startTimes[i].validateTimestamp();
-            data.purchaseRulesPerUser[index][_accountTypes[i]] = TaggedRules.PurchaseRule(_purchaseAmounts[i], _purchasePeriods[i], _startTimes[i]);
+            if (_purchaseAmounts[i] == 0 || _purchasePeriods[i] == 0) revert ZeroValueNotPermited();
+            data.purchaseRulesPerUser[index][_accountTypes[i]] = TaggedRules.PurchaseRule(_purchaseAmounts[i], _purchasePeriods[i]);
 
             unchecked {
                 ++i;
             }
         }
+        data.startTime = _startTime;
```
oscarsernarosero: this would mean that every time a rule changes the startTime, it would change for all rules in the protocol

src/economic/ruleProcessor/TaggedRuleDataFacet.sol
```diff
@@ -79,43 +80,44 @@ contract TaggedRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents,
...
             if (_accountTypes[i] == bytes32("")) revert BlankTag();
             if (_sellAmounts[i] == 0 || _sellPeriod[i] == 0) revert ZeroValueNotPermited();
-            _startTimes[i].validateTimestamp();
-            data.sellRulesPerUser[index][_accountTypes[i]] = TaggedRules.SellRule(_sellAmounts[i], _sellPeriod[i], _startTimes[i]);
+            data.sellRulesPerUser[index][_accountTypes[i]] = TaggedRules.SellRule(_sellAmounts[i], _sellPeriod[i]);
             unchecked {
                 ++i;
             }
         }
+        data.startTime = _startTime;
```
oscarsernarosero: same here

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -220,9 +220,9 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
         toTags.checkMaxTags();
         uint256 cumulativeTotal;
         for (uint i = 0; i < toTags.length; ) {
-            TaggedRules.PurchaseRule memory purchaseRule = getPurchaseRule(ruleId, toTags[i]);
+            (TaggedRules.PurchaseRule memory purchaseRule, uint64 startTime) = getPurchaseRule(ruleId, toTags[i]);
             if (purchaseRule.purchasePeriod > 0) {
-                if (purchaseRule.startTime.isWithinPeriod(purchaseRule.purchasePeriod, lastUpdateTime)) cumulativeTotal = purchasedWithinPeriod + amount;
+                if (startTime.isWithinPeriod(purchaseRule.purchasePeriod, lastUpdateTime)) cumulativeTotal = purchasedWithinPeriod + amount;
```
oscarsernarosero: the main idea is to place this line before the for loop since it is the same value across all sub rules. This is where the gas savings should come from

oscarsernarosero (changes requested):
There is a couple issues here:

1. The startTime is being saved at a protocol level. In other words, all rules of the same kind in the protocol will have the "same" startTime. The issue with this is that every time a rule is added, this startTime will be updated which will mess up the determinism of the rule.
2. The other issue is that the main purpose of the ticket was not addressed. The main purpose is to move the `startTime.isWithinPeriod()` check before the for loop so we can save gas in the processing. This check is still placed inside the loop which is unnecessary because the startTime doesn't depend on the sub-rules anymore.

src/economic/ruleProcessor/IRuleStorage.sol
ShaneDuncan602: Yes! Good catch.

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
ShaneDuncan602: We can't do that without making all the sub rules have the same period as well as startTime

oscarsernarosero: You're right. I guess we can still store the startTime locally before the for loop so we don't read from storage every time.

oscarsernarosero: You can also do a quick check if the startTime is in the future. In such case, you can return true for the check since the rule hasn't even started.

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -218,11 +218,12 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
      */
     function checkPurchaseLimit(uint32 ruleId, uint256 purchasedWithinPeriod, uint256 amount, bytes32[] calldata toTags, uint64 lastUpdateTime) external view returns (uint256) {
         toTags.checkMaxTags();
+        uint64 startTime = getPurchaseRuleStart(ruleId);
```
oscarsernarosero, line 221: I think a good idea would be to put right below this:
```
if(startTime > block.timestamp) 
    return true;
```
So we don't check for this rule if it hasn't even started yet

oscarsernarosero (approved):
Looks great Shane. I only have a small suggestion. I feel like there was another reason for this change, but I can't think of anything rn. It looks like it doesn't really save much gas. Maybe I just overestimated what this change would buy us. Great job tho!

PR #210, opened 2023-12-07
Changed the scope of the start time for minBalByDate rule

ShaneDuncan602:


TJ-Everett (approved):
External branch looks great! Approved. Also realize that this should have probably blocked my current ticket but no big deal at all to bring these changes in.

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -166,13 +166,14 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
      */
     function checkMinBalByDatePasses(uint32 ruleId, uint256 balance, uint256 amount, bytes32[] calldata toTags) external view {
         toTags.checkMaxTags();
+        uint64 startTime = getMinBalByDateRuleStart(ruleId);
```
oscarsernarosero, line 169: same as the other branch. Checkin if the rule hasn't started would save gas in such cases.

oscarsernarosero (approved):
Looks great. Same as the other one. I think checking for rule startingTime before the actual check can save a ton of gas if it hasn't started yet. Im approving since it's a small change.

PR #211, opened 2023-12-08
Se 403 tagged rules processing optimization by skipping empty rules

TJ-Everett:


src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -173,11 +173,14 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
                 if (toTags[i] != "") {
                     TaggedRules.MinBalByDateRule memory minBalByDateRule = getMinBalByDateRule(ruleId, toTags[i]);
                     uint256 holdPeriod = minBalByDateRule.holdPeriod;
-                    /// first check to see if still in the hold period
-                    if ((block.timestamp - (holdPeriod * 1 hours)) < startTime) {
-                        uint256 holdAmount = minBalByDateRule.holdAmount;
+                    uint256 holdAmount = minBalByDateRule.holdAmount;
+                    /// check if holdAmount is 0, 0 means it is an empty-rule/no-rule. a holdAmount should be greater than 0
+                    if(holdAmount != 0) {
```
oscarsernarosero: Is there a reason why we didn't use the `holdPeriod` that is already accessed?

oscarsernarosero (changes requested):
This looks good, but we need to do the same for all tagged rules

src/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
TJ-Everett: Great call - I will drop that back down a level and only call if we pass the holdPeriod check.

ShaneDuncan602 (approved):
Looks good.

PR #213, opened 2023-12-08
Se 461 add new amm created events to amm factory for UI tron py

TJ-Everett:
Added events to AMM Calculators 
Added AMMCalculatorAssigned event to AMMs


src/liquidity/ProtocolERC20AMM.sol
```diff
@@ -46,7 +46,7 @@ contract ProtocolERC20AMM is AppAdministratorOnly, IApplicationEvents,  AMMCalcu
         appManagerAddress = _appManagerAddress;
         /// Set the calculator and create the variable for it.
         _setCalculatorAddress(_calculatorAddress);
-        emit AMMDeployed(address(this));
+        emit AMMDeployed(address(this), _token0, _token1, _appManagerAddress);
```
oscarsernarosero: Should we add the `AMMCalculatorAssigned` right after this line?

src/liquidity/ProtocolERC721AMM.sol
```diff
@@ -48,7 +48,7 @@ contract ProtocolERC721AMM is AppAdministratorOnly, IERC721Receiver, IApplicatio
         appManagerAddress = _appManagerAddress;
         /// Set the calculator and create the variable for it.
         _setCalculatorAddress(_calculatorAddress);
-        emit AMMDeployed(address(this));
+        emit AMMDeployed(address(this), _ERC20Token, _ERC721Token, _appManagerAddress);
```
oscarsernarosero: same here?

src/liquidity/calculators/ProtocolAMMCalcCP.sol
```diff
@@ -21,6 +21,7 @@ contract ProtocolAMMCalcCP is IProtocolAMMFactoryCalculator {
     constructor(address _appManagerAddress) {
         if (_appManagerAddress == address(0)) revert ZeroAddress();
         appManagerAddress = _appManagerAddress;
+        emit AMMCalculatorDeployed();
```
oscarsernarosero: love it

oscarsernarosero (changes requested):
I love it! I just think we should add the AMMCalculatorAssigned event to the AMM constructor as well since that's the only way that DOOM can now what calculator the AMM is using (if there's been no update yet).

src/liquidity/ProtocolERC20AMM.sol
TJ-Everett: So I thought about that and actually had it there, then when I ran the tests it fires the event twice. Since it is firing from the _setcalculatorAddress() in the constructor. So it does fire at deployment in both types of AMMs and then it also fires when created from the factory as well.

oscarsernarosero: ah I see man. Perfect then. Very efficient.

oscarsernarosero (approved):
Great job Tayler!

ShaneDuncan602 (approved):
Looks great!

PR #219, opened 2023-12-18
Se 448 move all test setup logic to test common files

TJ-Everett:


test/util/TestArrays.sol
```diff
@@ -0,0 +1,386 @@
...
+
+
+
+    /****************** UINT256 ARRAY CREATION ******************
+    /**
+    * @dev This function creates a uint256 array to be used in tests 
+    * @notice This function creates a uint256 array size of 1 
+    * @return array uint256[] 
+    */
+    function createUint256SizeOneArray(uint256 arg1) public pure returns(uint256[] memory array) {
```
oscarsernarosero: I think it is better if you name the functions the same. I left an example in the [document](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/214401029/EVM+Protocol+Testing+Strategy+-+DRAFT#Creation-Of-Utility-Contracts). This way we only have one function name to create an array of size 1, 2 or 3.

oscarsernarosero (changes requested):
This looks great so far and it looks like it's going in the right direction. I left a comment to basically point to the fact that we should use polymorphism for the array-creation functions, and I think rule-creation encapsulation should be done in this ticket as explained in the ticket itself. We can do the setup process the way you proposed in the coming tickets, but the rule creation should be done here if we want to be efficient for the next tickets.

test/util/TestArrays.sol
TJ-Everett: I see your point and I don't necessarily disagree. I built out the unique names more to add explicit clarity in the test code. In my mind it was more clear this way. If we would rather have the same function name and just based on number of params I can work to change these.

TJ-Everett: @ShaneDuncan602 Can you be the tie breaker vote on polymorphism for this one? I am happy either way just need to know. 
Also @oscarsernarosero the rule creation encapsulation ticket is created and in the next sprint (planning session). Let me know if that satisfies what you wanted done in the this.

test/client/application/fuzz/ApplicationAppManagerFuzz.t.sol
```diff
@@ -1,25 +1,21 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.17;
 
-import "src/client/application/data/IPauseRules.sol";
 import "test/util/TestCommonFoundry.sol";
 
 contract ApplicationAppManagerFuzzTest is TestCommonFoundry {
-    bytes32 public constant SUPER_ADMIN_ROLE = ("SUPER_ADMIN_ROLE");
+
+    bytes32 public constant SUPER_ADMIN_ROLE = keccak256("SUPER_ADMIN_ROLE");
```
ShaneDuncan602: I think we would want all the role definitions in the TestCommon

test/client/application/integration/ApplicationAppManager.t.sol
```diff
@@ -1,66 +1,32 @@
...
-    ApplicationHandler public applicationHandler2;
     bytes32 public constant SUPER_ADMIN_ROLE = keccak256("SUPER_ADMIN_ROLE");
     bytes32 public constant USER_ROLE = keccak256("USER");
     bytes32 public constant APP_ADMIN_ROLE = keccak256("APP_ADMIN_ROLE");
     bytes32 public constant ACCESS_TIER_ADMIN_ROLE = keccak256("ACCESS_TIER_ADMIN_ROLE");
     bytes32 public constant RISK_ADMIN_ROLE = keccak256("RISK_ADMIN_ROLE");
-    bytes32 constant PROPOSED_SUPER_ADMIN_ROLE = keccak256("PROPOSED_SUPER_ADMIN_ROLE");
-    uint64 public constant TEST_DATE = 1666706998;
+    bytes32 public constant PROPOSED_SUPER_ADMIN_ROLE = keccak256("PROPOSED_SUPER_ADMIN_ROLE");
+
```
ShaneDuncan602: Same here. We shouldn't be redefining these roles.

test/client/application/fuzz/ApplicationAppManagerFuzz.t.sol
TJ-Everett: I tried this, because I thought the same thing, and it was blowing up. The role hashes expected by the test were not the same role hashes passed form common. I spent a good deal of time trying to troubleshoot but I can try again if we think that was a "me" issue.

test/client/token/ERC20/integration/ApplicationERC20.t.sol
```diff
@@ -1274,20 +1202,10 @@ contract ApplicationERC20Test is TestCommonFoundry {
...
-        holdAmounts[1] = uint256(2000 * (10 ** 18));
-        holdAmounts[2] = uint256(3000 * (10 ** 18));
-        uint16[] memory holdPeriods = new uint16[](3);
-        holdPeriods[0] = uint16(720); // one month
-        holdPeriods[1] = uint16(4380); // six months
-        holdPeriods[2] = uint16(17520); // two years
+        bytes32[] memory accs = createBytes32SizeThreeArray("Oscar","Tayler","Shane");
+        uint256[] memory holdAmounts = createUint256SizeThreeArray((1000 * (10 ** 18)), (2000 * (10 ** 18)), (3000 * (10 ** 18)));
+        // 720 = one month 4380 = six months 17520 = two years
+        uint16[] memory holdPeriods = createUint16SizeThreeArray(720, 4380, 17520);
```
ShaneDuncan602: These changes are so much cleaner. Good thinking.

test/client/token/ERC20/integration/ApplicationERC20Handler.t.sol
```diff
@@ -24,32 +11,14 @@ import "test/util/TestCommonFoundry.sol";
...
-    OracleDenied oracleDenied;
-    OracleAllowed oracleAllowed;
-    ApplicationERC20Handler applicationCoinHandlerSpecialOwner;
+
 
     function setUp() public {
         vm.warp(Blocktime);
         vm.startPrank(superAdmin);
-        setUpProtocolAndAppManagerAndTokens();
+        setUpProcotolAndCreateERC20AndHandlerSpecialOwner();
```
ShaneDuncan602, line 19: nice

test/client/token/ERC721/integration/ApplicationERC721U.t.sol
```diff
@@ -1,67 +1,15 @@
...
-    address rich_user = address(44);
-    address[] badBoys;
-    address[] goodBoys;
-    address proxyOwner = address(787);
 
     function setUp() public {
         vm.warp(Blocktime);
         vm.startPrank(superAdmin);
-        setUpProtocolAndAppManagerAndTokens();
+        setUpProtocolAndAppManagerAndTokensUpgradeable();
```
ShaneDuncan602, line 11: This kicks ass

test/util/TestArrays.sol
ShaneDuncan602: I agree with Oscar on this one. Sorry! I know that means more work, but I think the benefit of having one function handle all the sizes outweighs the explicit clarity. It just seems a lot cleaner.

ShaneDuncan602 (changes requested):
This looks really good. The tests are so much cleaner than before. I have a few suggestions. I think the role definitions should be moved to the common test files, I agree with Oscar on the array creation functions, and I think we need to look at DiamondTestUtil and RuleProcessorDiamondTestUtil. Can't we get rid of those now?

oscarsernarosero (approved):
Looks great Tayler

PR #220, opened 2023-12-19
index event Params

TJ-Everett:


oscarsernarosero (changes requested):
Looks great Tayler. We should take advantage of this ticket to fix the events to strictly have `uint256` in them instead of simply `uint`. The reason is that that small difference confuses people who need the signature of the event for the indexer for instance (Im telling you this because Rob already spent some time trying to figure out why the indexer was not picking up the new events). Having a `uint` in the code is perfectly fine, unless it is in an argument of a function, an error, or an event since that gives a misleading description of the signature. Everything else looks perfect.

ShaneDuncan602 (approved):
I agree with Oscar's comment. It seems like a good time to fix that as well. I'm going to go ahead and approve.

bfcoursewool:
> Looks great Tayler. We should take advantage of this ticket to fix the events to strictly have `uint256` in them instead of simply `uint`. The reason is that that small difference confuses people who need the signature of the event for the indexer for instance (Im telling you this because Rob already spent some time trying to figure out why the indexer was not picking up the new events). Having a `uint` in the code is perfectly fine, unless it is in an argument of a function, an error, or an event since that gives a misleading description of the signature. Everything else looks perfect.

Yeah, thanks Oscar! Since I'm usually only calculating keccak hashes for new/changed events, it's definitely helpful if I can just take what I see there and throw it directly into to hasher without having to think about any specific conversions of anything to anything else.

bfcoursewool:
Hey guys since this is approved and Oscar's change was added, I'm gonna go ahead and merge this right now because I need this stuff in the latest tron build in order to get the DOOM production environment working correctly. Hopefully doesn't ruffle any feathers...

PR #221, opened 2023-12-20
update example diagram and remove specific names

TJ-Everett:
Remove CHAMP and Prime Eternal from
ApplicationERC721U.t.sol


oscarsernarosero (approved):
Looks great Tayler

PR #222, opened 2023-12-20
Se 197 piecewise fungible token amm

oscarsernarosero:


src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,494 @@
...
+
+    /**
+     * @dev Set up the calculator and appManager for permissions
+     * @param _appManagerAddress appManager address
+     */
+    constructor(uint256 _x_trackerOffset, address _appManagerAddress) {
+        if (_appManagerAddress == address(0)) revert ZeroAddress();
+        appManagerAddress = _appManagerAddress;
+        x_trackerOffset = _x_trackerOffset;
+    }
```
TJ-Everett, line 57: We will want to either add in the calc type enum and fire the AMMCalcDeployed event or create a ticket to come back to this one and ensure all events and enums are firing for the UI in Q2

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,494 @@
...
+    function calculateSwap(uint256 _reserve0, uint256 _reserve1, uint256 _amount0, uint256 _amount1) external override returns (uint256 amountOut) {
+        amountOut = simulateSwap( _reserve0, _reserve1, _amount0, _amount1);  
+        if(_amount0 == 0)
+            x_tracker -= amountOut;
+        else
+            x_tracker += _amount0;
+    }
+
+    /**
+     * @dev This performs the swap from ERC20s to NFTs. It is a linear calculation.
```
TJ-Everett: Is this comment correct? Shouldn't this be that it is simulating a swap for the UI?

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,494 @@
...
+                    (regionOut, amountInLeft, amount0, amount1) = _calculateConstantProduct(_x_tracker, max_x, amountInLeft, amount0, amount1, i);
+                    amountOut += regionOut;
+                /// linear
+                }else if(sectionCurves[i].curveType == CurveTypes.LINEAR_FRACTION_B){
+                    (regionOut, amountInLeft, amount0, amount1) = _calculateLinear(_x_tracker, max_x, amountInLeft, amount0, amount1, i);
+                    amountOut += regionOut;
+                /// revert if type was none of the above
+                }else
+                    revert InvalidCurveType();
+                /// if we are in here, we modify i depending on the direction fo the swap
```
TJ-Everett: for *

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,494 @@
...
+    function updateAnUpperLimit(uint256 upperLimit, uint8 index) external appAdministratorOnly(appManagerAddress) {
+        if( ( (index > 0) && (upperLimit <= sectionUpperLimits[index - 1]) ) || 
+            ( (sectionUpperLimits.length > 0) && (upperLimit >= sectionUpperLimits[index + 1])) )
+                revert WrongArrayOrder();
+        sectionUpperLimits[index] = upperLimit;
+    }
+
+    /**
+    * @dev sets the whole upperLimits array at once
+    * @param upperLimits the array which is wished to be used as upperLimits
```
TJ-Everett: I do not think we should use the word "wish" here - seems not reassuring

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,494 @@
...
+    function addCurveToSection(SectionCurve calldata selectedCurve) external appAdministratorOnly(appManagerAddress){
+         _validateSectionCurve(selectedCurve);
+        sectionCurves.push(selectedCurve);
+    }
+
+    /**
+    * @dev sets a curve to a section
+    * @param selectedCurve of SectionCurve type where a pointer to a curve stored in the contract is set to a specific index of the sectionCurves list
+    */
+    function setCurveToSection( SectionCurve calldata selectedCurve, uint8 index) external appAdministratorOnly(appManagerAddress) {
```
TJ-Everett, line 236: Do we need a function to pop the last selectedCurve from the array as well?

src/common/IErrors.sol
```diff
@@ -114,11 +117,10 @@ interface IAppManagerErrors is INoAddressToRemove {
     error NotRegisteredHandler(address);
 }
 
-interface AMMCalculatorErrors {
+interface AMMCalculatorErrors is CommonInputErrors{
```
TJ-Everett, line 120: Love this!

TJ-Everett (changes requested):
This is some amazing work Oscar, I just have a couple of Questions in this one. Overall this is awesome and thank you for all the brain power this took.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: cool. I can create the event rn. Great catch.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: yes. Thank you!

src/client/liquidity/ProtocolAMMCalculatorFactory.sol
```diff
@@ -84,4 +85,14 @@ contract ProtocolAMMCalculatorFactory is AppAdministratorOnly, IZeroAddressError
         ProtocolAMMCalcSample01 protocolAMMCalcSample01 = new ProtocolAMMCalcSample01(_f_tracker, _g_tracker, _appManagerAddress);
         return address(protocolAMMCalcSample01);
     }
+
+    /**
+     * @dev This creates a concentrated liquidity with multiple curves calculator.
+     * @param _appManagerAddress address of the application's appManager
+     * @return _calculatorAddress
+     */
+    function createConcLiqMulCurves(address _appManagerAddress) external returns(address){
```
ShaneDuncan602: Is this really a "concentrated liquidity" calculator? The liquidity is still one bucket but the swaps happen under multiple curves.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: I don't think so because we took a builder approach which allows you to add but not to pop. So if you want to go from 3 to 2 sections, the contract wouldn't allow you to. You can simply have 2 sections with the same curve if that's what you want to do. This was thought as a first step I guess.

src/client/liquidity/ProtocolAMMCalculatorFactory.sol
oscarsernarosero: gosh it's not. I thought I had replaced all ConcLiq from the code. Great catch

TJ-Everett (approved):
Nice work again Oscar!

src/client/liquidity/calculators/ProtocolAMMCalcLinear.sol
```diff
@@ -59,9 +60,10 @@ contract ProtocolAMMCalcLinear is IProtocolAMMFactoryCalculator, IAMMCalculatorE
      * @return price
      */
     function simulateSwap(uint256 _reserve0, uint256 _reserve1, uint256 _amount0, uint256 _amount1) public view override returns (uint256) {
+        _reserve1;
```
ShaneDuncan602, line 63: I was thinking that we would remove the reserves from this function signature. Can we not do that?

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,497 @@
...
+        uint256 _reserve1, 
+        uint256 _amount0, 
+        uint256 _amount1
+    )
+    public 
+    view 
+    override 
+    returns (uint256 amountOut) 
+    {
+        _validateOneIsNotZero(_amount0, _amount1);
```
ShaneDuncan602, line 94: Shouldn't this be, validateBothNotZero?

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,497 @@
...
+        bool isAmount0Not0 = _amount0 != 0;
```
ShaneDuncan602: This boolean is only used twice. It seems unnecessary.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
```diff
@@ -0,0 +1,497 @@
...
+        x_trackerOffset = offset;
+    }
+
+    /**
+    * @dev Sets the equation variables
+    * @param _curve the definition of the linear ecuation
+    */
+    function addLinear(LinearInput memory _curve) external appAdministratorOnly(appManagerAddress) {
+        _validateSingleCurve(_curve);
+        linears.push(LinearFractionB(1,1,1,1));/// we add a dummy linear to be able to build it from library
```
ShaneDuncan602, line 163: I don't understand why this is necessary?

src/client/liquidity/calculators/dataStructures/CurveTypes.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.17;
+
+enum CurveTypes{
+    LINEAR_WHOLE_B,
+    LINEAR_FRACTION_B,
+    CONST_RATIO,
+    CONST_PRODUCT
```
ShaneDuncan602: Did we purposely leave Sample01 out of this?

ShaneDuncan602 (approved):
Really excellent work. I have some questions but I think this is ready to go. It's very cool.

src/client/liquidity/calculators/ProtocolAMMCalcLinear.sol
oscarsernarosero: Well, I thought about it, but we haven't really decided on that. I though it might be better to wait for the end of next quarter when we will have a better defined strategy of AMMs.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: No because you really need one to be zero and the other to be **not** zero. Otherwise it would mean that you are trying to swap on both directions at the same time.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: yeah. This I was hesitant to keep after a refactor I made where it basically wasn't really used anymore. I'll remove it.

src/client/liquidity/calculators/ProtocolAMMCalcMulCurves.sol
oscarsernarosero: This is because of the way we build the curve from the input. So we need to have a variable in storage, and then update that with the library creating the numerators and denominators. It is pretty straight forward with the other types of calculators because you have a single curve and not an array. So the variable exists already, but in this case, we have to create the variable first because it doesn't exist yet, and then we can build it from the input through the library with the actual values for numerators and denominators.

src/client/liquidity/calculators/dataStructures/CurveTypes.sol
oscarsernarosero: Yes, but I can add it. I just thought we were not sure about that one.

oscarsernarosero: Yes, but I can add it. I just thought we were not sure about that one.

oscarsernarosero: Im going with the name `Sample` since Tayler already has that name in the AMMType enum

PR #230, opened 2024-01-08
Multiple updates to support easier deployment.

mpetersoCode55:
1. Split the DeployAllModules into multiple scripts. Split between deplying facets and deploying the rule processor diamond. (verified this split works on mumbai)
2. Added a DeployExampleGame script that automates deploying locally and on an actual chain. Creates the contracts for the application (copies and modifies based on applicaiton name) deploys the protocol and the application (or just the application if the protocol is already deployed) and updates  the .env file for you.
3. Updated some documentation to specify the version of foundry needed.


.env
```diff
@@ -1,5 +1,32 @@
...
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+
```
ShaneDuncan602: I don't understand what this for? Are these the default addresses and they will get changed by the script? Also, are the duplicates supposed to be here?

.env
```diff
@@ -56,23 +83,23 @@ FOUNDRY_FUZZ_RUNS=8192
 
 ## Current Deployment details(these are used by the deploy scripts)
 # commit to repo with 0x0000000000000000000000000000000000000000 
-DEPLOYMENT_OWNER=0x0000000000000000000000000000000000000000
-DEPLOYMENT_OWNER_KEY=0x
+DEPLOYMENT_OWNER=0x15d34aaf54267db7d7c367839aaf71a00a2c6a65
+DEPLOYMENT_OWNER_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a
```
ShaneDuncan602: We need to keep these defaulted or the tests won't run correctly.

.env
```diff
@@ -56,23 +83,23 @@ FOUNDRY_FUZZ_RUNS=8192
...
 APP_ADMIN=0x75df377cb2fd583FDe95570766639E7B46c014cb
 APP_ADMIN_PRIVATE_KEY=0x
 RULE_ADMIN=0x
 ACCESS_TIER_ADMIN=0x
 RISK_ADMIN=0x
 # commit to repo with 0x0000000000000000000000000000000000000000
-DEPLOYMENT_RULE_PROCESSOR_DIAMOND=0x0000000000000000000000000000000000000000
+DEPLOYMENT_RULE_PROCESSOR_DIAMOND=0x88dD21924FbEeccC4dC8Ae3a0e207ecCA03c5b92
```
ShaneDuncan602: Same here.

.env
```diff
@@ -90,7 +117,50 @@ CHAIN_ID=31338
...
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
+APP_RULE_DATA=0x0B306BF915C4d645ff596e518fAf3F9669b97016
+FEE_RULE_DATA=0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1
```
ShaneDuncan602: I think something happened to this file. Something bad. Something unfortunate. Something that it won't be able to just wash off.

docs/userGuides/README.md
```diff
@@ -20,6 +20,11 @@ API documentation can be found [here](src/src/README.md).
 
 `pip install eth-abi`
 
+Note: Due to an issue with the latest version of foundry at the time of writing, we are currently using a specific foundry verion. To guarantee expected behavior please run the following command:
```
ShaneDuncan602: verion

.env
mpetersoCode55: These are the defaults from an anvil deployment, and yes the script will update them as you deploy. Good catch on the duplicates I'll remove them

.env
mpetersoCode55: Ok I'll shift them back to the defaults

.env
mpetersoCode55: Resolved.

.env
mpetersoCode55: Resolved.

.env
mpetersoCode55: Resolved.

.env
mpetersoCode55: Originally was piping the variables to the .env file, updated to use sed to update what was already there ... missed these to when I was removing the original piping

mpetersoCode55: Resolved.

docs/userGuides/README.md
mpetersoCode55: Resolved.

script/DeployAllModulesPt1.s.sol
```diff
@@ -0,0 +1,83 @@
...
+
+            /// Deploy the facet.
+            bytes memory bytecode = vm.getCode(string.concat(facet, ".sol"));
+            address facetAddress;
+            assembly {
+                facetAddress := create(0, add(bytecode, 0x20), mload(bytecode))
+            }
+        }
+    }
+}
```
ShaneDuncan602: Things would be much simpler if you deployed the diamond in this first script. I thought that was the plan so that you wouldn't have to retrieve all of those facet addresses. All you would need is to retrieve the Diamond address and then use that in the subsequent scripts. So essentially you would create the diamond, add some facets in the first script. Then the other scripts would just grab the diamond address and add more facets to it.

script/DeployExampleGame.sh
```diff
@@ -0,0 +1,473 @@
...
+echo "################################################################"
+echo Running foundryUp
+echo "################################################################"
+echo
+foundryup --version nightly-09fe3e041369a816365a020f715ad6f94dbce9f2 &> /dev/null
+
+echo "Is this a local deployment (y or n)?"
+read LOCAL
+
+if [ "$LOCAL" = "y" ]; then
```
ShaneDuncan602, line 34: I think we need to validate these entry variables. I know it's a little tedious, but currently a "Y" would make it try to deploy remotely. Maybe not all the variables will need it but probably the main ones.

mpetersoCode55: Do you think just forcing it to lower case is enough validation? Or checking for different variants (y, Y, yes, YES, etc)?

ShaneDuncan602 (changes requested):
This is really cool. I'm going to go ahead and set to "Request changes" until the deployment script gets reworked. You did a great job on this.

.env
oscarsernarosero: Just to document what we agreed on in the slack discussion, we still want to keep the rule-processor address here and update it through the script created in this ticket.

script/DeployAllModulesPt1.s.sol
mpetersoCode55: Resolved.

script/DeployExampleGame.sh
mpetersoCode55: Resolved.

script/DeployAllModules.s.sol
```diff
@@ -60,19 +60,19 @@ contract DeployAllModulesScript is Script {
...
+            // "ApplicationAccessLevelProcessorFacet",
+            // "ApplicationPauseProcessorFacet",
+            // "ERC20TaggedRuleProcessorFacet",
+            // "ERC721TaggedRuleProcessorFacet",
+            // "RiskTaggedRuleProcessorFacet",
+            // "RuleApplicationValidationFacet",
+            // "RuleDataFacet",
+            // "TaggedRuleDataFacet",
+            // "AppRuleDataFacet",
+            // "FeeRuleDataFacet"
```
TJ-Everett: Q: Was this left as commented out on purpose? 
If so, we should either remove the whole script file and have the make deployAll command run the new scripts and inform Rob / Brandon we have a new way of deploying.

script/DeployAllModulesPt1.s.sol
```diff
@@ -0,0 +1,102 @@
...
+import {RuleDataFacet} from "src/protocol/economic/ruleProcessor/RuleDataFacet.sol";
+
+/**
+ * @title The initial deployment script for the Protocol. It deploys the rule processor diamond and the initial set of facets.
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett, @mpetersoCode55
+ * @notice This contract deploys the rule processor diamond and the initial set of facets
+ * @dev This script will set contract addresses needed by protocol interaction in connectAndSetUpAll()
+ */
+
+contract DeployAllModulesScript is Script {
```
TJ-Everett: Bit of a nitpick but will help Brandon on the Tron side is to have the file name match the contract name so each one of the scripts should be Pt1, Pt2, etc.

TJ-Everett (approved):
Really great work overall Michael and I am glad we have these working both with deploying the diamond and calling diamond cut to upgrade. This will be key to the longevity of the protocol. 
1 question and some tediousness in the naming. Approved overall.

script/DeployAllModules.s.sol
mpetersoCode55: Good catch, meant to revert that before committing I'll fix it!

ShaneDuncan602: I agree, we need to keep this script working for DOOM.

mpetersoCode55: Resolved.

script/DeployAllModulesPt1.s.sol
mpetersoCode55: Resolved.

ShaneDuncan602 (approved):

docs/userGuides/README.md
```diff
@@ -20,6 +20,11 @@ API documentation can be found [here](src/src/README.md).
 
 `pip install eth-abi`
 
+Note: Due to an issue with the latest version of foundry at the time of writing, we are currently using a specific foundry version. To guarantee expected behavior please run the following command:
+```bash
+foundryup --version nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
+```
+
```
oscarsernarosero, line 27: nice

oscarsernarosero (approved):
Looks great Michael. Great job.

PR #231, opened 2024-01-09
Se 482 move amm only rules to use token transfer fields

oscarsernarosero:


docs/userGuides/rules/ACCOUNT-PURCHASE-RULE.md
```diff
@@ -0,0 +1,233 @@
...
+6. The processor checks if the cumulative purchases amount is greater than the `purchase amount` defined in the rule. If true, the transaction reverts.
+7. Steps 4 and 5 are repeated for each of the account's tags. In the case where multiple tags apply, the most restrictive is applied.
+8. Returns the cumulative purchases amount.
+
+###### *see [ERC20TaggedRuleProcessorFacet](../../../src/protocol/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol) -> checkPurchaseLimit*
+
+## Evaluation Exceptions 
+This rule doesn't apply when:
+- An approved Trading-Rule Whitelisted address is in the *to* side of the transaction.
+- ruleByapasser account is in the *from* or *to* side of the transaction.
```
ShaneDuncan602: ruleByapasser

docs/userGuides/rules/ACCOUNT-PURCHASE-RULE.md
```diff
@@ -0,0 +1,233 @@
...
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
```
ShaneDuncan602, line 78: I think we should add a step in here that says the handler determines if the requested transfer is a purchase.

docs/userGuides/rules/ACCOUNT-SELL-RULE.md
```diff
@@ -0,0 +1,228 @@
...
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
```
ShaneDuncan602, line 77: Same as above

docs/userGuides/rules/ACCOUNT-SELL-RULE.md
```diff
@@ -0,0 +1,228 @@
...
+5. The processor then checks if the cumulative sales amount is greater than the `sell amount` defined in the rule. If true, the transaction reverts. 
+6. Steps 4 and 5 are repeated for each of the account's tags.
+7. Return the cumulative sales amount.
+
+###### *see [ERC20TaggedRuleProcessorFacet](../../../src/protocol/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol) -> checkSellLimit*
+
+## Evaluation Exceptions 
+This rule doesn't apply when:
+- An approved Trading-Rule Whitelisted address is in the *from* side of the transaction.
+- ruleByapasser account is in the *from* or *to* side of the transaction.
```
ShaneDuncan602: ruleByapasser

docs/userGuides/rules/TOKEN-PERCENTAGE-PURCHASE.md
```diff
@@ -0,0 +1,227 @@
...
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
```
ShaneDuncan602, line 58: same as above

docs/userGuides/rules/TOKEN-PERCENTAGE-PURCHASE.md
```diff
@@ -0,0 +1,227 @@
...
+6. The processor evaluates if the final purchase percentage of total supply would be greater than the `token percentage`. 
+    - If yes, the transaction reverts. 
+    - If no, the processor returns the `token percentage` value for the current `period` to the handler.
+
+###### *see [ERC20RuleProcessorFacet](../../../src/protocol/economic/ruleProcessor/ERC20RuleProcessorFacet.sol) -> checkPurchasePercentagePasses*
+
+## Evaluation Exceptions 
+This rule doesn't apply when:
+- An approved Trading-Rule Whitelisted address is in the *to* side of the transaction.
+- ruleByapasser account is in the *from* or *to* side of the transaction.
```
ShaneDuncan602: ruleByapasser

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,227 @@
...
+- This rule can only be configured in the protocol by a **rule administrator**.
+- This rule can only be set in the asset handler by a **rule administrator**.
+- This rule can only be activated/deactivated in the asset handler by a **rule administrator**.
+- This rule can only be updated in the asset handler by a **rule administrator**.
+
+
+## Rule Evaluation
+
+The rule will be evaluated with the following logic:
+
```
ShaneDuncan602, line 58: same as above

docs/userGuides/rules/TOKEN-PERCENTAGE-SELL.md
```diff
@@ -0,0 +1,227 @@
...
+6. The processor evaluates if the final sell percentage of total supply would be greater than the `token percentage`. 
+    - If yes, the transaction reverts. 
+    - If no, the processor returns the `token percentage` value for the current `period` to the handler.
+
+###### *see [ERC20RuleProcessorFacet](../../../src/protocol/economic/ruleProcessor/ERC20RuleProcessorFacet.sol) -> checkSellPercentagePasses*
+
+## Evaluation Exceptions 
+This rule doesn't apply when:
+- An approved Trading-Rule Whitelisted address is in the *from* side of the transaction.
+- ruleByapasser account is in the *from* or *to* side of the transaction.
```
ShaneDuncan602: ruleByapasser

src/client/application/AppManager.sol
```diff
@@ -723,14 +727,13 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
...
      * @dev Check Application Rules for valid transactions.
-     * @param _action Action to be checked
      * @param _from address of the from account
      * @param _to address of the to account
      * @param _usdBalanceTo recepient address current total application valuation in USD with 18 decimals of precision
      * @param _usdAmountTransferring valuation of the token being transferred in USD with 18 decimals of precision
      */
-    function checkApplicationRules(ActionTypes _action, address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring) external onlyHandler {
-        applicationHandler.checkApplicationRules(_action, _from, _to, _usdBalanceTo, _usdAmountTransferring);
+    function checkApplicationRules( address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring) external onlyHandler {
```
ShaneDuncan602: We can't remove action from this function signature. One of this sprint's tickets is to categorize each rule by its applicability and only check them for the required action. So, if we determine one of the App Level rules only applies to Purchases, then this variable is the only way that can be done.

src/client/application/AppManager.sol
```diff
@@ -760,9 +763,7 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
         tokenToAddress[_token] = _tokenAddress;
         addressToToken[_tokenAddress] = _token;
         if(!isTokenRegistered[_tokenAddress]){  
-            tokenToIndex[_tokenAddress] = tokenList.length;  
-            tokenList.push(_tokenAddress);
-            isTokenRegistered[_tokenAddress] = true;
+            _addAddressWithMapping(tokenList, tokenToIndex, isTokenRegistered, _tokenAddress);
```
ShaneDuncan602, line 768: I agree that it's a good idea to continue modularization efforts to increase maintainability and I believe this is a good place for that. However, we have to be careful to what extent we do it for user based transaction operations. Every time we send parameters to a new function, it is creating new memory variables. I did some benchmarking in my last ticket and found that it was cheaper to change an array from a calldata to memory and modify or create a new array than it was to move processing logic to a separate function.

This is a good change though. I'm just putting that concern on our radar.

src/client/application/AppManager.sol
```diff
@@ -896,6 +912,34 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
         _removeAddressWithMapping(treasuryList, treasuryToIndex, isTreasuryRegistered, _treasuryAddress);
     }
 
+    /**
+     * @dev manage the whitelist for trading-rule bypasser accounts
+     * @param _address account in the list to manage
+     * @param isApproved set to true to indicate that _address can bypass trading rules.
+     */
+    function approveAddressToTradingRuleWhitelist(address _address, bool isApproved) external onlyRole(APP_ADMIN_ROLE) {
```
ShaneDuncan602, line 922: Why do we need the isApproved? I was thinking the existence on the list itself is all we need. If we remove the isApproved, it will work like the other registrations

src/client/token/ERC20/ProtocolERC20.sol
```diff
@@ -64,7 +64,7 @@ contract ProtocolERC20 is ERC20, ERC165, ERC20Burnable, ERC20FlashMint, Pausable
      */
     function _beforeTokenTransfer(address from, address to, uint256 amount) internal override whenNotPaused {
         /// Rule Processor Module Check
-        require(handler.checkAllRules(balanceOf(from), balanceOf(to), from, to, amount, ActionTypes.TRADE));
+        require(handler.checkAllRules(balanceOf(from), balanceOf(to), from, to, _msgSender(), amount));
```
ShaneDuncan602, line 67: I like the removal of the action type here. The token cannot determine the action, only the handler can.

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -69,15 +69,13 @@ contract ProtocolApplicationHandler is Ownable, RuleAdministratorOnly, IApplicat
...
-     * @param _action Action to be checked. This param is intentially added for future enhancements.
      * @param _from address of the from account
      * @param _to address of the to account
      * @param _usdBalanceTo recepient address current total application valuation in USD with 18 decimals of precision
      * @param _usdAmountTransferring valuation of the token being transferred in USD with 18 decimals of precision
      * @return success Returns true if allowed, false if not allowed
      */
-    function checkApplicationRules(ActionTypes _action, address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring) external onlyOwner returns (bool) {
-        _action;
+    function checkApplicationRules(address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring) external onlyOwner returns (bool) {
```
ShaneDuncan602, line 78: See above comment about action type requirement

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -164,10 +175,20 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        uint256 _balanceFrom, 
+        uint256 _balanceTo, 
+        address _from, 
+        address _to, 
+        address _sender,
+        uint256 _amount
+    ) 
+    internal 
+    {
+        _checkTaggedIndividualRules(_balanceFrom, _balanceTo, _from, _to, _sender, _amount);
```
ShaneDuncan602: I think something is happening with your formatter. We don't want these functions to accordion out because it increases our line count for audit.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -177,23 +198,49 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        address _from, 
+        address _to, 
+        address _sender,
+        uint256 _amount
+    ) 
+    internal 
+    {
+        bytes32[] memory toTags;
+        bytes32[] memory fromTags;
+        ActionTypes action = determineTransferAction(_from, _to, _sender);
```
ShaneDuncan602: nice

src/client/token/ProtocolHandlerCommon.sol
```diff
@@ -35,6 +48,294 @@ abstract contract ProtocolHandlerCommon is IAppManagerUser, IOwnershipErrors, IZ
...
+                    sellPercentageRuleId,  
+                    IERC20(msg.sender).totalSupply(), 
+                    _amount, 
+                    previousSellTime, 
+                    totalSoldWithinPeriod
+                );
+                previousSellTime = uint64(block.timestamp); /// update with new blockTime if rule check is successful
+            }
+        }
+    }
```
ShaneDuncan602: I'm not sure that we want to put rule checks and rule id's into the handler common. I think this is something that we should discuss and make sure that we're not missing something. I agree that it is cleaner for the time being but my concern would be future iterations and skinny handlers. @jjbailey-io , can we meet about this?

Also, what if we determine that one of the rules need to be handled differently between 20's and 721's and requires a different processor function. This has them coupled together. Additionally, we have to remember that we will likely add 1155 at some point too.

src/protocol/economic/IRuleProcessor.sol
```diff
@@ -73,7 +73,7 @@ interface IRuleProcessor {
      * @param fromTags Account tags applied to sender via App Manager
      * @param lastUpdateTime block.timestamp of most recent transaction from sender.
      */
-    function checkSellLimit(uint32 ruleId, uint256 salesWithinPeriod, uint256 amount, bytes32[] calldata fromTags, uint256 lastUpdateTime) external view returns (uint256);
+    function checkSellLimit(uint32 ruleId, uint256 salesWithinPeriod, uint256 amount, bytes32[] calldata fromTags, uint64 lastUpdateTime) external view returns (uint256);
```
ShaneDuncan602, line 76: Good catch.

docs/userGuides/rules/ACCOUNT-BALANCE-BY-RISK.md
```diff
@@ -68,7 +68,7 @@ The rule will be evaluated with the following logic:
 ###### *see [ApplicationRiskProcessorFacet](../../../src/protocol/economic/ruleProcessor/ApplicationRiskProcessorFacet.sol) -> checkAccBalanceByRisk*
 
 ## Evaluation Exceptions 
-- This rule doesn't apply when an **app administrator** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
+- This rule doesn't apply when an **ruleBypassAccount** address is in either the *from* or the *to* side of the transaction. This doesn't necessarily mean that if an app administrator is the one executing the transaction it will bypass the rule, unless the aforementioned condition is true.
```
TJ-Everett, line 71: Nice catch!

test/client/token/TestTokenCommon.sol
```diff
@@ -0,0 +1,37 @@
...
+    }
+    
+    function onERC721Received(address _operator, address _from, uint256 _tokenId, bytes calldata _data) external pure returns (bytes4) {
+        _operator;
+        _from;
+        _tokenId;
+        _data;
+        return this.onERC721Received.selector;
+    }
+}
```
ShaneDuncan602, line 37: super cool

ShaneDuncan602 (changes requested):
This was a lot of work and is very cool. I have some suggestions and concerns. I also think some of the modularization bears further discussion with the team so I'm going to try and set a meeting for it.

src/client/application/AppManager.sol
TJ-Everett: That is really good forward thinking Shane, part of this ticket's requirements were to remove this and I was also thinking it would be removed in my current ticket. I will make sure that assumption doesn't change anything in mine as well.

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -104,32 +96,44 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, RuleAdministra
...
+        uint256 _tokenId
+    ) 
+    external 
+    override 
+    onlyOwner 
+    returns (bool) 
+    {
         bool isFromBypassAccount = appManager.isRuleBypassAccount(_from);
         bool isToBypassAccount = appManager.isRuleBypassAccount(_to);
+        uint256 _amount = 1; /// currently not supporting batch NFT transactions. Only single NFT transfers.
```
TJ-Everett, line 109: Would this leave a way for batch transfers to still occur and only have the first tokenId checked for rule passing? Since the transaction doesn't revert for batch transfers.

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -104,32 +96,44 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, RuleAdministra
...
+            /// pure transfer rules
             uint128 balanceValuation;
             uint128 transferValuation;
             if (appManager.requireValuations()) {
                 balanceValuation = uint128(getAccTotalValuation(_to, nftValuationLimit));
                 transferValuation = uint128(nftPricer.getNFTPrice(msg.sender, _tokenId));
             }
-            appManager.checkApplicationRules(_action, _from, _to, balanceValuation, transferValuation);
-            _checkTaggedRules(_balanceFrom, _balanceTo, _from, _to, _amount, _tokenId);
+            appManager.checkApplicationRules( _from, _to, balanceValuation, transferValuation);
```
TJ-Everett: Just so it isn't missed, we will want this action type added back as well as in the appManager/appHandler

src/client/token/ERC721/upgradeable/ProtocolERC721U.sol
```diff
@@ -10,7 +10,7 @@ import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/CountersUpgradeable.sol";
-import "../IProtocolERC721Handler.sol";
+import "../../IProtocolTokenHandler.sol";
 import "../../ProtocolTokenCommonU.sol";
```
TJ-Everett, line 14: I have an outstanding question about import statement formatting in a PR with Shane so just noting here. I think we should standardize as a team on import format and location (Libraries, protocol, client, common, etc) to keep the repo nice and clean. 
May be worth spawning a ticket for this as part of the pre-audit work

TJ-Everett (changes requested):
Great work Oscar and really appreciate the clean up efforts that are in here. Looks like we never brought the event param changes over from internal and you caught a bunch of doc errors. 
Just have a question about the batch transfers in ERC721Handler

docs/userGuides/rules/ACCOUNT-PURCHASE-RULE.md
oscarsernarosero: Totally. I forgot to do that.

src/client/application/AppManager.sol
oscarsernarosero: I remember deciding to remove the action types from the repo. It is actually written in the ticket. I didn't remove the enum because Im still using it internally. On the other hand, I don't remember deciding to use it in the appManager, and I don't see the ticket you're mentioning Shane. So Im not sure what to do here since it is very clearly stated in this ticket.

src/client/application/AppManager.sol
oscarsernarosero: that is a great point Shane. In this case, I don't think we are adding much gas since we are not passing variables from the calldata (except for the _address), but mostly pointers to storage. However, I do think this is a great fact to keep in mind. Specially for arrays and stuff like that. I can take a look in Remix later tho.

src/client/application/AppManager.sol
oscarsernarosero: Well my thought was to have a single function for both add and remove. The isApproved is to tell if you are adding or removing. That way we don't bloat the code, more maintainability, less gas (only 20 units tho) in the overall consumption, etc. I based this in the setApprovalForAll function from the ERC721 standard

src/client/token/ERC20/ProtocolERC20Handler.sol
oscarsernarosero: ah ok. I did this manually actually. I'll fix it.

src/client/token/ProtocolHandlerCommon.sol
oscarsernarosero: I see your point. We can maybe create a separate contract then. Something like HandlerCommonTradingRules since I do think we should avoid code duplication, and that was one of ToB things.

src/protocol/economic/IRuleProcessor.sol
oscarsernarosero: That was actually @TJ-Everett . I spent hours trying to understand why my code wasn't working. It was this bad boy here.

test/client/token/TestTokenCommon.sol
oscarsernarosero: thank you

src/client/token/ERC721/ProtocolERC721Handler.sol
oscarsernarosero: Not really. It was decided that we are not currently supporting batch transactions, and I thought that leaving the door open for something that seems still far in the future was creating more problems than necessary. So I decided to shut the door for rn. We can open it back when we think we decide to go for it.

src/client/token/ERC721/ProtocolERC721Handler.sol
oscarsernarosero: This ticket stated that it needed to be removed. I couldn't find the ticket Shane was talking about. Still, if it is decided to use it in the appManager, it can be done in that ticket. I just don't want to fall in the trap of adding and removing something back and forth forever.

src/client/token/ERC721/upgradeable/ProtocolERC721U.sol
oscarsernarosero: I agree Tayler. That's a great point. I know we had previously agreed on absolute paths,  but since the npm package works with relatives, we went back to relatives. So I decided to stick with relatives. But I agree we should all be on the same page on what to do moving forward.

src/client/application/AppManager.sol
ShaneDuncan602: It's the ticket I'm currently working. Its scope doesn't cover the app level rules yet but that would be step 2. We don't need the action types in the call from the token to the handler but I think we will have to have them in the call to checkAppRules.

src/client/token/ProtocolHandlerCommon.sol
ShaneDuncan602: I like that idea. It still gives us the modularity but without the overt dependency when we don't need it.

src/client/token/ERC721/upgradeable/ProtocolERC721U.sol
TJ-Everett: Ahh yeah that was the reasoning, I couldn't remember why we were using absolute. I agree and do not have a preference either way - just picking and sticking one works for me.

src/client/application/AppManager.sol
oscarsernarosero: Ok Shane. I'll put it back, but for future reference, we should add those specifics to the tickets. Even Tayler though that we were getting rid entirely of the actions. A whole workflow was thought out and implemented around the fact that we wouldn't need actions anymore. Luckily I just checked and that's not too much of a biggie in this case.

ShaneDuncan602 (approved):
Fantastic

TJ-Everett (approved):
Looks amazing Oscar. Great work

src/client/application/IAppManager.sol
```diff
@@ -146,15 +146,28 @@ interface IAppManager is IAppManagerErrors, IPermissionModifierErrors, IInputErr
...
     /**
      * @dev Jump through to the gobal rules to see if the requested action is valid.
-     * @param _action Action to be checked
      * @param _from address of the from account
      * @param _to address of the to account
      * @param _usdBalanceTo recepient address current total application valuation in USD with 18 decimals of precision
      * @param _usdAmountTransferring valuation of the token being transferred in USD with 18 decimals of precision
      */
-    function checkApplicationRules(ActionTypes _action, address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring) external;
+    function checkApplicationRules(address _from, address _to, uint128 _usdBalanceTo, uint128 _usdAmountTransferring, ActionTypes action) external;
```
mpetersoCode55, line 170: Don't we still need a @param for action here?

src/protocol/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
```diff
@@ -208,4 +208,105 @@ contract ERC20RuleProcessorFacet is IInputErrors, IRuleProcessorErrors, IERC20Er
...
+        uint32 ruleId,
+        uint256 currentTotalSupply,
+        uint256 amountToTransfer,
+        uint64 lastPurchaseTime,
+        uint256 purchasedWithinPeriod
+    ) external view returns (uint256) {
+        NonTaggedRules.TokenPercentagePurchaseRule memory percentagePurchaseRule = getPctPurchaseRule(ruleId);
+        uint256 totalPurchasedWithinPeriod = amountToTransfer; /// resets value for purchases outside of purchase period
+        uint256 totalSupply = percentagePurchaseRule.totalSupply;
+        /// check if totalSupply in rule struct is 0 and if it is use currentTotalSupply, if < 0 use rule value
```
mpetersoCode55: Should this read if > 0?

src/protocol/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
```diff
@@ -208,4 +208,105 @@ contract ERC20RuleProcessorFacet is IInputErrors, IRuleProcessorErrors, IERC20Er
...
+     * @param currentTotalSupply total supply value passed in by the handler. This is for ERC20 tokens with a fixed total supply.
+     * @param amountToTransfer total number of tokens to be transferred in transaction.
+     * @param lastSellTime time of the most recent purchase from AMM. This starts the check if current transaction is within a purchase window.
+     * @param soldWithinPeriod total amount of tokens sold within current period
+     */
+    function checkSellPercentagePasses(uint32 ruleId, uint256 currentTotalSupply, uint256 amountToTransfer, uint64 lastSellTime, uint256 soldWithinPeriod) external view returns (uint256) {
+        NonTaggedRules.TokenPercentageSellRule memory percentageSellRule = getPctSellRule(ruleId);
+        uint256 totalSoldWithinPeriod = amountToTransfer; /// resets value for purchases outside of purchase period
+        uint256 totalSupply = percentageSellRule.totalSupply;
+        /// check if totalSupply in rule struct is 0 and if it is use currentTotalSupply, if < 0 use rule value
```
mpetersoCode55: same question as above

mpetersoCode55 (approved):

src/client/application/IAppManager.sol
oscarsernarosero: I moved it to the end. I had removed it first, and put it back. The other functions have it at the end so I wanted them to have parity there.

src/protocol/economic/ruleProcessor/ERC20RuleProcessorFacet.sol
oscarsernarosero: I think you're right. Great catch!

src/client/application/IAppManager.sol
mpetersoCode55: Doesn't it still need a @param in the comment above even though it's at the end?

TJ-Everett: @mpetersoCode55 I needed these in external as a blocker to my ticket but I promise this will be added as a param in there.

PR #232, opened 2024-01-09
SE-487-AMM-Fee-clean-up

ShaneDuncan602:


src/client/liquidity/ProtocolAMMHandler.sol
```diff
@@ -104,27 +109,150 @@ contract ProtocolAMMHandler is Ownable, ProtocolHandlerCommon, IProtocolAMMHandl
...
+                    }
+                }
+            }
+        }
+        // if the total fees - discounts is greater than 100 percent, revert
+        if (totalFeePercent - int24(discount) > 10000) {
+            revert FeesAreGreaterThanTransactionAmount(_from);
+        }
+        return (feeCollectorAccounts, feePercentages);
+    }
```
oscarsernarosero: could we create a contract to place this function and inherit from it since this function can be found in multiple contracts?

src/client/liquidity/ProtocolAMMHandler.sol
```diff
@@ -104,27 +109,150 @@ contract ProtocolAMMHandler is Ownable, ProtocolHandlerCommon, IProtocolAMMHandl
...
+        _tags = new bytes32[](_fromTags.length+1);
+        // copy the array to larger one
+        for (uint i; i < _fromTags.length; ) {
+            _tags[i] = _fromTags[i];
+            unchecked {
+                ++i;
+            }
+        }
+        // append blank tag
+        _tags[_fromTags.length] = BLANK_TAG;
```
oscarsernarosero: same with thisone

oscarsernarosero (approved):
Looks great Shane. I only have one suggestion.

src/client/liquidity/ProtocolAMMHandler.sol
```diff
@@ -7,6 +7,7 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "src/protocol/economic/IRuleProcessor.sol";
 import "./IProtocolAMMHandler.sol";
+import "../token/data/Fees.sol";
```
TJ-Everett: I have been trying to catch these import statements as they pop up and standardize to the  absolute path (I think is the right term) instead of some having ../, we always use src/path.  We should standardize this across the team to help keep the imports formatted the same.

src/client/liquidity/ProtocolAMMHandler.sol
TJ-Everett: Should we drop it into handler common then? Instead of a new contract?

TJ-Everett (approved):
Great work on this one Shane. I have the same Q that Oscar does about dropping the fee functions into the HandlerCommon. Then a Question about import formats but that doesn't need to hold up the ticket at all. So approved for me and I will monitor the resolution on fee function locations.

src/client/liquidity/ProtocolAMMHandler.sol
ShaneDuncan602: Good idea. It's a helper function and should be in common now.

mpetersoCode55 (approved):

PR #234, opened 2024-01-10
Updated asset handlers to allow for multiple oracle rules (up to 10)

mpetersoCode55:
Implements [SE-529](https://thracklelabs.atlassian.net/browse/SE-529)

[SE-529]: https://thracklelabs.atlassian.net/browse/SE-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/token/ERC20/ProtocolERC20Handler.sol
+++ b/src/client/token/ERC20/ProtocolERC20Handler.sol
@@ -422,9 +422,9 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
                 oracleRules[oracleRuleIndex].oracleRuleActive = _on;
 
                 if (_on) {
-                    emit ApplicationHandlerActivated(ORACLE, address(this));
+                    emit ApplicationHandlerActivated(ORACLE);
                 } else {
-                    emit ApplicationHandlerDeactivated(ORACLE, address(this));
+                    emit ApplicationHandlerDeactivated(ORACLE);
                 }
             }
             unchecked {
--- a/src/client/token/ERC20/ProtocolERC20Handler.sol
+++ b/src/client/token/ERC20/ProtocolERC20Handler.sol
@@ -25,9 +25,14 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
     Fees fees;
     bool feeActive;
 
+    struct OracleRule {
+        uint32 oracleRuleId;
+        bool oracleRuleActive;
+    }
+
     /// RuleIds
     uint32 private minTransferRuleId;
-    uint32 private oracleRuleId;
+    OracleRule[] private oracleRules;
     uint32 private minMaxBalanceRuleId;
     uint32 private adminWithdrawalRuleId;
     uint32 private minBalByDateRuleId;
@@ -36,12 +41,13 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
 
     /// on-off switches for rules
     bool private minTransferRuleActive;
-    bool private oracleRuleActive;
     bool private minMaxBalanceRuleActive;
     bool private adminWithdrawalActive;
     bool private minBalByDateRuleActive;
     bool private tokenTransferVolumeRuleActive;
     bool private totalSupplyVolatilityRuleActive;
+
+    uint16 constant MAX_ORACLE_RULES = 10;
 
     /**
      * @dev Constructor sets params
@@ -56,6 +62,7 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
         appManagerAddress = _appManagerAddress;
         appManager = IAppManager(_appManagerAddress);
         ruleProcessor = IRuleProcessor(_ruleProcessorProxyAddress);
+
         transferOwnership(_assetAddress);
         if (!_upgradeMode) {
             deployDataContract();
@@ -112,7 +119,13 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      */
     function _checkNonTaggedRules(address _from, address _to, uint256 _amount) internal {
         if (minTransferRuleActive) ruleProcessor.checkMinTransferPasses(minTransferRuleId, _amount);
-        if (oracleRuleActive) ruleProcessor.checkOraclePasses(oracleRuleId, _to);
+
+        for (uint256 oracleRuleIndex; oracleRuleIndex < oracleRules.length; ) {
+            if (oracleRules[oracleRuleIndex].oracleRuleActive) ruleProcessor.checkOraclePasses(oracleRules[oracleRuleIndex].oracleRuleId, _to);
+            unchecked {
+                ++oracleRuleIndex;
+            }
+        }
         if (tokenTransferVolumeRuleActive) {
             transferVolume = ruleProcessor.checkTokenTransferVolumePasses(tokenTransferVolumeRuleId, transferVolume, IToken(msg.sender).totalSupply(), _amount, lastTransferTs);
             lastTransferTs = uint64(block.timestamp);
@@ -386,22 +399,37 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      * @param _ruleId Rule Id to set
      */
     function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        if (oracleRules.length >= MAX_ORACLE_RULES) {
+            revert OracleRulesPerAssetLimitReached();
+        }
         ruleProcessor.validateOracle(_ruleId);
-        oracleRuleId = _ruleId;
-        oracleRuleActive = true;
+
+        OracleRule memory newEntity;
+        newEntity.oracleRuleId = _ruleId;
+        newEntity.oracleRuleActive = true;
+        oracleRules.push(newEntity);
         emit ApplicationHandlerApplied(ORACLE, _ruleId);
     }
 
     /**
      * @dev enable/disable rule. Disabling a rule will save gas on transfer transactions.
      * @param _on boolean representing if a rule must be checked or not.
+     * @param ruleId the id of the rule to activate/deactivate
      */
-    function activateOracleRule(bool _on) external ruleAdministratorOnly(appManagerAddress) {
-        oracleRuleActive = _on;
-        if (_on) {
-            emit ApplicationHandlerActivated(ORACLE);
-        } else {
-            emit ApplicationHandlerDeactivated(ORACLE);
+    function activateOracleRule(bool _on, uint32 ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        for (uint256 oracleRuleIndex; oracleRuleIndex < oracleRules.length; ) {
+            if (oracleRules[oracleRuleIndex].oracleRuleId == ruleId) {
+                oracleRules[oracleRuleIndex].oracleRuleActive = _on;
+
+                if (_on) {
+                    emit ApplicationHandlerActivated(ORACLE, address(this));
+                } else {
+                    emit ApplicationHandlerDeactivated(ORACLE, address(this));
+                }
+            }
+            unchecked {
+                ++oracleRuleIndex;
+            }
         }
     }
 
@@ -409,18 +437,56 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      * @dev Retrieve the oracle rule id
      * @return oracleRuleId
      */
-    function getOracleRuleId() external view returns (uint32) {
-        return oracleRuleId;
+    function getOracleRuleIds() external view returns (uint32[] memory ) {
+        uint32[] memory ruleIds = new uint32[](oracleRules.length);
+        for (uint256 oracleRuleIndex; oracleRuleIndex < oracleRules.length; ) {
+            ruleIds[oracleRuleIndex] = oracleRules[oracleRuleIndex].oracleRuleId;
+            unchecked {
+                ++oracleRuleIndex;
+            }
+        }
+        return ruleIds;
     }
 
     /**
      * @dev Tells you if the Oracle Rule is active or not.
+     * @param ruleId the id of the rule to check
      * @return boolean representing if the rule is active
      */
-    function isOracleActive() external view returns (bool) {
-        return oracleRuleActive;
+    function isOracleActive(uint32 ruleId) external view returns (bool) {
+        for (uint256 oracleRuleIndex; oracleRuleIndex < oracleRules.length; ) {
+            if (oracleRules[oracleRuleIndex].oracleRuleId == ruleId) {
+                return oracleRules[oracleRuleIndex].oracleRuleActive;
+            }
+            unchecked {
+                ++oracleRuleIndex;
+            }
+        }
+        return false;
     }
 
+    /**
+     * @dev Removes an oracle rule from the list.
+     * @param ruleId the id of the rule to remove
+     */
+    function removeOracleRule(uint32 ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        OracleRule memory lastId = oracleRules[oracleRules.length -1];
+        if(ruleId != lastId.oracleRuleId){
+            uint index = 0;
+            for (uint256 oracleRuleIndex; oracleRuleIndex < oracleRules.length; ) {
+                if (oracleRules[oracleRuleIndex].oracleRuleId == ruleId) {
+                    index = oracleRuleIndex; 
+                    break;
+                }
+                unchecked {
+                    ++oracleRuleIndex;
+                }
+            }
+            oracleRules[index] = lastId;
+        }
+
+        oracleRules.pop();
+    }
 
     /**
      * @dev Set the AdminWithdrawalRule. Restricted to app administrators only.
@@ -638,14 +704,14 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      * @dev This function is used to propose the new owner for data contracts.
      * @param _newOwner address of the new AppManager
      */
-    function proposeDataContractMigration(address _newOwner) external appAdministratorOnly(appManagerAddress) {
+    function proposeDataContractMigration(address _newOwner) external appAdministratorOrOwnerOnly(appManagerAddress) {
         fees.proposeOwner(_newOwner);
     }
 
     /**
      * @dev This function is used to confirm this contract as the new owner for data contracts.
      */
-    function confirmDataContractMigration(address _oldHandlerAddress) external appAdministratorOnly(appManagerAddress) {
+    function confirmDataContractMigration(address _oldHandlerAddress) external appAdministratorOrOwnerOnly(appManagerAddress) {
         ProtocolERC20Handler oldHandler = ProtocolERC20Handler(_oldHandlerAddress);
         fees = Fees(oldHandler.getFeesDataAddress());
         fees.confirmOwner();
--- a/src/client/token/ERC721/ProtocolERC721Handler.sol
+++ b/src/client/token/ERC721/ProtocolERC721Handler.sol
@@ -289,9 +289,9 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
                 oracleRules[oracleRuleIndex].oracleRuleActive = _on;
 
                 if (_on) {
-                    emit ApplicationHandlerActivated(ORACLE, address(this));
+                    emit ApplicationHandlerActivated(ORACLE);
                 } else {
-                    emit ApplicationHandlerDeactivated(ORACLE, address(this));
+                    emit ApplicationHandlerDeactivated(ORACLE);
                 }
             }
             unchecked {
@@ -591,7 +591,7 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
      *@param _token address of the token to call totalSupply() of.
      */
     function getTotalSupply(address _token) internal view returns (uint256) {
-        return IERC20(_token).totalSupply();
+        return IERC20Decimals(_token).totalSupply();
     }
 
     /// -------------SIMPLE RULE SETTERS and GETTERS---------------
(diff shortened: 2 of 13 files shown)


docs/src/src/client/token/ERC20/ProtocolERC20Handler.sol/contract.ProtocolERC20Handler.md
```diff
@@ -630,13 +630,14 @@ function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManag
```
TJ-Everett: Anything in the dcos/src/src directory is auto generated by foundry based on the natspec commentary - so we don't update these as they will auto update through github actions

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -36,9 +37,16 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
     Fees fees;
     bool feeActive;
 
+    struct OracleRuleStruct {
```
oscarsernarosero: I think just naming this OracleRule would be a good name.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -36,9 +37,16 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        address oracleAddress;
+        uint32 oracleRuleId;
+        bool oracleRuleActive;
+    }
+
     /// RuleIds
     uint32 private minTransferRuleId;
-    uint32 private oracleRuleId;
+    OracleRuleStruct[] private oracleRuleIds;
```
oscarsernarosero: I know you were trying to keep the naming patter, but since this rule is basically breaking the pattern, I think it's ok if you name this differently. Maybe `oracleRules`

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -420,18 +431,34 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
      * @param _ruleId Rule Id to set
      */
     function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        if (nextOracleRuleIndex >= 10) {
```
oscarsernarosero: 10 must be set in a constant/variable. We shouldn't have magic numbers

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -36,9 +37,16 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
     Fees fees;
     bool feeActive;
 
+    struct OracleRuleStruct {
+        address oracleAddress;
```
oscarsernarosero: also, I don't think you need this variable. That's supposed to be stored at the protocol level.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -36,9 +37,16 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        uint32 oracleRuleId;
+        bool oracleRuleActive;
+    }
+
     /// RuleIds
     uint32 private minTransferRuleId;
-    uint32 private oracleRuleId;
+    OracleRuleStruct[] private oracleRuleIds;
+    uint32 private nextOracleRuleIndex;
```
oscarsernarosero: could this simply be extracted from the array length?

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -139,7 +147,10 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
      */
     function _checkNonTaggedRules(address _from, address _to, uint256 _amount) internal {
         if (minTransferRuleActive) ruleProcessor.checkMinTransferPasses(minTransferRuleId, _amount);
-        if (oracleRuleActive) ruleProcessor.checkOraclePasses(oracleRuleId, _to);
+
+        for (uint256 oracleRuleIndex = 0; oracleRuleIndex < nextOracleRuleIndex; oracleRuleIndex++) {
+            if (oracleRuleIds[oracleRuleIndex].oracleRuleActive) ruleProcessor.checkOraclePasses(oracleRuleIds[oracleRuleIndex].oracleRuleId, _to);
+        }
```
oscarsernarosero: nice. This is perfect. Only thing is that to save gas, for loops in Solidity should have the increment of the iterator in an `unchecked` block at the end of the body. You can look for one in the repo for an example. This is not necessary in the latest Solidity version, but we still using 0.8.17 so we still need it.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -420,18 +431,34 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
      */
     function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        if (nextOracleRuleIndex >= 10) {
+            revert OracleRulesPerAssetLimitReached();
+        }
         ruleProcessor.validateOracle(_ruleId);
-        oracleRuleId = _ruleId;
-        oracleRuleActive = true;
+
+        address oracleAddress = ERC20RuleProcessorFacet(address(ruleProcessor)).getOracleRule(_ruleId).oracleAddress;
```
oscarsernarosero: We shouldn't be doing this. We don't really need the oracleAddress. That's what the protocol is for.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -420,18 +431,34 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
     }
 
     /**
      * @dev enable/disable rule. Disabling a rule will save gas on transfer transactions.
      * @param _on boolean representing if a rule must be checked or not.
+     * @param oracleAddress the address of the oracle contract
      */
-    function activateOracleRule(bool _on) external ruleAdministratorOnly(appManagerAddress) {
-        oracleRuleActive = _on;
+    function activateOracleRule(bool _on, address oracleAddress) external ruleAdministratorOnly(appManagerAddress) {
```
oscarsernarosero: If you get rid of the oracleAddress, you would need the ruleId here.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -420,18 +431,34 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        for (uint256 oracleRuleIndex = 0; oracleRuleIndex < nextOracleRuleIndex; oracleRuleIndex++) {
```
oscarsernarosero: same about the iterator as above

docs/src/src/client/token/ERC20/ProtocolERC20Handler.sol/contract.ProtocolERC20Handler.md
mpetersoCode55: That makes so much more sense, I kept thinking man this is a pain

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -441,18 +468,30 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
 
     /**
      * @dev Retrieve the oracle rule id
+     * @param oracleAddress the address of the oracle contract
      * @return oracleRuleId
      */
-    function getOracleRuleId() external view returns (uint32) {
-        return oracleRuleId;
+    function getOracleRuleId(address oracleAddress) external view returns (uint32) {
```
oscarsernarosero: I would ask for an index in the array instead of an oracleAddress

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -441,18 +468,30 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
     }
 
     /**
      * @dev Tells you if the Oracle Rule is active or not.
+     * @param oracleAddress the address of the oracle contract
      * @return boolean representing if the rule is active
      */
-    function isOracleActive() external view returns (bool) {
-        return oracleRuleActive;
+    function isOracleActive(address oracleAddress) external view returns (bool) {
```
oscarsernarosero: Maybe ask for the ruleId

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -441,18 +468,30 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
+        for (uint256 oracleRuleIndex = 0; oracleRuleIndex < nextOracleRuleIndex; oracleRuleIndex++) {
+            if (oracleRuleIds[oracleRuleIndex].oracleAddress == oracleAddress) {
+                return oracleRuleIds[oracleRuleIndex].oracleRuleActive;
+            }
+        }
+        return false;
     }
```
oscarsernarosero, line 466: Since we are now dealing with an array, the question is how are you managing that array? how do you remove rules from it? There is already a pattern regarding arrays in Solidity since it is not a straight forward matter. You can check the repo for examples. For instance, the `_removeAddressWithMapping` function in the AppManager. Although, in this case it would be even simpler beccause I don't think we need to keep track of indexes, but maybe we could. I'll leave that decision to you.

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -618,12 +664,12 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, RuleAdministra
...
+    /**
+     * @dev function to check if Minumum Hold Time is active
+     * @return bool
      */
     function isMinimumHoldTimeActive() external view returns (bool) {
-        return minimumHoldTimeRuleActive; 
+        return minimumHoldTimeRuleActive;
     }
 
     /**
```
oscarsernarosero, line 639: same as above

oscarsernarosero (changes requested):
This is a great first ticket Michael. I left some comments.

docs/src/src/client/token/ERC20/ProtocolERC20Handler.sol/contract.ProtocolERC20Handler.md
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
ShaneDuncan602: This is actually funny. I am doing something very similar in the action rule ticket. Change is coming!

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Ironically I originally had this as the index into the array, I changed it because I thought from a usability perspective it would make more sense to pass in the address of the oracle the rule corresponds to instead of having to track the order in which you added the rules so you'll know the index to pass in. 
Also when I add the ability to remove rules from the array like you suggested below, tracking the index will get more complicated, vice just checking against the address of the oracle the rule corresponds to.
What are your thoughts?

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
ShaneDuncan602: Agreed. You only need to maintain a list of refId's

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -420,18 +431,34 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, AppAdministrato
...
-    function activateOracleRule(bool _on) external ruleAdministratorOnly(appManagerAddress) {
-        oracleRuleActive = _on;
+    function activateOracleRule(bool _on, address oracleAddress) external ruleAdministratorOnly(appManagerAddress) {
+        for (uint256 oracleRuleIndex = 0; oracleRuleIndex < nextOracleRuleIndex; oracleRuleIndex++) {
+            if (oracleRuleIds[oracleRuleIndex].oracleAddress == oracleAddress) {
+                oracleRuleIds[oracleRuleIndex].oracleRuleActive = _on;
+            }
+        }
+
         if (_on) {
```
ShaneDuncan602: I think you would only emit the event if the refId exists. Probably need to validate here as well. For instance, if they have 2 oracle rules, but make a mistake on the refId for this deactivation, they would get a completed transaction and an event but nothing would have really changed.

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
ShaneDuncan602: Yep, we need to remove every reference to oracle address in here.

src/client/token/ERC20/ProtocolERC20Handler.sol
ShaneDuncan602: In this situation, I don't think it warrants a mapping. It has a max of 10 so it could just loop through until it found it and then deactivated/removed it.

src/client/token/ERC20/ProtocolERC20Handler.sol
ShaneDuncan602: I could see where using the address would be easier for the devs but we have to keep this decoupled from the rule itself. It's up to the devs to know what refId is for each oracle. DOOM will also help them with this.

oscarsernarosero: Usually, with Solidity arrays, you rely on a mapping to keep track of the indexes. That would solve this issue since you can pass the ruleId instead of the index or oracleAddress, and that would be the best solution from the user and technical perspective imo since we keep parity with the other rules. Checkout implementations of ERC721Enumerable when they deal with removing or adding tokenIds, or you can also checkout `_removeAddressWithMapping` (it is the same concept but applied to a specific case). But basically, you have a mapping with indexes  `mapping (oracleRuleId => index) oracleRuleIdToIndex` and you keep it up to date every time you update the array.

TJ-Everett (commented):
I will re-review once you finish with the feedback

src/client/token/ERC20/ProtocolERC20Handler.sol
oscarsernarosero: sorry. scratch all this. I was thinking this was for another function

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -134,7 +139,13 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      */
     function _checkNonTaggedRules(address _from, address _to, uint256 _amount) internal {
         if (minTransferRuleActive) ruleProcessor.checkMinTransferPasses(minTransferRuleId, _amount);
-        if (oracleRuleActive) ruleProcessor.checkOraclePasses(oracleRuleId, _to);
+
+        for (uint256 oracleRuleIndex = 0; oracleRuleIndex < oracleRules.length; ) {
```
oscarsernarosero: This looks much better man. This is just a comment. There is no real need to go and change this, but I forgot to tell you in the past PR that Solidity's default values are 0, address(0), false, etc. So it is not necessary to initialize to a default value.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -432,39 +443,97 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
      * @param _ruleId Rule Id to set
      */
     function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        if (oracleRules.length >= 10) {
```
oscarsernarosero: We still have this magic number. You should create a constant to set this value. You can check GeneralTags under src/client/application/data for a similar piece.

src/client/token/ERC20/ProtocolERC20Handler.sol
```diff
@@ -432,39 +443,97 @@ contract ProtocolERC20Handler is Ownable, ProtocolHandlerCommon, ProtocolHandler
...
+                    index = oracleRuleIndex; 
+                    ruleToSwap = oracleRules[index];
+                    break;
+                }
+                unchecked {
+                    ++oracleRuleIndex;
+                }
+            }
+            oracleRules[index] = lastId;
+            oracleRules[oracleRules.length - 1] = ruleToSwap;
```
oscarsernarosero: you can skip this last step. It doesn't matter what we have here since we are gonna pop it.

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -62,6 +67,7 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
     mapping(uint256 => uint256) ownershipStart;
     /// Max Hold time hours
     uint16 constant MAX_HOLD_TIME_HOURS = 43830;
+    uint16 constant MAX_ORACLE_RULES = 10;
```
oscarsernarosero, line 69: this is why Im proposing to move these rules to a contract we can inherit. You did this here, but not in the other one.

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -282,41 +295,100 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
      * @param _ruleId Rule Id to set
      */
     function setOracleRuleId(uint32 _ruleId) external ruleAdministratorOnly(appManagerAddress) {
+        if (oracleRules.length >= 10) {
```
oscarsernarosero: still not using it tho

test/client/token/ERC20/integration/ApplicationERC20Handler.t.sol
```diff
@@ -310,31 +310,53 @@ contract ApplicationERC20HandlerTest is TestCommonFoundry {
...
+        assertEq(ruleCheck.oracleAddress, address(oracleDenied));
+
+        NonTaggedRules.OracleRule memory ruleCheckTwo = ERC20RuleProcessorFacet(address(ruleProcessor)).getOracleRule(_indexTwo);
+        assertEq(ruleCheckTwo.oracleType, 1);
+        assertEq(ruleCheckTwo.oracleAddress, address(oracleAllowed));
+
         switchToAppAdministrator();
         // add an allowed address
         goodBoys.push(address(59));
+        goodBoys.push(address(69));
```
oscarsernarosero: lol I love it, but Tayler already got mad at me about this before.

oscarsernarosero (changes requested):
Almost there.

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Since I'm already changing things in these files I went ahead and fixed this.

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC721/ProtocolERC721Handler.sol
mpetersoCode55: Resolved

mpetersoCode55: Agreed we should definitely move the common logic into a contract we inherit from, there's a lot of duplication

src/client/token/ERC721/ProtocolERC721Handler.sol
mpetersoCode55: Resolved

src/client/token/ERC20/ProtocolERC20Handler.sol
mpetersoCode55: Resolved.

test/client/token/ERC20/integration/ApplicationERC20Handler.t.sol
mpetersoCode55: No fun! I updated it to be 68

oscarsernarosero (approved):
Looks great Michael

PR #235, opened 2024-01-10
Se 484 enforce a single pricing contract for an application

TJ-Everett:
Moved pricing functions to Application Handler 
This required moving transactionLimitByRiskScore to ProtocolApplicationHandler.sol 
Updated tests and interfaces as necessary


src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -56,54 +69,82 @@ contract ProtocolApplicationHandler is Ownable, RuleAdministratorOnly, IApplicat
...
 
     /**
-     * @dev checks if any of the balance prerequisite rules are active
-     * @return true if one or more rules are active
+     * @dev checks if any of the Application level rules are active
+     * @return true if one or more rules are active  
      */
-    function requireValuations() public view returns (bool) {
-        return accountBalanceByRiskRuleActive || accountBalanceByAccessLevelRuleActive || maxTxSizePerPeriodByRiskActive || withdrawalLimitByAccessLevelRuleActive;
+    function requireApplicationRulesChecked() public view returns (bool) {
```
oscarsernarosero, line 80: do we still need this to be public? I think we can make this private or internal.

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -56,54 +69,82 @@ contract ProtocolApplicationHandler is Ownable, RuleAdministratorOnly, IApplicat
...
-        if (requireValuations() || AccessLevel0RuleActive) {
-            _checkRiskRules(_from, _to, _usdBalanceTo, _usdAmountTransferring);
-            _checkAccessLevelRules(_from, _to, _usdBalanceTo, _usdAmountTransferring);
+        /// Based on the Handler Type retrieve pricing valuations 
+        if (_handlerType == HandlerTypes.ERC20HANDLER) {
+            balanceValuation = uint128(getAccTotalValuation(_to, 0));
+            price = uint128(_getERC20Price(_tokenAddress));
+            transferValuation = uint128((price * _amount) / (10 ** IToken(_tokenAddress).decimals()));
+        } 
+        if (_handlerType == HandlerTypes.ERC721HANDLER) {
```
oscarsernarosero: I would just make this and `else if` to save a tiny bit of gas

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -49,7 +47,7 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
     uint32 private minimumHoldTimeHours;
 
     /// NFT Collection Valuation Limit
-    uint256 private nftValuationLimit = 100;
+    uint16 private nftValuationLimit = 100;
```
oscarsernarosero, line 50: Should we make this a constant? now that you brought it under our radar

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -109,15 +107,9 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
...
-            /// pure transfer rules
-            uint128 balanceValuation;
-            uint128 transferValuation;
-            if (appManager.requireValuations()) {
-                balanceValuation = uint128(getAccTotalValuation(_to, nftValuationLimit));
-                transferValuation = uint128(nftPricer.getNFTPrice(msg.sender, _tokenId));
-            }
-            appManager.checkApplicationRules( _from, _to, balanceValuation, transferValuation, action);
-            _checkTaggedAndTradingRules(_balanceFrom, _balanceTo, _from, _to, _amount, _tokenId, action);
+            if (_amount > 1) revert BatchMintBurnNotSupported(); // Batch mint and burn not supported in this release
```
oscarsernarosero: so you want to keep it here in case we upgrade later to support batch transactions?

src/client/token/ERC721/ProtocolERC721Handler.sol
```diff
@@ -170,20 +162,14 @@ contract ProtocolERC721Handler is Ownable, ProtocolHandlerCommon, ProtocolHandle
      * @param _from address of the from account
      * @param _to address of the to account
      * @param _amount number of tokens transferred
-     * @param tokenId Id of the NFT being transferred
      * @param action if selling or buying (of ActionTypes type)
      */
-    function _checkTaggedAndTradingRules(uint256 _balanceFrom, uint256 _balanceTo, address _from, address _to,uint256 _amount, uint256 tokenId, ActionTypes action) internal {
+    function _checkTaggedAndTradingRules(uint256 _balanceFrom, uint256 _balanceTo, address _from, address _to,uint256 _amount, ActionTypes action) internal {
```
oscarsernarosero, line 166: wow. Changes in this ticket cleaned up a lot of stuff. That means we are doing the right moves.

src/client/token/ProtocolHandlerCommon.sol
```diff
@@ -69,116 +62,6 @@ abstract contract ProtocolHandlerCommon is
...
-            totalValueInThisContract = _tokenAmount * uint256(nftPricer.getNFTCollectionPrice(_tokenAddress));
-        } else {
-            revert PricingModuleNotConfigured(erc20PricingAddress, nftPricingAddress);
-        }
-    }
-    
-
     /**
      * @dev gets the version of the contract
      * @return VERSION
```
oscarsernarosero, line 68: if we don't use the IToken here, we should get rid of it then. Im actually thinking we should have it in its own file in the common directory if we use it in more than one place.

oscarsernarosero (changes requested):
This is an amazing job Tayler. I have some very minor suggestions. You can tell that these changes are cleaning a lot the code transversally which means we are doing the right moves. I think making the risk rule an app level rule was a major win.

src/client/application/ProtocolApplicationHandler.sol
TJ-Everett: I thought about an if else but I went this way for other handler types and they come in (AMM, 1155, etc)

src/client/token/ERC721/ProtocolERC721Handler.sol
TJ-Everett: This was asked by Erik to be changeable for the game devs.

src/client/token/ERC721/ProtocolERC721Handler.sol
TJ-Everett: SORRY! Merging your branch in was a pain and I knew I was going to miss something.

src/client/application/ProtocolApplicationHandler.sol
TJ-Everett: I am still using this as a check in side the appManager check - this allows a small gas savings since we are always checking to the appManager checkApplicationRules() - that will check this to ensure the next steps are necessary.
So it serves a dual purpose.

src/client/token/ProtocolHandlerCommon.sol
TJ-Everett: It is still used in the token handlers for the getTotalSupply() but dropping it into a shared interface is not a bad idea either. We can let Shane and Michael weigh in.

mpetersoCode55: "but dropping it into a shared interface is not a bad idea either" I'm a fan of this route

mpetersoCode55 (approved):

src/client/token/ProtocolHandlerCommon.sol
TJ-Everett: It has been done.

src/client/application/ProtocolApplicationHandler.sol
ShaneDuncan602: It would still function the same. I think Oscar is saying to use `else if`, not just `else`. When we add the other handler types, they would be separate `else if` statements.

ShaneDuncan602 (approved):
Great job, T

PR #247, opened 2024-01-19
Se 540 combined the transaction size limits by risk rules

oscarsernarosero:


ShaneDuncan602 (approved):
This looks fantastic. Great job.

mpetersoCode55 (commented):
Looks great! One question though: 
1. It looks like we're accepting a single period value and applying it across all of the supplied max sizes and risk scores. In my rule combination (MinMaxBalance and MinBalanceByDate) I allowed for an array of periods. If nothing else I think we should apply the same design to both. If we think a single period is better I can change mine, butI think they should match.

PR #248, opened 2024-01-19
SE-483-Make-tags-optional

ShaneDuncan602:


ShaneDuncan602:
Here is the gas usage report for two approaches. 

The first was to create a new bytes array with a blank and change the parameter array to memory.
The second was to create a separate internal array with the processing logic. 
New bytes array wins. 

current before changes
| checkMinMaxAccountBalancePasses                                                            | 3212            | 4838  | 3349   | 9443  | 4       |
create new bytes array with blank tag
| checkMinMaxAccountBalancePasses                                                            | 3212            | 4967  | 3439   | 9781  | 4       |
extra internal function
| checkMinMaxAccountBalancePasses                                                            | 3212            | 5013  | 3520   | 9803  | 4       |

src/protocol/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -64,31 +65,44 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
...
+    function maxAccountBalanceCheck(uint256 balanceTo, bytes32[] memory toTags, uint256 amount, uint32 ruleId) public view {
         /// This Function checks the max account balance for accounts depending on GeneralTags.
         /// Function will revert if a transaction breaks a single tag-dependent rule
         toTags.checkMaxTags();
+        /// If the rule applies to all users, check blank only. Otherwise loop through tags and check for specific application
+        /// This was done in a minimal way to allow for modifications later while not duplicating rule check logic.
+        if(getMinMaxBalanceRule(ruleId, BLANK_TAG).maximum > 0){
+            toTags = new bytes32[](1);
+            toTags[0] = BLANK_TAG;
+        }
```
oscarsernarosero, line 77: nice

src/protocol/economic/ruleProcessor/RuleProcessorCommonLib.sol
```diff
@@ -69,10 +70,19 @@ library RuleProcessorCommonLib {
...
-        if (_tags.length > MAX_TAGS) revert MaxTagLimitReached();
+        if (_tags.length > MAX_TAGS) revert MaxTagLimitReached();    
+    }
+
+    /**
+     * @dev determine if the rule applies to all users
+     * @param _tags the timestamp the rule was enabled
+     * @param _isAll true if applies to all users
+     */
+    function isAllUsers(bytes32[] memory _tags) internal pure returns(bool _isAll){
```
oscarsernarosero: Could we name this a bit more specifically? like `isApplicableToAllUsers`? the way it is now sounds like it is determining if all tags are `users`.

src/protocol/economic/ruleProcessor/TaggedRuleDataFacet.sol
```diff
@@ -137,6 +139,7 @@ contract TaggedRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents,
         if (_accountTypes.length != _minimum.length || _accountTypes.length != _maximum.length) revert InputArraysMustHaveSameLength();
         // since all the arrays must have matching lengths, it is only necessary to check for one of them being empty.
         if (_accountTypes.length == 0) revert InvalidRuleInput();
+        _accountTypes.areTagsValid();
```
oscarsernarosero, line 142: I really like this, but it looks like we are looping twice through the account array, and it is basically doing the same as the removed line ` if (_accountTypes[i] == bytes32("")) revert BlankTag();` but just throwing a different error. I would suggest maybe creating a boolean variable that checks if the _accountTypes.length is 1 in which case blank tags are allowed, and add it in an AND conditional:
```
bool isLengthGreaterThan1 = _accountTypes.length > 1;
..... loop header and stuff....
if (isLengthGreaterThan1 && _accountTypes[i] == bytes32("")) revert TagListMustBeSingleBlankOrValueList();
```

This would be replacing the original line (which was removed):
` if (_accountTypes[i] == bytes32("")) revert BlankTag();`

So we wouldn't need the `_accountTypes.areTagsValid();` line.

test/client/token/ERC20/integration/ApplicationERC20.t.sol
```diff
@@ -125,6 +125,54 @@ contract ApplicationERC20Test is TestCommonFoundry, DummyAMM {
...
+        uint256[] memory min = createUint256Array(10);
+        uint256[] memory max = createUint256Array(1000);
+        // add the actual rule
+        switchToRuleAdmin();
+        uint32 ruleId = TaggedRuleDataFacet(address(ruleProcessor)).addMinMaxBalanceRule(address(applicationAppManager), accs, min, max);
+        ///update ruleId in coin rule handler
+        // create the default actions array
+        ActionTypes[] memory actionTypes = new ActionTypes[](2);
+        actionTypes[0] = ActionTypes.P2P_TRANSFER;
+        actionTypes[1] = ActionTypes.SELL;
```
oscarsernarosero, line 145: looks like a good candidate for the createArrayFunctions. I added a "bonus" section in the rule creation ticket to do this.

oscarsernarosero (approved):
This looks really good Shane. I have just a couple minor suggestions, So I'll go ahead and approve and I'll leave it to you to decide on the suggestions.

src/protocol/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
```diff
@@ -64,31 +65,44 @@ contract ERC20TaggedRuleProcessorFacet is IRuleProcessorErrors, IInputErrors, IT
      * @param amount Number of tokens to be transferred
      * @param ruleId Rule identifier for rule arguments
      */
-    function maxAccountBalanceCheck(uint256 balanceTo, bytes32[] calldata toTags, uint256 amount, uint32 ruleId) public view {
+    function maxAccountBalanceCheck(uint256 balanceTo, bytes32[] memory toTags, uint256 amount, uint32 ruleId) public view {
         /// This Function checks the max account balance for accounts depending on GeneralTags.
         /// Function will revert if a transaction breaks a single tag-dependent rule
         toTags.checkMaxTags();
+        /// If the rule applies to all users, check blank only. Otherwise loop through tags and check for specific application
```
mpetersoCode55, line 72: If I'm following this correctly, we're saying if the blank tag has been applied, only check the blank tag. I thought we decided you could have blank tags and more restrictive tags applied to the same rule. Couldn't there be a blank tag and a more restrictive tag that are both applicable? If so wouldn't we want to check the more restrictive tag if the blank tag passes?

src/protocol/economic/ruleProcessor/RuleProcessorCommonLib.sol
```diff
@@ -91,4 +101,22 @@ library RuleProcessorCommonLib {
         }
         return maxSize; 
     }
+
+     /**
+     * @dev validate tags to ensure only a blank or valid tags were submitted.
+     * @param _accountTags the timestamp the rule was enabled
+     * @return _valid returns true if tag entry is valid
+     */
+    function areTagsValid(bytes32[] calldata _accountTags) internal pure returns (bool) {
```
mpetersoCode55, line 110: are we only allowing either a blank tag or more restrictive tags? I thought we were going to allow both?

src/protocol/economic/ruleProcessor/RuleProcessorCommonLib.sol
ShaneDuncan602: Great idea

src/protocol/economic/ruleProcessor/TaggedRuleDataFacet.sol
ShaneDuncan602: I considered doing this, but since the rule creation is a dev transaction I prioritized modularization above gas usage. I am a firm believer in making validation modular, especially in this case. We know we will be modifying this in the future to allow both a blank and specific tags so I wanted to make sure the validation was as shared as possible to change it easily.

test/client/token/ERC20/integration/ApplicationERC20.t.sol
ShaneDuncan602: There are many specific usages and I wasn't sure if we should create separate functions for the special cases or not. I'm fine either way.

src/protocol/economic/ruleProcessor/ERC20TaggedRuleProcessorFacet.sol
ShaneDuncan602: Yep, sorry. It was decided to make the rules allow either a blank or a set of specific tags, but not both.

src/protocol/economic/ruleProcessor/RuleProcessorCommonLib.sol
ShaneDuncan602: Yep, sorry. It was decided to make the rules allow either a blank or a set of specific tags, but not both.

TJ-Everett (approved):
Awesome work Shane. Approved for me.

PR #250, opened 2024-01-23
Se 449 create GitHub workflow for necessist run

bfcoursewool:
Everything for automating necessist. Includes a dockerfile for the necessist build, github workflow for building and pushing to ECR as well as redeploying the k8s job from the new container, and a k8s manifest file describing the new necessist job. 

Run the workflow from here in github under Actions -> Run Necessist, and when it finally finishes (the GH workflow itself only takes 5 minutes) there will be a new necessist-YY-mm-dd.db file in the "necessist database" bucket of S3 in our AWS account. Voila!

### DIFF
--- a/Dockerfile.anvil
+++ b/Dockerfile.anvil
@@ -13,4 +13,4 @@ ENV CHAIN_ID 31337
 
 RUN pip install -r requirements.txt
 
-ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID
+ENTRYPOINT anvil --host 0.0.0.0 --code-size-limit 0x60000 --chain-id $CHAIN_ID
--- a/Dockerfile.anvil
+++ b/Dockerfile.anvil
@@ -13,4 +13,4 @@ ENV CHAIN_ID 31337
 
 RUN pip install -r requirements.txt
 
-ENTRYPOINT anvil --host 0.0.0.0 --code-size-limit 0x60000 --chain-id $CHAIN_ID
+ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID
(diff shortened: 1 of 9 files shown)


bcamp136 (approved):

oscarsernarosero (commented):
Looks good to me Rob. I didn't really mess at all with Necessist, so Shane will be the real test.

foundry.toml
```diff
@@ -5,12 +5,15 @@ libs = ['lib']
 
 [profile.local]
 eth-rpc-url = 'http://localhost:8545'
+code-size-limit = 0x60000
```
ShaneDuncan602: I don't think we need the 0x for this value. We should be fine with actually using the kb value of 45000 or so.

Dockerfile.anvil
```diff
@@ -13,4 +13,4 @@ ENV CHAIN_ID 31337
 
 RUN pip install -r requirements.txt
 
-ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID
\ No newline at end of file
+ENTRYPOINT anvil --host 0.0.0.0 --code-size-limit 0x60000 --chain-id $CHAIN_ID
```
ShaneDuncan602: We could probably leave out the 0x value and just use the actual Kb

necessist-run.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+source ~/.bashrc
+source ~/.cargo/env
+source .venv/bin/activate
+cargo install necessist
+necessist --verbose --framework foundry -- --ffi
```
ShaneDuncan602, line 8: ToB had to give us a fix for a problem that we encountered. We have to use

`cargo install --git https://github.com/trailofbits/necessist --branch fix-foundry-build`

ShaneDuncan602 (changes requested):
Just a couple of questions

ShaneDuncan602 (approved):
Great job

PR #251, opened 2024-01-23
Se 503 document tagging configurations

TJ-Everett:
Changed GeneralTags -> Tags and generalTag -> tag 
This was confirmed with product team to streamline understanding of what tags are/do.


docs/userGuides/tags/PROTOCOL-TAGS-STRUCTURE.md
```diff
@@ -0,0 +1,149 @@
...
+## Scope 
+
+### Account Tags: 
+- Tags can be applied to individual accounts and used to assess fees or facilitate rule checks throughout the procotol. When an account (user) is tagged, they will be subject to any rules that are active that utilize that tag. 
+
+###### *see [TAGGED-RULES](./TAGGED-RULES.md)* 
+
+### Address Tags: 
+- Tags can be applied to addresses of contracts. These address tags are used in a similar way to account tags as they are used to assess fees or facilitate rule checks for specific addresses.
+
```
oscarsernarosero: Im not sure we should make this distinction in whole sections. It is a good idea to mention that tags can be applied to addresses which could be EOAs or smart contracts (and their different rules), but I don't think they should be in separate sections. The protocol doesn't really distinguish between these 2 categories when applying tags.

docs/userGuides/tags/README.md
```diff
@@ -0,0 +1,18 @@
+# Tags Guide
+[![Project Version][version-image]][version-url]
+--- 
+
+## Tag Information Documentation: 
+
+The protocol uses tags to assess fees and facilitate rule checks for account and addresses. Tags are applied to accounts or addresses via an addTag function in the app manager by application administators. Tags are stored as bytes32 in a mapping inside the tags data contract. A maximum of 10 tags may be applied to each address or account.
```
oscarsernarosero, line 7: `The protocol uses tags to assess fees and perform rule checks for tag-based rules`?

oscarsernarosero: very nice line in general tho

oscarsernarosero (approved):
I love it. Just some minor suggestions.

mpetersoCode55 (approved):

PR #252, opened 2024-01-24
Se 505 document risk level configurations

TJ-Everett:
Also updated risk rule params in rule storage diamond to match risk score (instead of level) verbiage 
Added a removeRiskScore() to appManager as this was missing.


oscarsernarosero (changes requested):
Looks great. I still don't feel comfortable with "addresses and accounts". In my opinion, this is confusing and is not adding any information. If we just say "addresses" we will be providing the same amount of information with less words. Also, I didn't find a place where it says that the default risk score is 0. Other than that, this looks really good.

mpetersoCode55 (approved):
Looks great!

PR #258, opened 2024-01-31
SE-471. Create Deployment Tests and Github Actions.

mpetersoCode55:
1. Added scripts to parse output and update env file after deploying the protocol and application.
2. Added verification checks into the Application deployment script.
3. Added tests to verify an application deployment against an actual chain and test setup
4. Fixed issues with Protocol Deployment tests when run against actual chain
5. Setup a Github worflow to run all of the above and: a. on success kick off the k8s workflow b. on failure send an email notification to the team.


Dockerfile.anvil
```diff
@@ -13,4 +13,4 @@ ENV CHAIN_ID 31337
 
 RUN pip install -r requirements.txt
 
-ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID
\ No newline at end of file
+ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID --code-size-limit 30000
```
bfcoursewool, line 16: So I just checked and in the necessist branch I set code-size-limit to 45000, and I put those values in foundry.toml... I'm not hugely attached to any particular approach to this setting, either where we set it, or what value we set it to, as long as it works, but I guess I do think we should just do it in one or the other place.

Dockerfile.deploymentCheck
```diff
@@ -0,0 +1,29 @@
...
+ARG AWS_REGION
+ARG BRANCH_NAME
+
+ENV AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+ENV AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+ENV AWS_DEFAULT_REGION=${AWS_REGION}
+ENV GITHUB_BRANCH_NAME=${BRANCH_NAME}
+
+CMD ["./install-deploy-check.sh"]
+
```
bfcoursewool, line 28: Dang, this dockerfile looks like it was made by a real DevOps guy...

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      RUST_LOG: backend,api,node,rpc=warn
+    restart: unless-stopped
+    ports:
+      - "8545:8545"
+  tron-setup:
+    depends_on:
+      - anvil
+    build:
+      context: .
+      dockerfile: Dockerfile.deploymentCheck
```
bfcoursewool, line 20: This probably doesn't make much difference since I think nobody actually uses this docker-compose file for anything in the tron dev process, but maybe it's worth pointing out in like a comment or something that this is using the "deploy check" tron container, rather than the "real" one that actually just deploys tron. I think in theory if someone did build this docker-compose stack on their machine, then it would actually go run the tests and potentially even actually kick off the k8s workflow in GH and stuff, which may not be ideal.

script/ParseApplicationDeploy.sh
```diff
@@ -0,0 +1,194 @@
...
+echo APPLICATION_ERC20_HANDLER_ADDRESS_2=$APPLICATION_ERC20_HANDLER_ADDRESS_2
+echo
+
+os=$(uname -a)
+if [[ $os == *"Darwin"* ]]; then
+  sed -i '' 's/TEST_DEPLOY_APPLICATION_ERC20_HANDLER_ADDRESS_2=.*/TEST_DEPLOY_APPLICATION_ERC20_HANDLER_ADDRESS_2='$APPLICATION_ERC20_HANDLER_ADDRESS_2'/g' $ENV_FILE
+else
+  sed -i 's/TEST_DEPLOY_APPLICATION_ERC20_HANDLER_ADDRESS_2=.*/TEST_DEPLOY_APPLICATION_ERC20_HANDLER_ADDRESS_2='$APPLICATION_ERC20_HANDLER_ADDRESS_2'/g' $ENV_FILE 
+fi
+
```
bfcoursewool, line 194: The bash-fu is strong with this one.

bfcoursewool (approved):
Looks great to me!

Dockerfile.anvil
mpetersoCode55: Good catch, yeah I prefer the foundry.toml approach, seems cleaner to keep it all in one location. I'll update accordingly

Dockerfile.deploymentCheck
mpetersoCode55: Don't you put that evil on me

docker-compose.yml
mpetersoCode55: Yeah I agree, the better we document the flow and purpose the less pain future us will have to endure. Not no pain, just less

Dockerfile.anvil
bfcoursewool: @mpetersoCode55 -- ah, so actually we might need to go with the command line parameter approach... I've been messing with tron over here today for the rules rename stuff and discovered that something isn't working with the code-size-limit value that's getting set in foundry.toml... I suppose we could continue to experiment and figure out why it isn't working there and how to make it work there, but I'm lazy and it definitely works the way you have it here in this branch so I'm inclined to just go with that for right now...

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+
+        // Checking to make sure the AppManager has a handler
+        if(address(applicationHandler) == address(0)) {
+            revert("Application Handler not properly deployed");
+        }
+
+        // Checking to make sure the handler is connected back to the AppManager
+        if(applicationHandler.appManagerAddress() != address(applicationAppManager)) {
+            revert("Application Handler not correctly connected to the App Manager");
+        }
```
oscarsernarosero, line 136: Nice. We should also check for the appManager having connected its handler.

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+        verifyAppManager();
+        verifyERC20(true);
+        verifyERC20(false);
+        verifyERC721();
+        vm.stopBroadcast();
+        // --------------------------------------------------------------------------------------
+    }
+
+    function verifyAppManager() public view {
+        // Checking to make sure AppManager is deployed
```
oscarsernarosero: I would just change these comments to something more accurate like `// Checking XXX-contract environment variable has been set` since that is what we are actually checking here. We will know for sure if the contract has been deployed when we call functions on that address and we get our expected results like you do in the next steps.

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+
+       // Make sure the ERC721 Handler Contract has been deployed
+        if(address(nft1) == address(0)) {
+            revert("ERC721 Handler not properly deployed");
+        }
+
+       // Make sure the ERC20 Pricing Module Contract has been deployed
+        if(address(openOcean) == address(0)) {
+            revert("ERC721 Pricing Module not properly deployed");
+        }
```
oscarsernarosero: Maybe call a public function in this address and test that it returns what it should to make sure the contract lives in that address.

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+            revert("ERC721 not properly registered with the AppManager");
+        }
+
+        // Checking to make sure the ERC721's Handler is registered with the AppManager
+        if(!applicationAppManager.isRegisteredHandler(address(applicationNFTHandler))) {
+            revert("ERC721 Handler not properly registered with the AppManager");
+        }
+    }
+
+    function verifyERC20(bool isCoin1) public view {
```
oscarsernarosero, line 183: nice. I like the flag.

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+
+       // Make sure the ERC20 Handler Contract has been deployed
+        if(address(isCoin1 ? applicationCoinHandler : applicationCoinHandler2) == address(0)) {
+            revert("ERC20 Handler not properly deployed");
+        }
+
+       // Make sure the ERC20 Pricing Module Contract has been deployed
+        if(address(exchange) == address(0)) {
+            revert("ERC20 Pricing Module not properly deployed");
+        }
```
oscarsernarosero: Same here. Try calling a public function and check its returned data.

script/clientScripts/ApplicationDeployAll.s.sol
```diff
@@ -99,5 +112,128 @@ contract ApplicationDeployAllScript is Script {
...
+            }
+        } else {
+            if(coin2.getHandlerAddress() != address(applicationCoinHandler2)) {
+                revert("Frankenstein Coin not correctly connected to its handler");
+            }
+        }
+
+        // Checking to make sure the handler is connected to the ERC20
+        if((isCoin1 ? applicationCoinHandler.owner() : applicationCoinHandler2.owner()) != address(isCoin1 ? coin1 : coin2)) {
+            revert("ERC20 Handlers owner should be the Frankenstein Coin");
```
oscarsernarosero: Is this error message accurate?

test/client/application/deployment/ApplicationDeployment.t.sol
```diff
@@ -0,0 +1,143 @@
...
+contract ApplicationDeploymentTest is Test, TestCommonFoundry {
+
+    address appManagerAddress;
+    bool forkTest;
+
+    function setUp() public {
+        if(vm.envAddress("TEST_DEPLOY_APPLICATION_APP_MANAGER") != address(0x0)) {
+            // Verify App Manager has been deployed
+            superAdmin = vm.envAddress("QUORRA");
+            applicationAppManager = ApplicationAppManager(payable(vm.envAddress("TEST_DEPLOY_APPLICATION_APP_MANAGER")));
```
oscarsernarosero: why do we need the `payable` attribute for this address argument?

oscarsernarosero (changes requested):
Looks really good and thorough Michael. I just have some minor suggestions.

Dockerfile.anvil
mpetersoCode55: Ok let's go with the command line parameter approach for now then. I think we just want something that we know works for now since it's going to be removed once we finish the handler refactor anyways (so we don't have false positives moving forward. I'll remove it from foundry.toml and we can just rely on what I'm passing into anvil, in the anvil container

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: Resolved

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: Resolved

test/client/application/deployment/ApplicationDeployment.t.sol
mpetersoCode55: Good catch! Resolved

script/ParseApplicationDeploy.sh
```diff
@@ -0,0 +1,194 @@
...
+  sed -i 's/TEST_DEPLOY_APPLICATION_ERC20_ADDRESS=.*/TEST_DEPLOY_APPLICATION_ERC20_ADDRESS='$APPLICATION_ERC20_ADDRESS'/g' $ENV_FILE
+fi
+
+# Retrieve the ERC 20 Handler
+APPLICATION_ERC20_HANDLER_ADDRESS=$(parseContractAddress "ApplicationERC20Handler" $CHAIN_ID 0)
+echo APPLICATION_ERC20_HANDLER_ADDRESS=$APPLICATION_ERC20_HANDLER_ADDRESS
+echo
+
+os=$(uname -a)
+if [[ $os == *"Darwin"* ]]; then
```
ShaneDuncan602, line 79: It's sad we have to do this.

script/clientScripts/ApplicationDeployAll.s.sol
ShaneDuncan602: I agree. Verify usually refers to verifying the contract for block explorers.

script/clientScripts/ApplicationDeployAll.s.sol
ShaneDuncan602, line 1: Is DOOM the only place we use this script? Since we are already maintaining the Application_Deploy_01 through 08 maybe we should abandon this altogether and use those instead. This might be too much scope creep for this ticket but those other scripts didn't exist at the time of ticket creation so I think we should consider it as long as the ApplicationDeployAll and all the mini scripts do the same thing.

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: Resolved

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: Resolved

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: Resolved

docker-compose.yml
mpetersoCode55: Resolved.

script/clientScripts/ApplicationDeployAll.s.sol
mpetersoCode55: (Resolution pulled from Slack conversation)
I agree that retiring the DeployAll script is a good call. I think we should have a dedicated effort though. I'd want to also remove the DeployAllModules script and update the repo to make sure every reference to either is replaced with the relevant broken up scripts. (I think we have a pretty decent mix of the two approaches throughout the repo right now so I'd vote for a dedicated ticket)

ShaneDuncan602 (approved):
Good job. Adding a new ticket to retire the ApplicationDeployAll ticket.

TJ-Everett (approved):
Awesome job on this one!

PR #261, opened 2024-02-01
Se 486 refine rule list

TJ-Everett:


mpetersoCode55 (approved):

docs/userGuides/rules/README.md
```diff
@@ -0,0 +1,40 @@
...
+| [Minimum Transaction](./TOKEN-MIN-TRANSACTION-SIZE.md)| The purpose of the token-min-tx-size rule is to prevent micro-trades or dust trades within an ecosystem. A developer can set the minimum number of tokens required per transfer to prevent these types of trades. |
+
+## How Rules Work 
+
+The protocol allows [rule administrators](../permissions/ADMIN-ROLES.md) of an application to add rules to the Rule Processor Diamond. The rule id generated by the protocol is then used within an application to apply to [economic actions](./ACTION-TYPES.md) users can take. The rules are immutable and can only ever by added to the protocol and cannot be removed or modified once created. A rule administator may add a new rule to the Rule Processor diamond at any time. The protocol will return a rule id that is used by the application to set the rule within each applicable handler contract.
```
ShaneDuncan602: probably don't need "of an application"
Can we change "application to apply to [economic actions](./ACTION-TYPES.md) users can take" to "application to apply to user [economic actions](./ACTION-TYPES.md)"?
Can we change "The rules are immutable and can only ever by added to the protocol and cannot be removed or modified once created." to "The rules are immutable and cannot be removed or modified once created."

docs/userGuides/rules/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Token level rules are set in token handlers and application level rules in the application handler. Within each rule document above is the scope of that rule and the level the rule applies too. Within the handler contract is an activate rule function that receives the rule id for each type of rule that may be activated for that handler. A token handler does not have an activate function for application level rules and vice versa. The activate rule function will set the rule id and [economic action type](./ACTION-TYPES.md) that the rule will apply too. Each rule id will only apply to that action type. If multiple action types are using the same rule id, then this step must be repeated for each action type.
```
ShaneDuncan602: 1. "Token level rules are set in token handlers and application level rules in the application handler. Within each rule document above is the scope of that rule and the level the rule applies too." to "Token level rules are applied in token handlers and application level rules in the application handler. Within each rule document above is the scope of that rule and the level in which it applies."
2. "Within the handler contract is an activate rule function that receives the rule id for each type of rule that may be activated for that handler." to "Within the handler contract is a set rule function that receives the rule id for each applicable rule for that handler."
3. "The activate rule function will set the rule id and [economic action type](./ACTION-TYPES.md) that the rule will apply too." to "The set rule function sets the rule per applicable [economic action type](./ACTION-TYPES.md)."
4. "Each rule id will only apply to that action type." to "Each rule will only apply to the action types set in the activate rule function."
5. "If multiple action types are using the same rule id, then this step must be repeated for each action type." to "Multiple action types may be used for each for each rule by sending all the desired types to the set rule function."
6. Also, I think we need to discuss the activate and deactivate functionality.

docs/userGuides/rules/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Rule application is a two step processes. First the rule admin must obtain a rule id, either from creating a rule or using an already created rule. Then apply the rule id to the applicable handler through the appropriate activate rule function. Once the rule is added to the protocol and activated in the handler, all economic actions of that type will validate the action through the protocol using the provided rule parameters of that rule id. Some, all or select rules may be active at any time depending on the application developer.
```
ShaneDuncan602: 1. "processes" to "process"
2. "First the rule admin must obtain a rule id, either from creating a rule or using an already created rule. Then apply the rule id to the applicable handler through the appropriate activate rule function." to "First, the rule admin must obtain a rule id, either from creating a rule or using an already created rule. Then, they must apply the rule id to the applicable handler through the appropriate set rule function."
3. "Once the rule is added to the protocol and activated in the handler, all economic actions of that type will validate the action through the protocol using the provided rule parameters of that rule id." We can probably remove id from the end.
4. "Some, all or select rules may be active at any time depending on the application developer." to "The application developer sets the rule applicability and may activate and deactivate as needed. One or many rules may be active at a time."

ShaneDuncan602 (changes requested):
This is great. Just have some wordsmithing.

docs/userGuides/rules/README.md
```diff
@@ -25,11 +25,13 @@
...
 
-Rule application is a two step processes. First the rule admin must obtain a rule id, either from creating a rule or using an already created rule. Then apply the rule id to the applicable handler through the appropriate activate rule function. Once the rule is added to the protocol and activated in the handler, all economic actions of that type will validate the action through the protocol using the provided rule parameters of that rule id. Some, all or select rules may be active at any time depending on the application developer. 
+Rule application is a two step processes. First the rule admin must obtain a rule id, either from creating a rule or using an already created rule. Then apply the rule id to the applicable handler through the appropriate activate rule function. Once the rule is added to the protocol and activated in the handler, all [economic actions](./ACTION-TYPES.md) of that type will validate the action through the protocol using the provided rule parameters of that rule id. Some, all or select rules may be active at any time depending on the application developer.
```
ShaneDuncan602: We need to make sure we use the "set" term instead of the "activate" because it may confuse them. I think "activate rule function. Once the rule is added to the protocol and activated in the" should be "set rule function. Once the rule is added to the protocol and set in the"

docs/userGuides/rules/README.md
```diff
@@ -29,9 +29,9 @@ The protocol allows [rule administrators](../permissions/ADMIN-ROLES.md) to add
...
 
-Rule deactivation is done through the same activate function. Deactivate the rule by sending a false boolean with the desired [economic action types](./ACTION-TYPES.md). When a rule is deactivated the [economic action type](./ACTION-TYPES.md) will not be validated through the protocol. Rule admins can check the active status of a rule at any time. Within the handler contracts are is rule active functions that return the status of the rule. The rule id for each rule and [economic action type](./ACTION-TYPES.md) may be retrieved through the get rule id functions in each handler. 
+Rule deactivation is done through the same set function. Deactivate the rule by sending a false boolean with the desired [economic action types](./ACTION-TYPES.md). When a rule is set to deactive the [economic action type](./ACTION-TYPES.md) will not be validated through the protocol. Rule admins can check the active status of a rule at any time. Within the handler contracts are is rule active functions that return the status of the rule. The rule id for each rule and [economic action type](./ACTION-TYPES.md) may be retrieved through the get rule id functions in each handler.
```
ShaneDuncan602: Rule deactivate is done through the activate function. 

Can we change it to this: "Rule deactivation is done by sending a "false" boolean with the desired [economic action types](./ACTION-TYPES.md) to the activate function. When rule active status is set to "false", the [economic action type](./ACTION-TYPES.md) will not be validated through the protocol. Rule admins can check the activation status of a rule at any time. Within the handler contracts are functions that return the activation status of the rule. The rule id for each rule and [economic action type](./ACTION-TYPES.md) may be retrieved through the get rule id functions in each handler.

ShaneDuncan602 (approved):
Looks great!

PR #268, opened 2024-02-09
Non interactive deploys

bfcoursewool:
Just a few small changes here and there to support the proper functioning of DOOM...


src/client/application/AppManager.sol
```diff
@@ -1009,7 +1009,6 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents {
         if (_newApplicationHandler == address(0)) revert ZeroAddress();
         applicationHandler = ProtocolApplicationHandler(_newApplicationHandler);
         applicationHandlerAddress = _newApplicationHandler;
-        emit HandlerConnected(applicationHandlerAddress, address(this));
```
ShaneDuncan602, line 1012: DOOM doesn't need to track this? This event would be the action that opens up their app to being able to apply rules at the application level. It wouldn't really be the AppManager deploy.

ShaneDuncan602 (approved):
Approved but with one question.

PR #274, opened 2024-02-14
SE-490-Ensure-naming-conventions-are-aligned-between-UI-and-Documentation

bcamp136:
Just fixed a few errors I noticed in the rule docs while looking through them to make DOOM updates.


TJ-Everett (approved):
Thank you for catching these Brandon!  Approved

PR #277, opened 2024-02-20
Se 544 create documentation for handlers

TJ-Everett:
App Handler md
Asset Handler Diamond mds


docs/userGuides/Architecture/Client/Application/APP-HANDLER.md
```diff
@@ -0,0 +1,43 @@
+# App Handler
```
ShaneDuncan602: I think we should use the full name on titles and in the descriptions. Application Handler sounds fancy.

docs/userGuides/Architecture/Client/Application/APP-HANDLER.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Purpose
+
+The App Handler supports the App Manager by storing the application level rules data and functions. Procotol supported asset handler contracts call check application level rules via the App Manager. The App Manager then checks the associated App Handler where application level rule data is stored. The App Handler contract also serves as the App Manager's connection to the protocol rule processor diamond for the application level rules.
+
+
+### Application Level Rules
+
+The App Handler facilitates the rule checks for each application level rule. The first function called by the App Manager is:
```
ShaneDuncan602: Can we define what an Application Level Rule is here?

docs/userGuides/Architecture/Client/Application/APP-HANDLER.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+### Application Level Rules
+
+The App Handler facilitates the rule checks for each application level rule. The first function called by the App Manager is: 
+
+```c
+function requireApplicationRulesChecked() public view returns (bool)
+```
+This function allows the App Manager to know if any application level rules are active and if the call should continue to the handler to check the active rules.
```
ShaneDuncan602: Since there isn't a specific function for pause rules, I think we should mention it checking them.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND-FACET-LIST.md
```diff
@@ -0,0 +1,30 @@
...
+
+Facet Location 
+
+## Rule Facets 
+
+### ERC20 Rule Facets: 
+- [ERC20HandlerMainFacet.sol](../../../../../src/client/token/handler/diamond/ERC20HandlerMainFacet.sol) 
+- [ERC20TaggedRuleFacet.sol](../../../../../src/client/token/handler/diamond/ERC20TaggedRuleFacet.sol) 
+- [ERC20NonTaggedRuleFacet.sol](../../../../../src/client/token/handler/diamond/ERC20NonTaggedRuleFacet.sol)
+- [FeesFacet.sol](../../../../../src/client/token/handler/diamond/FeesFacet.sol)
```
ShaneDuncan602, line 14: It's tough to say whether this is a rule facet or a data facet. I tend to think of it as a data facet since it stores the fees per asset. I'm good either way though.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
```diff
@@ -0,0 +1,87 @@
+# Protocol Asset Handler Diamond Structure 
+
+## Purpose
+
+The Protocol Asset Handler Diamond serves as the access point  to the protocol for a protocol supported asset. The protocol supported asset accepts the Handler Diamond proxy address and calls the check all rules function. The Handler Diamond stores all asset level rule data, rule activation status', and connects the token to the App Manager for role based access control.
```
ShaneDuncan602: double space between point and to

"The protocol supported asset accepts the Handler Diamond proxy address and calls the check all rules function." Can we change this to "The protocol supported asset stores the Handler Diamond proxy address and uses it to call the check all rules function."

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
```diff
@@ -0,0 +1,87 @@
...
+
+Asset level rules are set by [Rule administrators](../permissions/ADMIN-ROLES.md). The protocol supplied rule id for each [Rule](../../../rules/README.md) is stored with an active status boolean inside a struct. These structs are stored in a mapping with the [action type](../../../rules/ACTION-TYPES.md) as the pointer. When setting a rule to active in the handler, the status is applied per applicable [action type](../../../rules/ACTION-TYPES.md). The Handler Diamond then stores each action type and rule mapping into a struct within the [Rule Storage Facet](./PROTOCOL-ASSET-HANDLER-DIAMOND-FACET-LIST.md). 
+
+Each Protocol supported asset type (ERC20, ERC721, etc) will need a handler diamond deployed and connected to the asset. The Handler diamond architecture will remain the same for each asset type. The proxy contract, libraries and storage facets may be the same for different asset handler diamonds. Each asset type will have unique facets to support the asset type:
```
ShaneDuncan602: "will need a handler diamond " to "will need one handler diamond "

"The proxy contract, libraries and storage facets may be the same for different asset handler diamonds." I'm a little confused by this statement. We might want to say that they share the same code, but they will never share the same deployed facets.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
```diff
@@ -0,0 +1,87 @@
...
+[Protocol Fungible Handler](./PROTOCOL-FUNGIBLE-TOKEN-HANDLER.md) 
+[Protocol NonFungible Handler](./PROTOCOL-NONFUNGIBLE-TOKEN-HANDLER.md)
```
ShaneDuncan602: Are we getting away from the 20 and 721 labels?

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
```diff
@@ -0,0 +1,87 @@
...
+
+Each Protocol supported asset type (ERC20, ERC721, etc) will need a handler diamond deployed and connected to the asset. The Handler diamond architecture will remain the same for each asset type. The proxy contract, libraries and storage facets may be the same for different asset handler diamonds. Each asset type will have unique facets to support the asset type:
+[Protocol Fungible Handler](./PROTOCOL-FUNGIBLE-TOKEN-HANDLER.md) 
+[Protocol NonFungible Handler](./PROTOCOL-NONFUNGIBLE-TOKEN-HANDLER.md) 
+
+#### *[see diamond diagram](INSERT LINK AFTER REBASE)*
+
+### Diamond Pattern
+
+The diamond pattern allows the handler to upgrade, add new features and improvements through the use of a proxy contract. New facet contracts can be deployed and connected to the diamond via a specialized function called diamondCut. New facets and functions allow the handler to grow while maintaining address immutability with the proxy contract. Calling contracts will only need to set the address of the diamond proxy at deployment, without having to worry about that address changing overtime. The Handler Diamond follows ERC 2535 standards for storage and functions.
```
ShaneDuncan602: overtime should be two words

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-FUNGIBLE-TOKEN-HANDLER.md
```diff
@@ -0,0 +1,104 @@
...
+
+- **event DiamondCut(_diamondCut, init, data)**: 
+    - Emitted when: the Asset Handler Diamond has been upgraded.
+    - Parameters:
+        - _diamondCut: Facets Array
+        - init: Address of the contract or facet to execute "data"
+        - data: A function call, including function selector and arguments calldata is executed with delegatecall on "init"
+
+## Upgrading
+- The new facet that is to be added to the diamond should first be deployed to the network the Asset Handler Diamond is deployed to.
```
ShaneDuncan602, line 64: "to the network the Asset Handler Diamond" should be "to the same network the Asset Handler Diamond"

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-NONFUNGIBLE-TOKEN-HANDLER.md
```diff
@@ -0,0 +1,88 @@
...
+
+- **event DiamondCut(_diamondCut, init, data)**: 
+    - Emitted when: the Asset Handler Diamond has been upgraded.
+    - Parameters:
+        - _diamondCut: Facets Array
+        - init: Address of the contract or facet to execute "data"
+        - data: A function call, including function selector and arguments calldata is executed with delegatecall on "init"
+
+## Upgrading
+- The new facet that is to be added to the diamond should first be deployed to the network the Asset Handler Diamond is deployed to.
```
ShaneDuncan602: "to the network the Asset Handler Diamond" to "to the same network the Asset Handler Diamond"

ShaneDuncan602 (changes requested):
Solid work here, Tayler. I have a few minor suggestions but this is really good.

docs/userGuides/Architecture/Client/Application/APP-HANDLER.md
oscarsernarosero: I second that

docs/userGuides/Architecture/Client/Application/APP-HANDLER.md
```diff
@@ -0,0 +1,43 @@
+# App Handler
+
+## Purpose
+
+The App Handler supports the App Manager by storing the application level rules data and functions. Procotol supported asset handler contracts call check application level rules via the App Manager. The App Manager then checks the associated App Handler where application level rule data is stored. The App Handler contract also serves as the App Manager's connection to the protocol rule processor diamond for the application level rules.
```
oscarsernarosero: `check application level rules` this can sound very confusing to the person not familiar with the code. So I think we have a couple options here:
`check-application-level-rules function` or `checkApplicationLevelRules`. If the latter is preferred, I would even do the code single quote so it is clear that it is a function

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND-FACET-LIST.md
oscarsernarosero: Yeah. I think it is merged since in other rules we have the data in a contract and the activation flag in the handler, but not in this case. They all live in the same place, so I would say, for simplicity, it is a good idea to treat it as a rule in general.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
oscarsernarosero: same as above about `check all rules`

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
```diff
@@ -0,0 +1,87 @@
+# Protocol Asset Handler Diamond Structure 
+
+## Purpose
+
+The Protocol Asset Handler Diamond serves as the access point  to the protocol for a protocol supported asset. The protocol supported asset accepts the Handler Diamond proxy address and calls the check all rules function. The Handler Diamond stores all asset level rule data, rule activation status', and connects the token to the App Manager for role based access control. 
+
+Asset level rules are set by [Rule administrators](../permissions/ADMIN-ROLES.md). The protocol supplied rule id for each [Rule](../../../rules/README.md) is stored with an active status boolean inside a struct. These structs are stored in a mapping with the [action type](../../../rules/ACTION-TYPES.md) as the pointer. When setting a rule to active in the handler, the status is applied per applicable [action type](../../../rules/ACTION-TYPES.md). The Handler Diamond then stores each action type and rule mapping into a struct within the [Rule Storage Facet](./PROTOCOL-ASSET-HANDLER-DIAMOND-FACET-LIST.md).
```
oscarsernarosero: I wouldn't go as deep as to mention the data structure types. If we want to show the data structures we can do that with a code block. I would just mention at a high level how the rules are stored as you have already done here.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
oscarsernarosero: They could tho. In theory, you just need to deploy a different proxy contract that points to the same facets as another handler diamond, and you have the exact same behavior without the cost of deploying the facets.

oscarsernarosero: That might actually be a good guideline for devs since that way they save money on deployments and upgrades

oscarsernarosero (changes requested):
This is an amazing job Tayler. Very minor suggestions.

docs/userGuides/Architecture/Client/AssetHandler/PROTOCOL-ASSET-HANDLER-DIAMOND.md
TJ-Everett: Yeah that was something we had talked about ages ago to try and keep parity with Zelda and then also limit cognitive requirements for non tron devs reading through docs.

mpetersoCode55: parity with who?

mpetersoCode55 (approved):
Looks great Tayler! I don't see anything else outside of what the guys already pointed out

oscarsernarosero (approved):
Looks good.

ShaneDuncan602 (approved):
Great job.

PR #281, opened 2024-02-22
Se 492 tron testing enhancement rule creation encapsulation functions

TJ-Everett:


src/protocol/economic/ruleProcessor/RuleCodeData.sol
```diff
@@ -8,7 +8,6 @@ bytes32 constant ACCOUNT_MAX_SELL_SIZE = "ACCOUNT_MAX_SELL_SIZE";
 bytes32 constant TOKEN_MAX_BUY_VOLUME = "TOKEN_MAX_BUY_VOLUME";
 bytes32 constant TOKEN_MAX_SELL_VOLUME = "TOKEN_MAX_SELL_VOLUME";
 bytes32 constant ACCOUNT_MIN_MAX_TOKEN_BALANCE = "ACCOUNT_MIN_MAX_TOKEN_BALANCE";
-bytes32 constant MIN_ACCT_BAL_BY_DATE = "MIN_ACCT_BAL_BY_DATE";
```
ShaneDuncan602, line 11: Good catch

test/util/RuleCreation.sol
ShaneDuncan602, line 1: I think these all should return the ruleId. If we have a testing scenario where multiple rules are created and applied to separate actions or tokens, we would have to discern the ruleId by the number of times the functions are called. If we used this on active deployments, we wouldn't know what ruleId to use.

ShaneDuncan602: Ok, I think there was some confusion on this one. I think the ticket was asking for the rule creation and activations to be in separate functions. The idea was to abstract out each step so that they could be as modular as possible and then each step could be added to a testXXXRule function.

ShaneDuncan602 (changes requested):
I think we need a little restructuring on this one.

test/client/token/ERC721/util/ERC721Util.sol
```diff
@@ -0,0 +1,124 @@
...
+
+
+/**
+ * @title Rule Creation Functions  
+ * @author @ShaneDuncan602 @oscarsernarosero @TJ-Everett
+ * @dev This contract is an abstract template to be reused by all the tests.
+ * This contract holds the functions for adding a protocol rule for tests.  
+ */
+
+abstract contract ERC721Util is RuleCreation {
```
mpetersoCode55: A lot of the code in here appears to be duplicated in ERC20Util.sol. Can we move anything that's shared between them out to a separate contract that they inherit to remove the duplicate code?

mpetersoCode55 (changes requested):
Looks great, really cleans things up! I just had one suggestion on a way to avoid code duplication

test/client/token/ERC721/util/ERC721Util.sol
ShaneDuncan602: That's a good idea. Maybe a SharedUtil that both ERC20Util and ERC721Util could inherit from.

ShaneDuncan602 (approved):
Looks really good

test/client/token/ERC721/util/ERC721Util.sol
TJ-Everett: Great idea guys! Added in a tokenUtils with all the shared rule creation functions.

test/client/token/ERC20/fuzz/ApplicationERC20Fuzz.t.sol
```diff
@@ -205,18 +189,15 @@ contract ApplicationERC20FuzzTest is TestCommonFoundry {
...
         uint8 risk = uint8((uint16(_risk) * 100) / 256);
         /// we give some trillions to user1 to spend
         applicationCoin.mint(user1, 10_000_000_000_000 * (10 ** 18));
 
         /// we register the rule in the protocol
-        switchToRuleAdmin();
-        uint32 ruleId = AppRuleDataFacet(address(ruleProcessor)).addAccountMaxTxValueByRiskScore(address(applicationAppManager), _maxValue, _riskScore, period, Blocktime);
-        /// now we set the rule in the applicationHandler for the applicationCoin only
-        applicationHandler.setAccountMaxTxValueByRiskScoreId(ruleId);
+        uint32 ruleId = createAccountMaxTxValueByRiskRule(25, 50, 75, 100_000_000, 10_000, 1, period);
```
oscarsernarosero: This looks amazing, but Im curious if you could specify an array as a parameter in these rule-creation functions, and then create the arrays at function-calling time. So, something like:
```
createAccountMaxTxValueByRiskRule( createUint8Array(25, 50, 75),  createUint48Array(100_000_000, 10_000, 1), period );
```
so this way the function signature doesn't rely on polymorphism but in variable-size arrays for their parameters which would be much more readable and will consolidate the rule-creation functions.

test/client/token/ERC20/fuzz/ApplicationERC20Fuzz.t.sol
```diff
@@ -261,27 +242,19 @@ contract ApplicationERC20FuzzTest is TestCommonFoundry {
...
         switchToRuleAdmin();
         /// let's deactivate the rule before minting to avoid triggering the rule
         applicationHandler.activateAccountMaxTxValueByRiskScore(false);
         switchToAppAdministrator();
         /// let get some trillions to user2 to spend
         applicationCoin.mint(user2, 90_000_000_000_000 * (10 ** 18));
 
+        _riskScore[0] = 10;
+        _riskScore[1] = 40;
+        _riskScore[2] = 90;
```
oscarsernarosero: not sure why we need this.

oscarsernarosero (changes requested):
This looks really good man. Im just wondering if we can shrink the rule-creation file by accepting variable-size arrays as parameters instead of polymorphism where it is not clear what each argument is.

test/client/token/ERC20/fuzz/ApplicationERC20Fuzz.t.sol
```diff
@@ -205,18 +189,15 @@ contract ApplicationERC20FuzzTest is TestCommonFoundry {
...
         uint8 risk = uint8((uint16(_risk) * 100) / 256);
         /// we give some trillions to user1 to spend
         applicationCoin.mint(user1, 10_000_000_000_000 * (10 ** 18));
 
         /// we register the rule in the protocol
-        switchToRuleAdmin();
-        uint32 ruleId = AppRuleDataFacet(address(ruleProcessor)).addAccountMaxTxValueByRiskScore(address(applicationAppManager), _maxValue, _riskScore, period, Blocktime);
-        /// now we set the rule in the applicationHandler for the applicationCoin only
-        applicationHandler.setAccountMaxTxValueByRiskScoreId(ruleId);
+        uint32 ruleId = createAccountMaxTxValueByRiskRule(createUint8Array(25, 50, 75), createUint48Array(100_000_000, 10_000, 1), period);
```
oscarsernarosero: Not sure if you use the `_riskScore` variable from line 192 somewhere else, but if you do then you should use it here as well. Otherwise prob better to just get rid of that line (192).

test/client/token/ERC20/fuzz/ApplicationERC20Fuzz.t.sol
```diff
@@ -290,21 +260,21 @@ contract ApplicationERC20FuzzTest is TestCommonFoundry {
         applicationCoin.transfer(user1, 1 * (10 ** 18));
         /// 1
         /// now, if the user's risk profile is in the highest range, this transfer should revert
-        if (risk >= _riskScore[2]) vm.expectRevert();
+        if (risk >= 90) vm.expectRevert();
```
oscarsernarosero, line 263: I think this is where you use `_riskScore`. So I would keep the line 192 and use the array fields here instead of hard coded values, and also use the array in the rule creation function since you already declared it.

test/client/token/ERC20/fuzz/ApplicationERC20Fuzz.t.sol
```diff
@@ -370,13 +333,8 @@ contract ApplicationERC20FuzzTest is TestCommonFoundry {
         uint48 riskBalance4 = _amountSeed;
         // add the rule.
         uint8[] memory _riskScore = createUint8Array(25, 50, 75, 90);
```
oscarsernarosero, line 335: same here

test/client/token/ERC20/integration/ApplicationERC20.t.sol
```diff
@@ -484,16 +423,17 @@ contract ApplicationERC20Test is TestCommonFoundry, DummyAMM {
 
         ///Assign Risk scores to user1 and user 2
         switchToRiskAdmin();
-        applicationAppManager.addRiskScore(user1, riskScores[0]);
-        applicationAppManager.addRiskScore(user2, riskScores[1]);
-        applicationAppManager.addRiskScore(user5, riskScores[3]);
+        applicationAppManager.addRiskScore(user1, 10);
+        applicationAppManager.addRiskScore(user2, 40);
+        applicationAppManager.addRiskScore(user5, 99);
```
oscarsernarosero: Im thinking that as a general rule for the riskScore case, it is better to declare the array before the rule creation so we can use it later for score assignment and testing.

oscarsernarosero (approved):
Great job man. Just some minor stuff around riskScores, but approving now.

PR #282, opened 2024-02-23
SE-549-Update-DOOM-to-align-with-Handler-refactor-into-a-diamond

bcamp136:
Just added the HandlerDeployed event to the HandlerDiamond.


TJ-Everett (approved):
Quick and easy! Looks perfect Brandon

PR #287, opened 2024-03-04
SE-514-Create-Invariant-Tests-for-Protocol

ShaneDuncan602:
This contains the Rule Storage invariant test infrastructure and examples for testing Token Min Transaction Size Rule


test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeActorManager.sol
```diff
@@ -0,0 +1,31 @@
...
+import "../util/RuleStorageInvariantActorManagerCommon.sol";
+
+/**
+ * @title RuleStorageTokenMinTxSizeActorManager
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett, @mpetersoCode55
+ * @dev This is the actor manager for the TokenMinTxSize rule. It will manage a variable number of invariant handlers
+ */
+contract RuleStorageTokenMinTxSizeActorManager is RuleStorageInvariantActorManagerCommon {
+
+    RuleStorageTokenMinTxSizeHandler[] handlers;
```
oscarsernarosero: Could we rename this for something like "agents" or "actors"?

test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeMulti.t.i.sol
```diff
@@ -0,0 +1,71 @@
...
+    // There can be only a total of 2**32 of each rule type.
+    function invariant_rulesTotalMinTxSizeLessThanMax() public {
+        assertLe(ERC20RuleProcessorFacet(address(ruleProcessor)).getTotalTokenMinTxSize(), maxRuleCount);
+    }
+
+    /// The next ruleId created in a specific rule type will always be the same as the previous ruleId + 1.
+    function invariant_rulesTotalMinTxSizeIncrementsByOne() public {
+        uint256 previousTotal = ERC20RuleProcessorFacet(address(ruleProcessor)).getTotalTokenMinTxSize();
+        // not incrementing previousTotal by one due to zero based ruleId
+        assertLe(previousTotal, RuleDataFacet(address(ruleProcessor)).addTokenMinTxSize(address(applicationAppManager), 1));
```
mpetersoCode55: Should this be assertLt instead of assertLe?

mpetersoCode55 (changes requested):
Looks great! One small question about one of the test cases

docs/userGuides/rules/ACCOUNT-APPROVE-DENY-ORACLE.md
```diff
@@ -178,7 +178,7 @@ The following validation will be carried out by the create function in order to
         ```
     - Function to know the activation state of the rule for the supplied action in an asset handler:
         ```c
-        function isOracleActive(ActionTypes _action, uint32 ruleId) external view returns (bool);
+        function isAccountAllowDenyOracleActive(ActionTypes _action, uint32 ruleId) external view returns (bool);
```
TJ-Everett, line 181: Great catch

TJ-Everett (approved):
This is BADASS Shane! Approved for me.

test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeMulti.t.i.sol
oscarsernarosero: this assertion makes sure that the new total is incrementing, but not necessarily checks that it is incrementing by 1. I think it is better to assert that new total is exactly the old one +1.

test/util/TestCommonFoundry.sol
```diff
@@ -18,72 +18,144 @@ abstract contract TestCommonFoundry is TestCommon {
...
-            bytes memory res = vm.ffi(inputs);
-            bytes4[] memory selectors = abi.decode(res, (bytes4[]));
-
-            // Create the FacetCut struct for this facet.
-            _ruleProcessorFacetCuts.push(FacetCut({facetAddress: facetAddress, action: FacetCutAction.Add, functionSelectors: selectors}));
-        }
-
+        _addNativeFacetsToFacetCut();
+        _addStorageFacetsToFacetCut();
+        _addProcessingFacetsToFacetCut();
```
oscarsernarosero, line 26: Im just not sure how this is different than looping through the string array of facets. Since this is a model for us to replicate, I think it is better to make sure this is the most optimal way of doing it.

test/util/TestCommonFoundry.sol
```diff
@@ -18,72 +18,144 @@ abstract contract TestCommonFoundry is TestCommon {
...
     }
 
+    function _addNativeFacetsToFacetCut() public {
+        // Protocol Facets
+        ProtocolNativeFacet protocolNativeFacet = new ProtocolNativeFacet();
+        ProtocolRawFacet protocolRawFacet = new ProtocolRawFacet();
+        VersionFacet versionFacet = new VersionFacet();
+        string[] memory inputs = new string[](3);
+        inputs[0] = "python3";
+        inputs[1] = "script/python/get_selectors.py";
```
oscarsernarosero: I thought I had encapsulated this process, but it looks like I only encapsulated the new scripts that I made in Python. I think it would be a good idea to encapsulate this in this ticket so this model can be replicated without fear of spreading code duplication. You can checkout `script/DiamondScriptUtil.sol` to see what I did for the other scripts if you want, but it is quite simple. I guess the harder part would be to decide where to put that function since it would be used in the scripts and the tests.

test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeActorManager.sol
oscarsernarosero: or rappers? jk... wrappers?

oscarsernarosero (changes requested):
Great job overall Shane. Just some minor suggestions since this will be the model to replicate.

test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeMulti.t.i.sol
ShaneDuncan602: Yep, reworking this one.

test/protocol/economic/invariant/rules/tokenMinTxSize/RuleStorageTokenMinTxSizeActorManager.sol
ShaneDuncan602: Going with actors. It's a good point because our use of handler will just confuse things. I don't like straying from foundry's terms but it makes sense for us.

test/util/TestCommonFoundry.sol
ShaneDuncan602: Yeah, this was to allow foundry to pick up the facets and their functions automatically. However, we are using the actor manager model for the invariants so we define the actors and their interfaces ourselves. I'm fine with reverting this back, but I left it in case we want to incorporate the other model and I think it's good to have the facets defined now that the design is stable. That way, the compiler will keep us honest if we change things.

oscarsernarosero (approved):
Nice job!

PR #288, opened 2024-03-05
Se 592 update solidity compiler version

mpetersoCode55:
Updated solidity, foundry and OZ dependencies to use the latest versions of each.

### DIFF
--- a/docs/userGuides/deployment/DEPLOY-PROTOCOL.md
+++ b/docs/userGuides/deployment/DEPLOY-PROTOCOL.md
@@ -35,13 +35,6 @@ is an overview of this deployment process:
    foundryUp
    ````
 
-   #### *Troubleshooting*
-
-   *if deployment fails (following the next steps), it might be caused by a faulty current version of Foundry. Try installing this specific version instead:*
-
-   ```bash
-   foundryup --version nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
-   ```
 6. In the same terminal as above, navigate to the cloned repo directory and run the build script
    ````
 	bash script/deploy/DeployProtocol.sh
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "version": "1.1.0",
   "description": "Set of contracts to easily start an application in rules protocol",
   "dependencies": {
-    "@openzeppelin/contracts": "4.9",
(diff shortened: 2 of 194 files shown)


ShaneDuncan602:
We also need to change the main README to reflect the changes. It still has 4.9 for the versions and talks about npm install. That should have forge install with the proper version.

Also, it looks like there are also still dependencies in the github workflows.

script/python/get_selectors.py
```diff
@@ -8,11 +8,12 @@
...
-    res = subprocess.run(
-        ["forge", "inspect", contract, "mi"], capture_output=True)
-    res = res.stdout.decode()
-    res = json.loads(res)
+    line = ""
+    while line == "":
+        res = subprocess.run(
+            ["forge", "inspect", contract, "mi"], capture_output=True )
+        line = res.stdout.decode()
+    res = json.loads(line)
```
ShaneDuncan602, line 16: This was an awesome change.

ShaneDuncan602 (changes requested):
Found a few missed dependencies but this is fantastic. Great work.

docs/userGuides/deployment/DEPLOY-PROTOCOL.md
```diff
@@ -35,13 +35,6 @@ is an overview of this deployment process:
    foundryUp
    ````
 
-   #### *Troubleshooting*
-
-   *if deployment fails (following the next steps), it might be caused by a faulty current version of Foundry. Try installing this specific version instead:*
-
-   ```bash
-   foundryup --version nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
-   ```
```
oscarsernarosero, line 44: great catch

package.json
```diff
@@ -7,8 +7,8 @@
   "version": "1.1.0",
   "description": "Set of contracts to easily start an application in rules protocol",
   "dependencies": {
-    "@openzeppelin/contracts": "4.9",
-    "@openzeppelin/contracts-upgradeable": "4.9"
+    "@openzeppelin/contracts": "5.0.2",
+    "@openzeppelin/contracts-upgradeable": "5.0.2"
```
oscarsernarosero, line 11: nice

oscarsernarosero (approved):
Looks great Michael. Im glad it didn't require lots of detail fixing.

PR #294, opened 2024-03-06
Temporary Github Action to test AWS CLI changes

excalq:
- Simply removes the `Install AWS CLI` step, as it likely is installed by the `Configure AWS Credentials` step.
- Creates a new action, which only sets up AWS, but does not deploy, which can be invoked manually.


bcamp136 (approved):
Looks good to me

PR #296, opened 2024-03-06
Se 562 no account with multiple roles

TJ-Everett:


.env
```diff
@@ -52,15 +54,16 @@ FRANKS=FRANK
...
 ## Forge Fuzz Test Variables
 FOUNDRY_FUZZ_RUNS=8192
 
 ## Current Deployment details(these are used by the deploy scripts)
 # commit to repo with 0x0000000000000000000000000000000000000000 
 DEPLOYMENT_OWNER=0x0000000000000000000000000000000000000000
-DEPLOYMENT_OWNER_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
-APP_ADMIN=0x75df377cb2fd583FDe95570766639E7B46c014cb
-APP_ADMIN_PRIVATE_KEY=0x
+DEPLOYMENT_OWNER_KEY=0x8b3a350cf5c34c9194ca85829a2df0ec3153be0318b5e2d3348e872092edffba
```
mpetersoCode55, line 64: Should this be filled out? I assume since the Deployment Owner is zeroed out this should be as well

script/ParseProtocolDeploy.sh
```diff
@@ -37,6 +37,9 @@ echo
 QUORRA=$(sed -n 's/QUORRA=//p' .env)
 QUORRA_PRIVATE_KEY=$(sed -n 's/QUORRA_PRIVATE_KEY=//p' .env) 
 
+APP_ADMIN_01=$(sed -n 's/APP_ADMIN_01=//p' .env)
```
mpetersoCode55, line 40: Noice, Tayler's just over here bashing away

mpetersoCode55 (changes requested):
Looks great Tayler! Just one question about an .env file entry

.env
oscarsernarosero: Agree. They both should be 0 or both should be non-zero values.

oscarsernarosero (approved):
This looks amazing Tayler!

PR #305, opened 2024-03-12
Atomic set rule copy

bfcoursewool:
- Moving the activate/deactivate events out of the loops in the functions that emit them, and emitting the entire actions array at once instead of in sequence is a lot more convenient on the front-end so I just went ahead and made that change.
- Two of the setRuleFull methods were emitting HandlerSimpleActionApplied instead of just HandlerActionApplied like the others, so I changed those two
- took off the 'indexed' keyword for a couple parameters in the new events so it's easier to read the data in the indexer
- change event signatures accordingly


ShaneDuncan602 (approved):
Good stuff here. The simple version of that event was originally to specify that it was a simple rule(one that is stored only in the handler and not in the RuleProcessor) but it's probably not useful information at this point. I'll ax it completely. I'm merging this into my branch.

PR #312, opened 2024-03-13
Create logic trees for Handler functions

TJ-Everett:


docs/userGuides/Architecture/Client/Application/APPLICATION-HANDLER.md
```diff
@@ -11,12 +11,51 @@ The Application Handler supports the Application Manager by storing the applicat
...
+├── or when accountMaxValueByAccessLevelActive is true
+├── or when accountMaxValueOutByAccessLevelActive is true
+└── or when accountDenyForNoAccessLevelRuleActive is true
+    └── it should return true
+    
 ```
 This function allows the Application Manager to know if any application level rules are active and if the call should continue to the handler to check the active rules. 
 
 The Application Manager then calls the function: 
 ```c
```
excalq, line 26: FYI, this syntax highlight makes all occurrences of keywords (if, continue, return) show as red in the tree descriptions.

TJ-Everett: For everyone review: Should I break these out into a new code block without syntax highlighting to preserve the function highlighting, are we okay with those keywords being highlighted, or no keyword highlighting?

docs/userGuides/Architecture/Client/Application/APPLICATION-HANDLER.md
```diff
@@ -11,12 +11,51 @@ The Application Handler supports the Application Manager by storing the applicat
...
  function checkApplicationRules(address _tokenAddress, address _from, address _to, uint256 _amount, uint16 _nftValuationLimit, uint256 _tokenId, ActionTypes _action, HandlerTypes _handlerType) external onlyOwner returns (bool)
+├── when the caller is not the owner
+│ └── it should revert
+└── when the caller is the owner
+    └── it should call the rule processor diamond and validate the transaction when application level rules are active 
+        ├── when pauseRuleActive rules are active 
+        │ └── it should validate pause rules through the rule processor diamond 
+        │    ├── if the rule processor diamond returns true
```
mpetersoCode55: Should these if's be when's for consistency with the rest of the syntax?

docs/userGuides/Architecture/Client/Application/APPLICATION-HANDLER.md
mpetersoCode55: I would lean strongly towards preserving the syntax highlighting you've done here Tayler. Since this is how we're highlighting code in all of Tron's documentation. 
If there's a better way to go I'm all for it, but I think that should be its own dedicated ticket to make the updates throughout all of our documentation and keep them all in line with each other.

mpetersoCode55 (approved):
Looks great TDT! Just one suggestion

ShaneDuncan602 (approved):
Great job. I know this wasn't easy but it's really helpful.

PR #313, opened 2024-03-13
Consolidated Application Deployment and Integration tests.

mpetersoCode55:
Also updated gitmodules to point back to OZ 4.9


ShaneDuncan602 (approved):
Looks great.

PR #314, opened 2024-03-13
[draft]: Get upgrade working

VoR0220:


src/client/token/ERC721/ProtocolERC721.sol
```diff
@@ -19,10 +18,9 @@ import "src/client/token/handler/diamond/IHandlerDiamond.sol";
  */
 
 contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, Pausable, ProtocolTokenCommon, AppAdministratorOrOwnerOnly, ReentrancyGuard {
-    using Counters for Counters.Counter;
     address public handlerAddress;
     IHandlerDiamond handler;
-    Counters.Counter internal _tokenIdCounter;
+    uint256 internal _tokenIdCounter;
```
ShaneDuncan602, line 23: I'm glad OZ has gotten rid of Counter. I never liked it. How difficult is it to just increment a counter? This is much more straight forward.

src/client/token/ERC721/ProtocolERC721.sol
```diff
@@ -93,35 +91,60 @@ contract ProtocolERC721 is ERC721Burnable, ERC721URIStorage, ERC721Enumerable, P
...
      * @param tokenId Id of token to be transferred
-     * @param batchSize the amount of NFTs to mint in batch. If a value greater than 1 is given, tokenId will
-     * represent the first id to start the batch.
+     * @param auth Auth argument is optional. If the value passed is non 0, then this function will check that
+     * `auth` is either the owner of the token, or approved to operate on the token (by the owner).
      */
-    function _beforeTokenTransfer(address from, address to, uint256 tokenId, uint256 batchSize) internal nonReentrant override(ERC721, ERC721Enumerable) whenNotPaused {
-        // Rule Processor Module Check
-        require(IHandlerDiamond(handler).checkAllRules(from == address(0) ? 0 : balanceOf(from), to == address(0) ? 0 : balanceOf(to), from, to, _msgSender(), tokenId));
+    function _update(address to, uint256 tokenId, address auth) internal override (ERC721, ERC721Enumerable) returns (address) {
```
ShaneDuncan602, line 123: This is going to be a problem for us. The beforeTokenTransfer has the "from" address in it but update does not. In our action determination logic, we need to know when the "from" address is different than the message.sender. That's helps us identify buys, sells, and even mints. I don't think adding our hook in this function is going to work for us. We will have to create a modifier that goes on their overridden transfer function. This definitely will require some planning and I'm not sure we want to pursue such a fundamental protocol change this close to the audit.

PR #316, opened 2024-03-14
fixing manifests

bfcoursewool:
Fixing the k8s manifests from my overzealous sublime text formatter...


excalq (approved):
Assuming the submodule changes are as expected

bfcoursewool:
> Assuming the submodule changes are as expected

Oh hm... I hadn't noticed those... I cut this branch from main and I had just pulled the latest from main right before I did it. So I'm not sure what exactly is going on there...

PR #336, opened 2024-04-05
added tests to attempt to break gas limit on getTotalAccValuation

VoR0220:
Findings: 

1. Nft limit valuation will do the trick, it will take unbounded gas usage that can reach 100 million units of gas used and convert it downwards to levels of about 69k units of gas used. Going forward, this should be assumed to be 0 when interacting with the protocol unless given a significant suggestion as to why not (perhaps around discussions of potential arbitrage from values not being counted properly from NFT standpoint). 

2. A large set of NFTs and ERC20s (which we've set to about 20 for a default but can be configurable beyond here), will yield  a gas usage of 278047. Thats 278k to evaluate 20 NFTs and 20 ERC20s for their price and how it tracks when used within the EVM. 

3. Created a new file to handle application manager interactions that use both tokens.

### DIFF
--- a/test/client/application/ProtocolApplicationHandlerTests.t.sol
+++ b/test/client/application/ProtocolApplicationHandlerTests.t.sol
@@ -33,7 +33,7 @@ contract ProtocolApplicationHandlerTests is TestCommonFoundry {
         uint gasEnd = gasleft();
         // console2.log("gas: ", gasBegin - gasEnd);
         // console2.log("valuation: ", valuation);
-        assertEq(valuation, loops * 100 * 100 * 1e18);
+        assertEq(valuation, loops * 100 * 100 * 1e17);
         assertLt(gasBegin - gasEnd, 15000000); // assert that it is less than the gas limit
     }
 }
--- a/test/client/application/ProtocolApplicationHandlerTests.t.sol
+++ b/test/client/application/ProtocolApplicationHandlerTests.t.sol
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.24;
+
+import "test/util/TestCommonFoundry.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+
+// tests the shared functionality of the erc20 and erc721 combined together
+
+contract ProtocolApplicationHandlerTests is TestCommonFoundry {
+
+    function setUp() external {
+        vm.warp(Blocktime);
+        setUpProcotolAndCreateERC20AndDiamondHandler();
+    }
+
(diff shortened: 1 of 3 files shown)


test/client/token/ERC721/integration/ERC721CommonTests.t.sol
```diff
@@ -1326,6 +1326,138 @@ abstract contract ERC721CommonTests is TestCommonFoundry, ERC721Util {
...
+        ERC721HandlerMainFacet(address(applicationNFTHandler)).setNFTValuationLimit(10);
+        
+        // track the gas
+        uint gasBegin = gasleft();
+        uint totalValuation = applicationHandler.getAccTotalValuation(appAdministrator, 10);
+        uint gasEnd = gasleft();
+        
+        //console2.log("Gas: ", gasBegin - gasEnd);
+
+        assertLt(gasBegin - gasEnd, 70000); // assert that this is less than 70k. This gives us a baseline to understand what our lower bound should be.
```
ShaneDuncan602, line 1457: These tests need to also work as fork tests. Will these numbers still work when run against a mainnet fork?

test/client/application/ProtocolApplicationHandlerTests.t.sol
```diff
@@ -0,0 +1,39 @@
...
+contract ProtocolApplicationHandlerTests is TestCommonFoundry {
+
+    function setUp() external {
+        vm.warp(Blocktime);
+        setUpProcotolAndCreateERC20AndDiamondHandler();
+    }
+
+    // note: make a test for get acc total valuation 
+    function test_getAccTotalValuation_SlamWithTokensFindGasBreakageValueLimitLow() public {
+        uint loops = vm.envOr("SLAM_TOKEN_LOOPS", uint(20));
```
ShaneDuncan602, line 18: I don't see this variable in the .env. It's currently in the .gitIgnore so you may have to manually add it.

ShaneDuncan602 (changes requested):
Looks really good. Just have some minor feedback.

test/client/token/ERC721/integration/ERC721CommonTests.t.sol
VoR0220: They absolutely should as gas as a unit is generally speaking a constant. The difference is the cost per unit across platforms. As long as we're forking onto anvil I don't see an issue.

test/client/application/ProtocolApplicationHandlerTests.t.sol
VoR0220: Added

ShaneDuncan602 (approved):
Looks good!

PR #337, opened 2024-04-11
Se 618 update npm package

oscarsernarosero:
# DO NOT MERGE BEFORE THE BIG SWARM
#### DO NOT MERGE BEFORE THE BIG SWARM

This npm package configuration was updated in preparation to the official launch of our npm package which will be done after the audit. This package is registered in a separate test package in the registry [here](https://www.npmjs.com/package/@thrackle-io/rules-protocol-client-test). The scripts and the package.json were updated to work with the new directory structure of the repo, and it was tested in a hardhat-project environment.

The [documentation](https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/185860099/NPM+Procedure+-+rules-protocol) was also updated to add support to Foundry projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,20 +34,22 @@ forge install OpenZeppelin/openzeppelin-contracts-upgradeable
 ## Usage
 
 ### User Guides
+
 For complete usage information and documentation, please visit our [User Guide][userGuide-url].
 
 ### A Simple Example
+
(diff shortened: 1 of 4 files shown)


TJ-Everett (approved):
Looks great! Compiles and tests on Hardhat and Foundry

PR #343, opened 2024-04-12
SE-630-Create-tests-for-action-identification-mechanism

ShaneDuncan602:


src/client/token/handler/common/HandlerUtils.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity ^0.8.24;
 import {ActionTypes} from "src/common/ActionEnum.sol";
 
 contract HandlerUtils{
+    event Action(uint8 _type);
```
oscarsernarosero, line 7: You can declare the event like this:
```
event Action(ActionTypes indexed _type);
```

This way you don't have to cast the action type to uint8 (it would do it automatically tho), and it would be queryable through the type since it is indexed.

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
         from = user1;
         assertEq(uint8(ActionTypes.SELL), uint8(determineTransferAction(from, to, sender)));
     }
+
+    /** Test that actions are properly determined when calling determineTransfer directly */
+    function testERC20_testSmartContractWallet() public {
+        SBAWallet wallet = new SBAWallet();
+        // make sure that wallet can hold ETH
+        vm.deal(address(wallet), 10 ether);
+        assertEq(10 * ATTO, wallet.getWalletBalance());
```
oscarsernarosero, line 97: not an actual suggestion cuz it doesn't really matter, but since we are dealing with Ether here, it would be better to use `ether` here instead of "ATTO".

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        assertEq(uint8(ActionTypes.P2P_TRANSFER), uint8(determineTransferAction({_from: user1, _to: address(wallet), _sender: user1})));
+
+        // Set up amm for buys and sells
+        DummyAMM amm = initializeERC20AMM(address(applicationCoin), address(applicationCoin2));
+
+        // test Sells 
+        assertEq(uint8(ActionTypes.SELL), uint8(determineTransferAction({_from: user1, _to: address(amm), _sender: address(amm)})));
+
+        // test Buys
+        assertEq(uint8(ActionTypes.BUY), uint8(determineTransferAction({_from: address(amm), _to: user1, _sender: address(amm)})));
```
oscarsernarosero, line 116: Shane I don't really like that we deploy an SBWallet and an AMM here since we are not really using them. It is kind of misleading. Since we are just testing the function, we don't really need to deploy anything.

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        vm.startPrank(user1);
+        // from EOA to SCA
+        vm.expectEmit();
+        emit Action(uint8(ActionTypes.P2P_TRANSFER));
+        applicationCoin.transfer(address(wallet),1);
+        // from EOA to EOA
+        vm.expectEmit();
+        emit Action(uint8(ActionTypes.P2P_TRANSFER));
+        applicationCoin.transfer(user2,1);
+    }
```
oscarsernarosero, line 162: I would split this test. One for EOAs to EOAs, another for EOAs to SCAs, another one for SCAs to EOAs, and [another one for SCAs to SCAs] (I know this last one probably not since it would fail, but at least leaving a place holder or something)

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        applicationCoin2.mint(address(wallet), 10 * ATTO);
+        
+        // Set up amm for buy and sell tests
+        DummyAMM amm = initializeERC20AMM(address(applicationCoin), address(applicationCoin2));
+        // test Sells 
+        vm.startPrank(address(wallet));
+        applicationCoin2.approve(address(amm), 50000);
+        applicationCoin.approve(address(amm), 50000);
+
+        // test Sell should issue two events: 1. sell from user, 2. buy from amm
```
oscarsernarosero: all the events are emitted from the user's perspective. Both buy and sell are from the user's perspective since he is selling one token to buy the other one. So this comment should reflect that.

oscarsernarosero: maybe it is more from the token perspective, so it should say `1. sell for applicationCoin, 2. buy for applicationCoin2`

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        applicationCoin2.mint(address(wallet), 10 * ATTO);
+        
+        // Set up amm for buy and sell tests
+        DummyAMM amm = initializeERC20AMM(address(applicationCoin), address(applicationCoin2));
+        // test Sells 
+        vm.startPrank(address(wallet));
+        applicationCoin2.approve(address(amm), 50000);
+        applicationCoin.approve(address(amm), 50000);
+
+        // test Buys should issue two events: 1. sell from amm 2. buy from user
```
oscarsernarosero: same here

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
oscarsernarosero: ok nm. I see you did this in testERC20_SmartContractWalletBroken

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        vm.deal(address(wallet), 10 ether);
+        assertEq(10 * ATTO, wallet.getWalletBalance());
+
+        vm.startPrank(address(wallet));
+        // test transfers 
+        // from SCA to EOA(comes back as BUY)
+        assertEq(uint8(ActionTypes.P2P_TRANSFER), uint8(determineTransferAction({_from: address(wallet), _to: user1, _sender: address(wallet)})));
+        // from SCA to SCA(comes back as BUY)
+        assertEq(uint8(ActionTypes.P2P_TRANSFER), uint8(determineTransferAction({_from: address(wallet), _to: address(wallet), _sender: address(wallet)})));
+    }
```
oscarsernarosero: I like that this is separated from the test that passes, but I think we should try to stick to the premise of "test one and only one thing". So we should split this test in 2 different tests.

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
oscarsernarosero: I still think we should split the test in 2: EOA to SCA and EOA to EOA

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+    /**
+        This test demonstrates current shortcomings with the protocol's detection of actions when a SCA is used. These issues will be addressed in v2.
+        TODO: Fix the determine transfer action function to work with SCA
+     */
+    function testERC20_SmartContractWalletBroken() public {
+        vm.skip(true);
+        SBAWallet wallet = new SBAWallet();
+        // make sure that wallet can hold ETH
+        vm.deal(address(wallet), 10 ether);
+        assertEq(10 * ATTO, wallet.getWalletBalance());
```
oscarsernarosero: I don't think we should test this here since we already have a test for that earlier

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        This test demonstrates current shortcomings with the protocol's detection of actions when a SCA is used. These issues will be addressed in v2.
+        TODO: Fix the determine transfer action function to work with SCA
+     */
+    function testERC20_SmartContractWalletWithProtocolSupportedAssetsBroken() public {
+        vm.skip(true);
+        SBAWallet wallet = new SBAWallet();
+        SBAWallet wallet2 = new SBAWallet();
+        // make sure that wallet can hold ETH
+        vm.deal(address(wallet), 10 ether);
+        assertEq(10 * ATTO, wallet.getWalletBalance());
```
oscarsernarosero: same here

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -51,4 +58,204 @@ contract ApplicationERC20HandlerTest is Test, HandlerUtils {
...
+        // test transfers 
+        // from SCA to EOA(comes back as BUY)
+        vm.expectEmit();
+        emit Action(uint8(ActionTypes.P2P_TRANSFER));
+        applicationCoin.transfer(user1,1);
+        // from SCA to SCA(comes back as BUY)
+        vm.expectEmit();
+        emit Action(uint8(ActionTypes.P2P_TRANSFER));
+        applicationCoin.transfer(address(wallet2),1);
+    }
```
oscarsernarosero: we should also split the tests here.

test/client/token/ERC20/util/SBAWallet.sol
```diff
@@ -0,0 +1,47 @@
...
+return address(this).balance;
+}
+
+//To know the owner balance
+function getOwnerBalance() external view returns (uint){
+return address(owner).balance;
+}
+
+
+}
```
oscarsernarosero, line 54: I like it. I just don't know why this whole file is not properly indented.

oscarsernarosero (changes requested):
This is good Shane. I just think we should be very disciplined with the way we structured our tests from now on so we don't have to go through huge swarms in the future. Everything else looks great!

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
ShaneDuncan602: It was just some overkill to make sure that the different types of contracts are in there. If we ever add an interface check, this could come in handy.

PR #350, opened 2024-04-18
Fixed deployment issues caused by underlying changes to foundry vm.ge…

mpetersoCode55:
…tCode

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -2,18 +2,23 @@
 src = 'src'
 out = 'out'
 libs = ['lib']
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.local]
 eth-rpc-url = 'http://localhost:8545'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.docker]
 eth-rpc-url = 'http://anvil:8545'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.deployment-check]
 eth-rpc-url = 'http://anvil-deploy-check:8545'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.polygon]
 eth-rpc-url = 'http://172.20.0.2:10002'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 remappings = ["forge-std/=lib/forge-std/src/",
     "ds-test/=lib/forge-std/lib/ds-test/src/",
--- a/install-deploy-check.sh
+++ b/install-deploy-check.sh
@@ -8,8 +8,11 @@ python3 -m venv .venv
 source .venv/bin/activate
 python3 -m pip install -r requirements.txt
 
+if [ $SCRIPT_MODE = "--with-build" ]; then
+  forge build   
+fi
+
 if [ $SCRIPT_MODE = "--with-deploy-check" ]; then
-  forge build
   source script/SetupProtocolDeploy.sh
   forge script script/DeployAllModulesPt1.s.sol --ffi --broadcast
   source script/ParseProtocolDeploy.sh
--- a/install-deploy-check.sh
+++ b/install-deploy-check.sh
@@ -8,11 +8,8 @@ python3 -m venv .venv
 source .venv/bin/activate
 python3 -m pip install -r requirements.txt
 
-if [ $SCRIPT_MODE = "--with-build" ]; then
-  forge build   
-fi
-
 if [ $SCRIPT_MODE = "--with-deploy-check" ]; then
+  forge build
   source script/SetupProtocolDeploy.sh
   forge script script/DeployAllModulesPt1.s.sol --ffi --broadcast
   source script/ParseProtocolDeploy.sh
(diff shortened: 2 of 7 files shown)


script/DeployAllModulesPt1.s.sol
```diff
@@ -98,7 +108,7 @@ contract DeployAllModulesPt1Script is Script, DiamondScriptUtil {
             initCalldata: abi.encode(type(IDiamondInit).interfaceId)
         });
 
-        /// Deploy the diamond.
+        // /// Deploy the diamond.
```
VoR0220: can we bring this to only ///?

script/clientScripts/DeployBase.s.sol
```diff
@@ -178,9 +201,10 @@ contract DeployBase is Script, DiamondScriptUtil {
         // Loop on each facet, deploy them and create the FacetCut.
         for (uint256 facetIndex = 0; facetIndex < facets.length; facetIndex++) {
             string memory facet = facets[facetIndex];
+            string memory directory = directories[facetIndex];
 
-            // Deploy the facet.
-            bytes memory bytecode = vm.getCode(string.concat(facet, ".sol"));
+            /// Deploy the facet.
```
VoR0220: ///

VoR0220 (approved):

foundry.toml
```diff
@@ -2,18 +2,23 @@
...
 eth-rpc-url = 'http://anvil:8545'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.deployment-check]
 eth-rpc-url = 'http://anvil-deploy-check:8545'
+fs_permissions = [{ access = "read", path = "./out"}]
 
 [profile.polygon]
 eth-rpc-url = 'http://172.20.0.2:10002'
+fs_permissions = [{ access = "read", path = "./out"}]
```
VoR0220, line 21: Can we get an explanation on how this works or a link to it? I imagine this is the breaking change right here.

mpetersoCode55: So essentially vs.getCode no longer pulls from filesystem based solidity files but instead pulls from the compiled json output in the out directory.
The foundry vm calls do not have read access to the out directory by default however, so we had to add this line to our foundry.toml on top of refactoring how we grab the files with vm.getCode

script/clientScripts/DeployBase.s.sol
mpetersoCode55: Good catch!

script/DeployAllModulesPt1.s.sol
mpetersoCode55: Good catch!

excalq (changes requested):
Posting details of why in a minute...

foundry.toml
VoR0220: Very useful knowledge. Hope they updated their docs to accompany this.

install-deploy-check.sh
```diff
@@ -8,11 +8,8 @@ python3 -m venv .venv
 source .venv/bin/activate
 python3 -m pip install -r requirements.txt
 
-if [ $SCRIPT_MODE = "--with-build" ]; then
-  forge build   
-fi
-
 if [ $SCRIPT_MODE = "--with-deploy-check" ]; then
+  forge build
```
excalq: [Dockerfile.tron](https://github.com/thrackle-io/tron/blob/main/Dockerfile.tron#L13-L15) needs to run build exclusively during the image build (the `RUN` step), and not on container startup (the `CMD` step). This causes every startup of the container to re-compile.

excalq (changes requested):

install-deploy-check.sh
mpetersoCode55: That makes sense, I'll set this back. I assume the build version is always run before any form of deployment? Deploying no longer works without an explicit build first

install-forge.sh
```diff
@@ -9,11 +9,8 @@ python3 -m venv .venv
 source .venv/bin/activate
 python3 -m pip install -r requirements.txt
 
-if [ $SCRIPT_MODE = "--with-build" ]; then
-	forge build
-fi
-
 if [ $SCRIPT_MODE = "--with-deploy" ]; then
```
excalq: Here as well

foundry.toml
mpetersoCode55: They did not, all of their examples in their docs for vs.getCode no longer work

install-deploy-check.sh
excalq: Yes, the Dockerfile always will run `RUN ./install-forge.sh --with-build` when the image is created. Deployment or any restart of the workload in K8s (or in Docker-Compose) will run `./install-forge.sh --with-deploy` at startup. (Maybe that could be better named?)

foundry.toml
excalq: Might be worth keeping a comment above the first use of it, with that info.

excalq (approved):
Thanks for the deployment fix

PR #352, opened 2024-04-23
SE-626-Add-feature-to-deployment-scripts-to-attach-ABIs-to-specific-deployments

ShaneDuncan602:
Created functionality to snapshot abi by chainId and timestamp. Also tweaked the diamond facet json snapshot.


deployments/diamonds.json
```diff
@@ -1,141 +0,0 @@
...
-                "ProtocolRawFacet": "0xbC9aF6bce6A0F0b6Ee74C778308a17750C801021",
-                "ERC20HandlerMainFacet": "0xa718d4dfb89FF08B6566591b0E98eC05f71A41ae",
-                "diamond": "0x0929D4EAA10788D81B5e3c7F3A456A487cd519e8",
-                "ERC20TaggedRuleFacet": "0xE53Df7b962649423EdA42095768Ffd095fA9fBa9",
-                "ERC20NonTaggedRuleFacet": "0xd6564f729e3906E6AdB2226e651B3749a1d8b006",
-                "TradingRuleFacet": "0x4a7F575a2a6d54E254f77ff296D8715bf971111a",
-                "FeesFacet": "0x4a458a11C32FFf1e5d919a6a7E7f368b5C1e7AC1"
-            }
-        }
-    }
```
VoR0220, line 140: While we could get this from the git logs, in order to avoid this becoming a PITA, do we have this stored elsewhere currently?

script/DeployABIUtil.sol
```diff
@@ -0,0 +1,26 @@
...
+    
+
+    function recordABI(string memory contractName, bool recordAllChains, uint256 timestamp) internal {
+        string[] memory recordABIInput = new string[](6);
+        recordABIInput[0] = "python3";
+        recordABIInput[1] = "script/python/record_abi.py";
+        recordABIInput[2] = contractName;
+        recordABIInput[3] = vm.toString(block.chainid);
+        recordABIInput[4] = vm.toString(timestamp);
+        recordABIInput[5] = recordAllChains ? "--allchains" : "--no-allchains";
```
VoR0220, line 21: How would I record an ABI for a singular chain if I only deployed a facet onto one chain and one chain only? I ask this because it's possible that down the line we need to make a hot fix for one chain/rollup's interpretation of the EVM (zkEVM for example), and don't need to apply this to others. How would we handle that with this setup? It seems we've accounted for it to be not on all chains, but how do I specify?

script/deploy/DeployProtocol.sh
```diff
@@ -22,10 +22,16 @@ echo "################################################################"
...
 echo
-forge script script/DeployAllModulesPt3.s.sol --ffi --broadcast -vvv --non-interactive
\ No newline at end of file
+forge script script/DeployAllModulesPt3.s.sol --ffi --broadcast -vvv --non-interactive --rpc-url=$ETH_RPC_URL
+
+echo "################################################################"
+echo Build, Deploy, and Add Facets Unit 3
+echo "################################################################"
+echo
+forge script script/DeployAllModulesPt4.s.sol --ffi --broadcast -vvv --non-interactive --rpc-url=$ETH_RPC_URL
```
VoR0220, line 37: ++ on fixing bugs in the scripts

VoR0220:
I just have a couple questions really just to be sure we're a) aware of what our tools can do and b) making sure we don't create an annoyance for us later. Overall this looks great and should be gtm.

VoR0220 (approved):

deployments/diamonds.json
ShaneDuncan602: The address, you mean? I don't understand.

script/DeployABIUtil.sol
ShaneDuncan602: If it's a hot fix, it will require a script that uses the recordABI function to pull the abi for the single fix or just copying the abi into the deployment folder would work.

deployments/diamonds.json
VoR0220: The addresses and their their contract names. Basically the entire file's contents.

script/DeployABIUtil.sol
VoR0220: Let me ask a second question then. When's a time when I'd use "---no-allchains"? Is this the "don't record anything" flag?

VoR0220: Nvm just confirmed this is in fact the case.

TJ-Everett (approved):
Approved for me. Script runs great (when I follow the clearly written instructions)

PR #356, opened 2024-04-24
SE-760-Resolve-Audit-Findings-Non-Security-Missing-Zero-address-checks

ShaneDuncan602:
Added Zero Address checks.

### DIFF
--- a/src/client/application/data/AccessLevels.sol
+++ b/src/client/application/data/AccessLevels.sol
@@ -53,6 +53,7 @@ contract AccessLevels is IAccessLevels, DataModule {
     function addAccessLevelToMultipleAccounts(address[] memory _accounts, uint8 _level) external virtual onlyOwner {
         if (_level > 4) revert AccessLevelIsNotValid(_level);
         for (uint256 i; i < _accounts.length; ++i) {
+            if (_accounts[i] == address(0)) revert ZeroAddress();
             levels[_accounts[i]] = _level;
             emit AD1467_AccessLevelAdded(_accounts[i], _level);
         }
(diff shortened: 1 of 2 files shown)


src/client/application/data/AccessLevels.sol
```diff
@@ -53,6 +53,7 @@ contract AccessLevels is IAccessLevels, DataModule {
     function addAccessLevelToMultipleAccounts(address[] memory _accounts, uint8 _level) external virtual onlyOwner {
         if (_level > 4) revert AccessLevelIsNotValid(_level);
         for (uint256 i; i < _accounts.length; ++i) {
+            if (_accounts[i] == address(0)) revert ZeroAddress();
```
TJ-Everett: This totally works. 
I had envisioned this going to through:
addLevel(_accounts[i], _level); 
like the function above just without the second index - then it would go through all the checks that function has.

TJ-Everett (commented):
Just an implementation question

src/client/application/data/AccessLevels.sol
ShaneDuncan602: Good point. I'll change it.

src/client/application/data/RiskScores.sol
```diff
@@ -54,8 +54,8 @@ contract RiskScores is IRiskScores, DataModule {
     function addRiskScoreToMultipleAccounts(address[] memory _accounts, uint8 _score) external virtual onlyOwner {
         if (_score > MAX_RISK) revert riskScoreOutOfRange(_score);
         for (uint256 i; i < _accounts.length; ++i) {
-            scores[_accounts[i]] = _score;
-            emit AD1467_RiskScoreAdded(_accounts[i], _score);
+            if (_accounts[i] == address(0)) revert ZeroAddress();
+            addScore(_accounts[i], _score);
         }
```
TJ-Everett, line 58: Missed this zero address conditional. We can remove this one.

TJ-Everett (commented):
Missed the conditional in the risk scores

PR #360, opened 2024-04-25
Implement Ways to Handle 4337 Account Abstraction

VoR0220:
I have added some changes to determineTransferAction that will enable us to reasonably assume when a contract is a smart contract account. In addition I have added tests for NFTs for action handling and modified the testing suite in the ERC20s to account for SCAs.


Palmerg4 (changes requested):
Looks good! Maybe just update PROTOCOL-ASSET-HANDLER-DIAMOND.md documenting how correctEntryPointAndHasUserOpValidation is used for SCAs and why there are "different" entryPoint method checks

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -117,6 +149,36 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
```
oscarsernarosero: We need to test the SELL and BUY her with `wallet` instead of `user1`

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -117,6 +149,36 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
...
+        assertEq(uint8(ActionTypes.P2P_TRANSFER), uint8(determineTransferAction({_from: address(wallet), _to: user1, _sender: address(wallet)})));
+
+        // Set up amm for buys and sells
+        DummyAMM amm = initializeERC20AMM(address(applicationCoin), address(applicationCoin2));
+
+        // test Sells 
+        assertEq(uint8(ActionTypes.SELL), uint8(determineTransferAction({_from: user1, _to: address(amm), _sender: address(amm)})));
+
+        // test Buys
+        assertEq(uint8(ActionTypes.BUY), uint8(determineTransferAction({_from: address(amm), _to: user1, _sender: address(amm)})));
```
oscarsernarosero: same here. Instead of `user1` we should plug `wallet`

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -275,12 +337,12 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
...
-        TODO: Fix the determine transfer action function to work with SCA
      */
-    function testERC20_SmartContractWalletBroken_SCA_to_EOA() public {
-        vm.skip(true);
+    function testERC20_SmartContractWallet_SCA_to_EOA() public {
+        address eoa = address(1);
+        vm.startPrank(eoa);
         SBAWallet wallet = new SBAWallet();
-
+        vm.stopPrank();
```
oscarsernarosero: these lines usually make our Necessist checks to fail. If you look closely, you don't need the `vm.startPrank(eoa);` from line 343. If you remove that line, you can get rid of this one as well.

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -289,12 +351,12 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
...
-        TODO: Fix the determine transfer action function to work with SCA
      */
-    function testERC20_SmartContractWalletBroken_SCA_SCA() public {
-        vm.skip(true);
+    function testERC20_SmartContractWallet_SCA_SCA() public {
+        address eoa = address(1);
+        vm.startPrank(eoa);
         SBAWallet wallet = new SBAWallet();
-
+        vm.stopPrank();
```
oscarsernarosero: same here

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -303,16 +365,17 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
...
+    function testERC20_SmartContractWalletWithProtocolSupportedAssets_SCA_to_EOA() public {
+        address eoa = address(1);
+        vm.startPrank(eoa);
         SBAWallet wallet = new SBAWallet();
         // make sure that wallet can hold ETH
         vm.deal(address(wallet), 10 ether);
         applicationCoin.mint(address(wallet), 10 * ATTO);
         applicationCoin.mint(user1, 10 * ATTO);
         
+        vm.stopPrank();
```
oscarsernarosero: same here

test/client/token/ERC20/integration/ApplicationERC20Actions.t.sol
```diff
@@ -323,18 +386,20 @@ contract ApplicationERC20HandlerTest is ERC20Util, HandlerUtils{
...
+        address eoa2 = address(2);
+        vm.prank(eoa2);
         SBAWallet wallet2 = new SBAWallet();
         // make sure that wallet can hold ETH
         vm.deal(address(wallet), 10 ether);
         applicationCoin.mint(address(wallet), 10 * ATTO);
         applicationCoin.mint(user1, 10 * ATTO);
         
-        vm.startPrank(address(wallet));
+        vm.prank(address(wallet));
```
oscarsernarosero: oh nice!

test/client/token/ERC721/integration/ApplicationERC721Actions.t.sol
```diff
@@ -0,0 +1,248 @@
...
+
+    /** Test that actions are properly determined when using protocol supported assets */
+    function testERC721_SmartContractWalletWithProtocolSupportedAssets_Sell() public {
+        SBAWallet wallet = new SBAWallet();
+        applicationCoin.mint(address(wallet), 10 * ATTO);
+        switchToAppAdministrator();
+        applicationNFTv2.safeMint(address(wallet));
+        
+        // Set up amm for buy and sell tests
+        DummyNFTAMM amm = initializeERC721AMM(address(applicationCoin), address(applicationNFTv2));
```
oscarsernarosero, line 162: although this is a valuable test, we are mostly interested in tests against marketplaces instead of AMMs. So, for instance, OpenSea, Blur, etc. The difference being that in a marketplace, usually the flow is different than in a AMM, and it goes something like this:

1. Approve your NFT to the marketplace (usually as operator).
2. List it for a price.
3. A buyer approves the marketplace to take his money (ETH/MATIC or an ERC20).
4. Buyer then calls the "buy" function in the marketplace.
5. Marketplace takes buyer's money (it usually holds it, and then it becomes claimable for the seller. Not relevant here tho.)
6. **Marketplace transfers the NFT from seller to buyer directly.**

This last piece is super important because this means that we have to check 2 actions simultaneously in the NFT case. These being BUY and SELL since a trade happens between 2 people in the same transaction, and not between a person an a contract. Or at least that is the suspicion. That was the intent for the ticket that was merged into this one.

oscarsernarosero (changes requested):
This looks good RJ!. I just found a couple minor stuff, but the ERC721 side is still incomplete since we need to emulate a marketplace and not an NFT AMM. Your logic seems sound and efficient regarding the action identification tho. Nice job.

test/client/token/ERC721/integration/ApplicationERC721Actions.t.sol
VoR0220: See bottom of the testing suite.

src/client/token/handler/common/HandlerUtils.sol
```diff
@@ -29,7 +30,7 @@ contract HandlerUtils{
         } else if(!(_sender == _from)){ 
             action = ActionTypes.SELL;
             emit Action(uint8(ActionTypes.SELL));
-        } else if(isContract(_from)) {
+        } else if(_isContract(_from) && !_isSmartContractAccount(_from)) {
```
oscarsernarosero: I would still put an EOA check here to filter a bit more false positives in the EOA cases:
```
else if(_isContract(_from) && (tx.origin == _to || !_isSmartContractAccount(_from))) {
```
This can also make the code more efficient for EOAs because it would skip the calls to check if _from is an SCA.

test/client/token/ERC721/integration/ApplicationERC721Actions.t.sol
```diff
@@ -0,0 +1,321 @@
...
+        applicationNFTv2.safeTransferFrom(address(wallet), user1, 0);
+    }
+
+    /** Test that actions are properly determined when using protocol supported assets */
+    function testERC721_SmartContractWalletWithProtocolSupportedAssets_Transfer_EOA_to_EOA() public {
+        SBAWallet wallet = new SBAWallet();
+        switchToAppAdministrator();
+        applicationNFTv2.safeMint(address(wallet));
+        applicationNFTv2.safeMint(user1);
+        applicationNFTv2.safeMint(user1);
```
oscarsernarosero: since these lines are becoming repetitive, I would create some global variables and do the repetitive steps in the setUp function.

Palmerg4 (commented):
Looks good, but what is the difference between a SBA and a SCA wallet? The contract is named SBAWallet but the tests are marked as SCA tests. Is there a difference?

test/client/token/ERC721/integration/ApplicationERC721Actions.t.sol
```diff
@@ -0,0 +1,321 @@
...
+        applicationNFTv2.approve(address(marketplace), 0);
+        marketplace.listItem(address(applicationNFTv2), 0, address(applicationCoin), 1 * ATTO);
+        vm.stopPrank();
+
+        vm.startPrank(address(wallet));
+        applicationCoin.approve(address(marketplace), 1 * ATTO);
+        vm.expectEmit(true,false,false,false,applicationCoin.getHandlerAddress());
+        emit Action(uint8(ActionTypes.BUY));
+        vm.expectEmit(true,false,false,false,applicationNFTv2.getHandlerAddress());
+        emit Action(uint8(ActionTypes.SELL));
```
oscarsernarosero, line 303: There is no BUY action being triggered with current implementation. This needs to be taken care of properly.

oscarsernarosero (changes requested):
Nice job at introducing the marketplace tests. I left some minor suggestions and a request to properly handle the marketplace trades since rn the BUY action is never triggered.

VoR0220:
> Looks good, but what is the difference between a SBA and a SCA wallet? The contract is named SBAWallet but the tests are marked as SCA tests. Is there a difference?

Noted, I'm going to rename it to SCA Wallet to alleviate confusion.

src/client/token/handler/common/HandlerUtils.sol
ShaneDuncan602, line 1: We should not be making any implementation changes on this ticket. It is strictly intended to be research and noting change suggestions for future tickets.

oscarsernarosero: I don't agree with this. The research ticket was done already and the ticket itself has the word  "implement" in it.

oscarsernarosero (approved):
Looks good RJ! amazing job

foundry.toml
```diff
@@ -2,6 +2,8 @@
 src = 'src'
 out = 'out'
 libs = ['lib']
+solc_version = "0.8.24"
+
```
ShaneDuncan602: Don't think we want to commit this version.

VoR0220: Are you sure? The last compiler run we were running 0.8.25...

lib/forge-std
ShaneDuncan602, line 1: Was this an intentional update of the submodule?

foundry.toml
ShaneDuncan602: We have always handled this with our local version. It might be something we want to discuss as a team, but I don't think we want to change it here.

lib/forge-std
VoR0220: not intentional, but afaiu it's not upgrading the branch we're using.

foundry.toml
VoR0220: I shall nix it then

lib/forge-std
ShaneDuncan602: I'm not sure. Those submodules don't generally show up unless updated.

VoR0220: Do we want to get rid of it here?

ShaneDuncan602: Did you do a submodule update at some point?

VoR0220: I did not. It feels as though I randomly get the lib/forge-std update as I'm developing throughout other initiatives, possibly due to `foundryup`.

ShaneDuncan602 (approved):
Ready to rock.

Palmerg4 (approved):
Looks great! Nice work

PR #364, opened 2024-05-01
SE-747-Optimize-AppRuleDataFacet-Update-NatSpec-for-addAccountMaxValueByRiskScore-

Palmerg4:
Combined the loops in addAccountMaxValueByRiskScore and addAccountMaxTxValueByRiskScore. 

I tried combining the loops in addAccountMaxValueByAccessLevel and addAccountMaxValueOutByAccessLevel, but quite a few tests were failing after I did so. I assumed its because of the different uint sizes for i in the two loops or because i starts at a different position in the two loops.

I tried type casting the i value to uint8 where needed, and obviously decremented the i value where needed, but still quite a few tests failing. Any ideas?

For context, this is an example of what I tried for addAccountMaxValueByAccessLevel:

        for (uint256 i = 1; i < length; ++i) {
            if (_maxValues[i] < _maxValues[i - 1]) revert WrongArrayOrder();
            data.accountMaxValueByAccessLevelRules[index][uint8(i) - 1] = _maxValues[i - 1];
        }
        
One of the failing tests is testERC721_ApplicationERC721Fuzz_NFTValuationLimit_Complex with error OverMaxValueByAccessLevel()

I wont spend too much more time on this one, probably not even much a gas savings for the access level loops because of the decrements each iteration

### DIFF
--- a/src/protocol/economic/ruleProcessor/AppRuleDataFacet.sol
+++ b/src/protocol/economic/ruleProcessor/AppRuleDataFacet.sol
@@ -33,10 +33,11 @@ contract AppRuleDataFacet is Context, RuleAdministratorOnly, IEconomicEvents, II
         RuleS.AccountMaxValueByAccessLevelS storage data = Storage.accountMaxValueByAccessLevelStorage();
         uint32 index = data.accountMaxValueByAccessLevelIndex;
         if (_maxValues.length != MAX_ACCESSLEVELS) revert BalanceAmountsShouldHave5Levels(uint8(_maxValues.length));
-        for (uint i = 1; i < _maxValues.length; ++i) {
+        uint256 length = _maxValues.length;
+        for (uint256 i = 1; i < length; ++i) {
             if (_maxValues[i] < _maxValues[i - 1]) revert WrongArrayOrder();
         }
-        for (uint8 i; i < _maxValues.length; ++i) {
+        for (uint8 i; i < length; ++i) {
             data.accountMaxValueByAccessLevelRules[index][i] = _maxValues[i];
         }
         ++data.accountMaxValueByAccessLevelIndex;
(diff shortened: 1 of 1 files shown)


TJ-Everett:
> Combined the loops in addAccountMaxValueByRiskScore and addAccountMaxTxValueByRiskScore.
> 
> I tried combining the loops in addAccountMaxValueByAccessLevel and addAccountMaxValueOutByAccessLevel, but quite a few tests were failing after I did so. I assumed its because of the different uint sizes for i in the two loops or because i starts at a different position in the two loops.
> 
> I tried type casting the i value to uint8 where needed, and obviously decremented the i value where needed, but still quite a few tests failing. Any ideas?
> 
> For context, this is an example of what I tried for addAccountMaxValueByAccessLevel:
> 
> ```
>     for (uint256 i = 1; i < length; ++i) {
>         if (_maxValues[i] < _maxValues[i - 1]) revert WrongArrayOrder();
>         data.accountMaxValueByAccessLevelRules[index][uint8(i) - 1] = _maxValues[i - 1];
>     }
> ```
It looks like to me we wouldn't need the -1 in the last line: 
data.accountMaxValueByAccessLevelRules[index][uint8(i)] = _maxValues[i];

This sets the index in the same way the current loop does. The -1 is to ensure the values are in the correct order (ascending values as access levels increase.

TJ-Everett (approved):
This looks good to me! 
I will happily review if the access level looping is solved as well, but for now approved.

Palmerg4:
Right, but the loops start at a different index. The setter loop starts at index 0 and the check loop starts at index 1, that was the thinking behind decrementing the second line index here

ShaneDuncan602:
> testERC721_ApplicationERC721Fuzz_NFTValuationLimit_Complex

I changed the loop to this and testERC721_ApplicationERC721Fuzz_NFTValuationLimit_Complex completes successfully.

```
for (uint i = 1; i < _riskScores.length; ++i) {
            if (_riskScores[i] <= _riskScores[i - 1]) revert WrongArrayOrder();
            if (_maxValue[i] > _maxValue[i - 1]) revert WrongArrayOrder();
}
```

ShaneDuncan602:
> addAccountMaxValueByAccessLevel

I think this should be two separate loops. The validation loop is a much different process than the data loader. Sometimes over-optimization is worse than the status quo.

ShaneDuncan602 (approved):

PR #369, opened 2024-05-03
Pins Foundry version using foundry.lock file

excalq:
Pins a Foundry version (via commit hash), until they provide versions: https://github.com/foundry-rs/foundry/issues/3895
This starts with the known good nightly from 2024.04.18

- **All foundry versions set via foundry.lock**
- **chmod +x on bash scripts**


.github/PRChecklist.md
```diff
@@ -30,9 +30,9 @@
       2. npx prettier --write .
 6. Update the documentation. This is done in 2 simple steps:
 
-   1. Make sure you update foundry every time or it won't work well:
+   1. Make sure `foundry.lock` is set to an appropriate version, then update Foundry.
 
-      `foundryup --version nightly-e0ea59cae26d945445d9cf21fdf22f4a18ac5bb2`
+      `foundryup --commit $(awk '$1~/^[^#]/' foundry.lock)`
```
excalq, line 35: The awk command allows comments in the foundry.lock file.

bcamp136 (approved):

ShaneDuncan602 (approved):
Looks great. Love the use of the foundry.lock

install-deploy-check.sh
```diff
@@ -3,7 +3,7 @@
 SCRIPT_MODE=$1
 
 source ~/.bashrc
-foundryup --version nightly-e0ea59cae26d945445d9cf21fdf22f4a18ac5bb2
+foundryup --commit $(awk '$1~/^[^#]/' foundry.lock)
```
VoR0220, line 7: Because Awk isn't a very natural command for me to understand, can we get a comment on what the awk command is actually doing?

install-forge.sh
```diff
@@ -4,7 +4,7 @@ set -e
 SCRIPT_MODE=$1
 
 source ~/.bashrc
-foundryup --version nightly-e0ea59cae26d945445d9cf21fdf22f4a18ac5bb2
+foundryup --commit $(awk '$1~/^[^#]/' foundry.lock)
```
VoR0220, line 8: same here

install-deploy-check.sh
excalq: For sure, that's fair.

```suggestion
# Pin foundry to a known-good commit hash. Awk ignores comments in `foundry.lock`
foundryup --commit $(awk '$1~/^[^#]/' foundry.lock)
```

PR #370, opened 2024-05-03
Remove registerAMM and registerTreasury functions

TJ-Everett:
Rename RuleBypass to Treasury


src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -171,13 +171,17 @@ contract ProtocolApplicationHandler is
     function _checkAccessLevelRules(address _from, address _to, uint128 _balanceValuation, uint128 _transferValuation, ActionTypes _action) internal {
         uint8 score = appManager.getAccessLevel(_to);
         uint8 fromScore = appManager.getAccessLevel(_from);
-        // Exempting address(0) allows for minting.
-        if (accountDenyForNoAccessLevel[_action].active && !appManager.isRegisteredAMM(_from)  && _from != address(0)) ruleProcessor.checkAccountDenyForNoAccessLevel(fromScore);
-        /// Exempting address(0) allows for burning.
-        if (accountDenyForNoAccessLevel[_action].active && !appManager.isRegisteredAMM(_to) && _to != address(0)) ruleProcessor.checkAccountDenyForNoAccessLevel(score);
```
ShaneDuncan602, line 176: Don't we need to check to see if it's the fee sink here? What would prevent the transfer of fees from triggering the access level error? Or are we assuming the fee sink will be a treasury?

TJ-Everett: I am not sure if I understand why we would check for that only with this rule.

ShaneDuncan602: Just in general. Any of the transfers that happen to load the fee sink will fire the rules. So, if fees are activated, many of the transfers will fail because it will check the fee sink address. This is something I don't think was considered originally.

Palmerg4 (approved):
Looks great! Very thorough, nice work

docs/userGuides/fees/PROTOCOL_FEE_STRUCTURE.md
```diff
@@ -10,6 +10,8 @@ Fees are applied to accounts via general tags in the [AppMananger](../../../src/
 
 Protocol supported tokens will always assess all fees asigned to the account executing the current function. If a token has fees active and an account is tagged with applicable fees or a blank tag is used to assign a default fee, those fees are assessed on token transfers (additive). Token fees are assessed and taken from the token itself, not a collateralized token, when fees are active in the token handler. 
 
+Fees are transferred to fee sink addresses when a fee is created for the token. These addresses will be subject to all active token and application level [rules](../rules/README.md). If a fee sink address is determined to be exempt from rules an [application administrator](../permissions/ADMIN-ROLES.md) can grant the role of Treasury Account to that address. 
+
```
ShaneDuncan602, line 14: I think we need to explain this exemption a little more and make it a recommendation. Can we say something like, "In order for fees to function in conjunction with other enabled rules, it is recommended the fee sink address be granted the role of Treasury Account."

ShaneDuncan602 (approved):
Great job. This was a tough one.

PR #372, opened 2024-05-03
SE-606-GitHub-Action-for-Markdown-Link-Validation

jjbailey-io:
Added a github workflow to check for broken markdown links


jjbailey-io:
Key lessons learned:

1.  This job does not like absolute paths so all references should be relative.
2. References outside of the repository being scanned seems to mark the links as broken so we have to ignore them.
3. The scan is case sensitive.
4. For autogenerated docs, the "blob/<hash>" seems to confuse the check since it's checking the existence of the page without credentials. I suspect if the page was public, we wouldn't need this because it wouldn't need credentials to prove the page exists.

Ok, so here's some specifics about how this is configured:

1. The workflow file is named .github/workflows/markdown.yml. It will scan everything by default. 
3. The config file is .github/workflows/mlc_config.json
   a. We're ignoring external links like infura, loclahost, and the rules-protocol links
   b. Removing "https://github.com/thrackle-io/Tron" references since it's a private repo (removing them just makes the scan check against the current repo state internally instead)
   c. We must remove the "blob/<hash>" on autogenerated docs to ensure it can resolve properly.

excalq (approved):
Great work on this Johnathan! I'm wondering if the "lessons learned" might be useful to capture under `https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/151584777/Github+Wiki`, or perhaps docs about generating Tron documentation?

bcamp136 (approved):

TJ-Everett (approved):
Approved for me.

ShaneDuncan602 (approved):
Looks awesome. This was a great idea.

jjbailey-io:
> Great work on this Johnathan! I'm wondering if the "lessons learned" might be useful to capture under `https://thracklelabs.atlassian.net/wiki/spaces/THRACKLELA/pages/151584777/Github+Wiki`, or perhaps docs about generating Tron documentation?

@excalq  Yeah, I struggled where to put these comments. Based on the lack of support on this github action, I hesitate to directly address it anywhere in our docs.

PR #375, opened 2024-05-06
Remove redundant rule validations in App Handler

TJ-Everett:


src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -368,7 +366,6 @@ contract ProtocolApplicationHandler is
      */
     // slither-disable-next-line calls-loop
     function setAccountMaxValueByRiskScoreIdUpdate(ActionTypes _action, uint32 _ruleId) internal {
-        IRuleProcessor(ruleProcessor).validateAccountMaxValueByRiskScore(createActionTypesArray(_action), _ruleId);
```
Palmerg4, line 374: Is this supposed to be removed? I think this is why it is redundant to have the check before calling this method. Maybe I'm misunderstanding.

TJ-Everett: You are right that this is the call. The other functions have checks prior to calling this one. It looked like this was how we were doing it in the setAccountMaxValueByAccessLevelId and accountDenyForNoAccess so I made the call to remove this and leave the validations in the calling function. 

I could be persuaded to remove the other two and keep the validation in the internal function. 

Am I understanding that correctly?

Palmerg4: Ahh that makes sense. I think either way is fine, no need to keep the validation in the internal function. But it does look like the validations in the calling functions are removed as well. That's where the confusion is coming from

ShaneDuncan602: IMO, the cleanest way to do these validations would be within the internal function. That way, there are no repetitious calls and less likelihood of them in the future. I could see us creating alternate entry points for changing rules. If so, having the validations in the internal function could be advantageous.

TJ-Everett: Good call guys, I think I was even on team internal function last time this popped up. I have all the validation calls dropped into the internal functions now.

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -341,11 +340,8 @@ contract ProtocolApplicationHandler is
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountMaxValueByRiskScoreIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(appManagerAddress) {
-        validateRuleInputFull(_actions, _ruleIds);
```
Palmerg4, line 344: Do we want to get rid of the length checks?

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -502,11 +497,8 @@ contract ProtocolApplicationHandler is
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountMaxValueByAccessLevelIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(appManagerAddress) {
-        validateRuleInputFull(_actions, _ruleIds);
```
Palmerg4, line 505: Here as well, I think we need to keep the length checks here

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -588,11 +581,8 @@ contract ProtocolApplicationHandler is
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountMaxValueOutByAccessLevelIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(appManagerAddress) {
-        validateRuleInputFull(_actions, _ruleIds);
```
Palmerg4, line 591: Same as previous comments, length check

src/client/application/ProtocolApplicationHandler.sol
```diff
@@ -674,11 +665,8 @@ contract ProtocolApplicationHandler is
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountMaxTxValueByRiskScoreIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(appManagerAddress) {
-        validateRuleInputFull(_actions, _ruleIds);
```
Palmerg4, line 677: Here as well

src/client/application/ProtocolApplicationHandler.sol
TJ-Everett: great catch thank you. I think I got over zealous with the trimming!

Palmerg4 (approved):
LGTM!

docs/userGuides/README.md
```diff
@@ -103,15 +103,15 @@ Be sure to [set environmental variables](./deployment/SET-ENVIRONMENT.md) and so
 `scripts/deploy/DeployProtocol.sh`
 This script is responsible for deploying all the protocol contracts. Take into account that no application-specific contracts are deployed here.
 
-#### Deploy Some Test Game Tokens
+#### Deploy Some Test Application Tokens
```
ShaneDuncan602, line 106: Holy cow. Can't believe we still had a game reference in here.

ShaneDuncan602 (approved):
Looks great.

PR #376, opened 2024-05-07
SE-783-Resolve-Audit-finding-Number-5-Fix-renounce-role-for-Treasury-formerly-RuleBypass-

Palmerg4:
Removed specific role renounce functions, added specific role events to renounceRole. I have a sneaking suspicion that I've missed something due to the number of story points for this ticket. The ticket mentioned _This function can have the necessary safeguards to prevent the renouncement of certain roles given certain conditions._ but the only condition these specific role renounce functions had was the checkForAdminMinTokenBalanceCapable for the now treasury account, which has since been removed. Please let me know if there are certain condition checks that are missing in the renounceRole function.

### DIFF
--- a/src/client/application/AppManager.sol
+++ b/src/client/application/AppManager.sol
@@ -119,6 +119,12 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
         /// enforcing the min-1-admin requirement. Only PROPOSED_SUPER_ADMIN_ROLE should be able to bypass this rule
         if (role == SUPER_ADMIN_ROLE) revert BelowMinAdminThreshold();
         AccessControl.renounceRole(role, account);
+
+        if (role == APP_ADMIN_ROLE) emit AD1467_AppAdministrator(_msgSender(), false);
+        if (role == RULE_ADMIN_ROLE) emit AD1467_RuleAdmin(_msgSender(), false);
+        if (role == TREASURY_ACCOUNT) emit AD1467_TreasuryAccount(_msgSender(), false);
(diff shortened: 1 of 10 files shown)


TJ-Everett (approved):
This looks great! Approved for me

src/client/application/AppManager.sol
```diff
@@ -119,6 +119,12 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
         /// enforcing the min-1-admin requirement. Only PROPOSED_SUPER_ADMIN_ROLE should be able to bypass this rule
         if (role == SUPER_ADMIN_ROLE) revert BelowMinAdminThreshold();
         AccessControl.renounceRole(role, account);
+
+        if (role == APP_ADMIN_ROLE) emit AD1467_AppAdministrator(_msgSender(), false);
+        if (role == RULE_ADMIN_ROLE) emit AD1467_RuleAdmin(_msgSender(), false);
+        if (role == TREASURY_ACCOUNT) emit AD1467_TreasuryAccount(_msgSender(), false);
+        if (role == ACCESS_LEVEL_ADMIN_ROLE) emit AD1467_AccessLevelAdmin(_msgSender(), false);
+        if (role == RISK_ADMIN_ROLE) emit AD1467_RiskAdmin(_msgSender(), false);
```
ShaneDuncan602: These seem like they should be changed to one event with an enum for the role type. I'm not sure if it's worth it at this time though. I'm going to pose the question in a group channel.

ShaneDuncan602: If we decide to keep this as is, it would need to be in an if-else block

ShaneDuncan602: It was decided to remove all the custom role events.

ShaneDuncan602 (approved):
Awesome. Looks great!

PR #377, opened 2024-05-07
Fix Test workflow pinning to release

excalq:
The GitHub Action Test Action uses foundry-rs/foundry-toolchain, which only supports setting version[ to a "release"](https://github.com/foundry-rs/foundry-toolchain/blob/master/src/utils.js#L16-L20). So I'm going to use May 2nd's nightly. I won't be surprised if this breaks at the end of the month.


TJ-Everett (approved):
Approved but sad this can't be the commit hash instead. 
NOTE:
Please hold on merging to ensure this resolves the testing issues

PR #378, opened 2024-05-07
Se 792 resolve audit finding non security boolean return statements that can be simplified

TJ-Everett:


Palmerg4 (approved):
Looks great! I like the new events

ShaneDuncan602 (approved):
Looks great.

PR #379, opened 2024-05-08
Se 745 deployment script and documentation for protocol erc 721 u

VoR0220:


script/DeployProtocolERC721U.sol
```diff
@@ -0,0 +1,37 @@
...
+pragma solidity ^0.8.24;
+
+import "forge-std/Script.sol";
+import "src/example/ERC721/upgradeable/ApplicationERC721UProxy.sol";
+import "src/example/ERC721/upgradeable/ApplicationERC721UpgAdminMint.sol";
+import {ApplicationAppManager} from "src/example/application/ApplicationAppManager.sol";
+
+/**
+ * @title The Post Deployment Configuration Step For the Token
+ * @author @VoR0220
```
TJ-Everett: I have been bad about this but we need to have everyone from the Tron team on the Author line :) 
You, Shane, Gordon, and myself moving forward

script/DeployProtocolERC721U.sol
```diff
@@ -0,0 +1,37 @@
...
+ * @notice This is an example script for how to deploy a protocol ERC721 upgradeable token and proxy. This is purely an example and should be modified to fit the needs of the deployment. Other possible implementations one might want to look at include the Openzeppelin Foundry Upgrades library. 
+ */
```
TJ-Everett: After the first sentence here I would remove the other two and replace with: 
** Requires .env variables to be set with correct addresses and Protocol Diamond addresses **
This fits better with the other deploy script commenting pattern we have 

I would also add in the other scripts before and after this one for successful deployment
i.e. we want to then have a pt2 that sets up, initializes and connects the handler 

This should be basically interchangeable with:
example/script/Application_Deploy_04_ApplicationNFT.s.sol --ffi --rpc-url $RPC_URL --broadcast -vvvv

So someone could run example/script/Application_Deploy_04_ApplicationNFTUpgradable.s.sol --ffi --rpc-url $RPC_URL --broadcast -vvvv for example then continue with the other scripts without them breaking the chain

script/DeployProtocolERC721U.sol
```diff
@@ -0,0 +1,37 @@
...
+
+        applicationAppManager = ApplicationAppManager(vm.envAddress("APPLICATION_APP_MANAGER"));
+        ApplicationERC721UpgAdminMint _applicationNFTU = new ApplicationERC721UpgAdminMint();
+        // substitute names that you would want here for name and symbol of NFT and base URI
+        bytes memory callData = abi.encodeWithSelector(_applicationNFTU.initialize.selector, "ERC721U", "ERC721U", address(applicationAppManager), "https://");
+        ApplicationERC721UProxy proxy = new ApplicationERC721UProxy(address(_applicationNFTU), appAdminAddress, callData);
+        vm.stopBroadcast();
+        console.log("Deployed ERC721U proxy at address: ", address(proxy));
+        console.log("Deployed ERC721U at address: ", address(_applicationNFTU));
+    }
```
TJ-Everett: We should also write an export script in parseApplicationDeploy.sh that will export these to the command line for the user. You can copy existing bash commands in there (If I can do it you can do it haha!) that will parse out the _applicationNFTU address 

My thinking is that we will want this to the ApplicationNFT for the remaining scripts not its own variable to make those usable - but we should probably confirm that with everyone at scrum. What I mean by that is that this script should be completely swappable with the other NFT deployment in that it is setting the proxy address as the Application NFT variable in the script series. This also means we would never want to run the NFT and NFTU scripts at the same time to avoid collisions

TJ-Everett (changes requested):
We want this one to be interchangeable with the current NFT deployment script schema.

script/DeployProtocolERC721U.sol
VoR0220: so effectively be able to interchange this with other protocol erc721

script/DeployProtocolERC721U.sol
VoR0220: Agreed.

script/DeployProtocolERC721U.sol
ShaneDuncan602: Those are my thoughts as well. We need to be as full service in the docs as possible.

script/DeployProtocolERC721U.sol
ShaneDuncan602: Great suggestion. We definitely want to reuse all the scripts we can. As long as we are setting the env variable to the proxy address, everything will function like a non upgradeable NFT.

docs/userGuides/deployment/DEPLOY-UPGRADEABLE-TOKEN-AND-HANDLERS.md
```diff
@@ -0,0 +1,60 @@
...
+## Upgradable NFT Deployment
+
+NOTE: 
+- NFT Batch minting and burning is not supported in this release.
+
+- Upgradable Tokens require 3 addresses during the deployment steps. First, the deployment owner that deploys the app manager and sets roles. Second, an Application Administrator address to deploy the token, token proxy and initialize the token. Third, a Configuration Application Administrator address to connect the token with the handler address once deployed and initilaized. The token and the handler contracts require initialize functions called and are controlled through the protocol RBAC roles.
+
+1. Ensure the [environment variables][environment-url] are set correctly. The `APPLICATION_APP_MANAGER`, `CONFIG_APP_ADMIN` and `DEPLOYMENT_OWNER_KEY` are used by the deployment script.  
+
+2. Inside the [ApplicationNFT script](../../../script/clientScripts/Application_Deploy_04_ApplicationNFTUpgradeable.s.sol) change the `DAWGUniversity` name, `DAWGU` symbol and `https://dawgieboisuniversity` inside the script to the desired name, symbol and token URI for your NFT deployment.
```
ShaneDuncan602: I think we need to stick with our common example theme of public domain classic monsters, Frankenstein, Dracula, etc.

script/clientScripts/Application_Deploy_04_ApplicationNFTUpgradeablePt2.s.sol
```diff
@@ -0,0 +1,66 @@
...
+import {ApplicationERC721AdminOrOwnerMint} from "src/example/ERC721/ApplicationERC721AdminOrOwnerMint.sol";
+import "src/example/ERC721/upgradeable/ApplicationERC721UProxy.sol";
+import {ApplicationAppManager} from "src/example/application/ApplicationAppManager.sol";
+import "./DeployBase.s.sol";
+
+/**
+ * @title Application Deploy 04 Application Non-Fungible Token Upgradeable Script
+ * @dev This script will deploy an ERC721 non-fungible token and Handler.
+ * @notice Deploys an application ERC721 non-fungible token and Handler..
+ * ** Requires .env variables to be set with correct addresses and Protocol Diamond addresses **
```
ShaneDuncan602, line 16: need an author line in here

docs/userGuides/deployment/DEPLOY-UPGRADEABLE-TOKEN-AND-HANDLERS.md
VoR0220: Is that what it is? Let me go find a monster then.

ShaneDuncan602 (approved):
Great job!

docs/userGuides/deployment/DEPLOY-UPGRADEABLE-TOKEN-AND-HANDLERS.md
```diff
@@ -0,0 +1,60 @@
...
+        ````
+4. Use the output from the deployment to set relevant environment variables:
+        ````
+        bash  script/ParseApplicationDeploy.sh 3 && source .env
+        ````
+5. Configure the contracts:
+        ````
+        forge script script/clientScripts/Application_Deploy_04_ApplicationNFTUpgradeablePt2.s.sol --ffi --broadcast
+        ````
+
```
TJ-Everett, line 49: We should put the new sh script right here as well sh script/SubUpgradeableTokenForRegularToken.sh 
That way if someone is here going doc by doc for deployment it will be more smooth for them.

TJ-Everett (approved):
One tiny thing in the docs otherwise approved for me!

docs/userGuides/deployment/DEPLOY-UPGRADEABLE-TOKEN-AND-HANDLERS.md
VoR0220: Good point.

PR #380, opened 2024-05-08
SE-784-Resolve-Audit-finding-Number-6-Prune-state-data-when-rule-is-deactivated-or-updated

ShaneDuncan602:


src/client/token/handler/ruleContracts/HandlerAccountMaxTradeSize.sol
```diff
@@ -65,6 +67,29 @@ contract HandlerAccountMaxTradeSize is RuleAdministratorOnly, ActionTypesArray,
...
+        clearAccountMaxTradeSizeAccumulators();
+        delete data.accountMaxTradeSize[_action];
+    }
+
+    /**
+     * @dev Clear the rule data accumulators
+     */
+    function clearAccountMaxTradeSizeAccumulators() internal {
+        AccountMaxTradeSizeS storage data = lib.accountMaxTradeSizeStorage();
+        for (uint i; i < data.transactors.length; ++i) {
```
VoR0220: This has me concerned. This could be a very large list of transactors. Are we sure we won't effectively be bricked here?

src/client/token/handler/ruleContracts/HandlerTokenMinHoldTime.sol
```diff
@@ -73,6 +86,19 @@ contract HandlerTokenMinHoldTime is RuleAdministratorOnly, ITokenHandlerEvents,
...
+     * @dev Clear the rule data structure accumulators
+     */
+    function clearTokenMinHoldTimeAccumulators() internal {
+        TokenMinHoldTimeS storage data = lib.tokenMinHoldTimeStorage();
+        for (uint i; i < data.tokenIds.length;) {
+            data.ownershipStart[data.tokenIds[i]] = 0;
+            unchecked {
+                ++i;
+            }
+        }
```
VoR0220: perhaps something more gas efficient might be to make a note of when the tokenMinHoldTime rule was activated, and then add a conditional to check if the ownershipStart field of the token was before the ruleActivation period and if so to delete it, or if the rule is not active, to delete the tokenId field on any action.

test/client/token/ERC20/integration/ERC20CommonTests.t.sol
```diff
@@ -1319,6 +1355,70 @@ abstract contract ERC20CommonTests is TestCommonFoundry, DummyAMM, ERC20Util {
...
+        // This one will pass so we know it is good
+        vm.startPrank(user1);
+        amm.dummyTrade(address(testCaseToken), address(applicationCoin2), 500, 500, true);
+        /// This one violates the rule
+        vm.expectRevert(0x523976c2);
+        amm.dummyTrade(address(testCaseToken), address(applicationCoin2), 500, 500, true);
+        /// Now, we deactivate and reactivate the rule and try again
+        switchToRuleAdmin();
+        TradingRuleFacet(address(applicationCoinHandler)).activateAccountMaxTradeSize(createActionTypeArray(ActionTypes.SELL), false);
+        TradingRuleFacet(address(applicationCoinHandler)).activateAccountMaxTradeSize(createActionTypeArray(ActionTypes.SELL), true);
```
VoR0220, line 1377: I think we need to replicate this test but with a lot more accounts in the mix. Try to expand the transactors field and see if we run into issues where the gas field is constrained to what we would normally expect from a regular transaction in a live environment.

src/client/token/handler/ruleContracts/HandlerTokenMaxDailyTrades.sol
```diff
@@ -49,16 +50,28 @@ contract HandlerTokenMaxDailyTrades is RuleAdministratorOnly, ActionTypesArray,
...
         }
     }
 
+    /**
+     * @dev Clear the rule data accumulators
+     */
+    function clearTokenMaxDailyTradesAccumulators(uint32 _ruleId) internal {
+
+        TokenMaxDailyTradesS storage data = lib.tokenMaxDailyTradesStorage();
+        for (uint i; i < data.tokenIdsByRuleId[_ruleId].length; ++i) {
```
VoR0220: Similar concerns here. If you have 10k token Ids, this quickly falls into bricking territory. You'd need to either have a way to paginate or find a way thats a bit more passive in how it prunes itself.

VoR0220 (changes requested):
To summarize: I have concerns that the current implementation is brickable and leaves you unable to activate a rule if you have too many registered token IDs in various places in the rules accumulator clearers. If you wouldn't mind putting together a test and/or an explanation for why I'm wrong to think this, I would be appreciative.

TJ-Everett (approved):
Great work on this one Shane. Approved for me.

VoR0220 (approved):
Much better. Approved.

Palmerg4 (approved):
Looks great!

PR #382, opened 2024-05-09
1/3: Foundry install/update script

excalq:
Each deploy script used to run `foundryup`, so composite scripts like install-deploy-check caused Foundry to compile 3 times! This refactor will greatly speed developer usage and reduce GHA spend.


TJ-Everett (approved):
This looks great from the Tron workflow side of things. We don't use those .sh scripts in the day to day or deployment testing. So no concerns from me. (I double checked this too)

bcamp136 (approved):

PR #383, opened 2024-05-09
2/3: Refactor install-deploy-check

excalq:
PR Note: Once #382 is merged, change this PR's base to `main`.

## Refactor of `install-deploy-check`

- Capture stdout and stderr in `[install-deploy-check.sh](http://install-deploy-check.sh/)`, only print anything on failures. (Successes should be silent)

- Use standardized exit codes (0: success, -1: failure)

- Separate it from Universus K8s deployment steps. That workflow can run this as a pre-step.


bfcoursewool (approved):
Love it.

PR #385, opened 2024-05-10
SE-767-Update-token-handlers-to-emit-version-in-event-when-upgraded

VoR0220:


src/protocol/diamond/VersionFacet.sol
```diff
@@ -18,15 +18,15 @@ contract VersionFacet is ERC173 {
     * versioning format: --> "MAJOR.MINOR.PATCH".
     */
     function updateVersion(string memory newVersion) external onlyOwner{
-        lib.versionStorage().version = newVersion;
+        lib.versionStorage().VERSION = newVersion;
```
VoR0220: I did a git restore and for some reason it restored to this. Maybe due to an outdated remote. Adding a note to fix this.

src/client/token/handler/diamond/HandlerVersionFacet.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import {StorageLib as lib} from "./StorageLib.sol";
+import {IHandlerDiamondEvents} from "../../../../common/IEvents.sol";
+import "../common/AppAdministratorOrOwnerOnlyDiamondVersion.sol";
+/**
+ * @title Protocol Version Facet 
+ * @author @ShaneDuncan602, @oscarsernarosero, @TJ-Everett, @VoR0220, @GordonPalmer
+ * @notice This is a facet that should be deployed for any protocol diamond.
+ * @dev setter and getter functions for Version of a diamond.
+ */
```
VoR0220, line 12: Adding a note to fix the docs here so that they define the HandlerVersionFacet.

test/client/application/ProtocolApplicationHandlerTests.t.sol
```diff
@@ -13,6 +14,23 @@ contract ProtocolApplicationHandlerTests is TestCommonFoundry {
...
 
+    function test_versionUpgradeFacet() public {
+        bool success;
+
+        switchToAppAdministrator();
+
+        vm.expectEmit();
+        emit AD1467_UpgradedToVersion(appAdministrator, "1.2.0");
+        (success, ) = address(applicationCoinHandler).call(abi.encodeWithSignature("updateVersion(string)", "1.2.0"));
+        assertTrue(success);
```
Palmerg4, line 25: Maybe just add another assert for each call checking that the version is now 1.2.0, or is this test just for the event emission?

VoR0220: was purely for event emission but I can add this as well.

TJ-Everett (approved):
Looks good to me!

Palmerg4 (approved):
Yep, LGTM

PR #387, opened 2024-05-14
SE-748-Change-Atomic-Rule-Set-functions-in-token-handlers-to-accept-ActionTypes-Array-length-of-0

Palmerg4:
Removed check for 0 length actions array input argument from all atomic full rule set functions. Added tests for each.

### DIFF
--- a/src/client/token/handler/ruleContracts/HandlerAccountApproveDenyOracle.sol
+++ b/src/client/token/handler/ruleContracts/HandlerAccountApproveDenyOracle.sol
@@ -37,7 +37,6 @@ contract HandlerAccountApproveDenyOracle is RuleAdministratorOnly, ActionTypesAr
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountApproveDenyOracleIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(lib.handlerBaseStorage().appManager) {
-        if (_actions.length == 0) revert InputArraysSizesNotValid();
         if (_actions.length != _ruleIds.length) revert InputArraysMustHaveSameLength();
         clearAccountApproveDenyOracle();
         for (uint i; i < _actions.length; ++i) {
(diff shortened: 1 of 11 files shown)


src/client/token/handler/ruleContracts/HandlerAccountApproveDenyOracle.sol
```diff
@@ -37,7 +37,6 @@ contract HandlerAccountApproveDenyOracle is RuleAdministratorOnly, ActionTypesAr
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountApproveDenyOracleIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(lib.handlerBaseStorage().appManager) {
-        if (_actions.length == 0) revert InputArraysSizesNotValid();
         if (_actions.length != _ruleIds.length) revert InputArraysMustHaveSameLength();
         clearAccountApproveDenyOracle();
```
VoR0220, line 42: Is this the portion doing the clearing in lieu of the cancelling of the zero length arrays?

src/client/token/handler/ruleContracts/HandlerAccountMaxTradeSize.sol
```diff
@@ -43,7 +43,6 @@ contract HandlerAccountMaxTradeSize is RuleAdministratorOnly, ActionTypesArray,
      * @param _ruleIds Rule Id corresponding to the actions
      */
     function setAccountMaxTradeSizeIdFull(ActionTypes[] calldata _actions, uint32[] calldata _ruleIds) external ruleAdministratorOnly(lib.handlerBaseStorage().appManager) {
-        if (_actions.length == 0) revert InputArraysSizesNotValid();
         if (_actions.length != _ruleIds.length) revert InputArraysMustHaveSameLength();
         clearAccountMaxTradeSize();
```
VoR0220, line 48: Same question here

src/client/token/handler/ruleContracts/HandlerAccountApproveDenyOracle.sol
Palmerg4: These clearing internals were always present. DOOM just needs a way to call this internal by passing zero length arrays so they can clear rules as needed without replacing them

TJ-Everett (approved):
Looks good to me. Just a suggestion to add comments in the natspec block for why we do not check for 0 length in these for future audits

VoR0220 (approved):
GTM

ShaneDuncan602 (approved):
Looks fantastic.

PR #388, opened 2024-05-15
2024.05.03 stop using foundry base

bfcoursewool:
Reworking all the docker containers into a single multi-stage dockerfile, and then also reworking the GH workflows to use the new container setups. There are still two workflows (tests.yml and slither.yml) which could benefit from re-work to use the new dockerfiles, but I didn't have a chance to get to those yet. They're still installing the foundry toolchain in GHA via the foundry-rs repos, so I think it's likely that they're still vulnerable to the same kinds of instability that we were experiencing in the rest of our stack. I think it would be pretty easy to base them off of our new "foundry-base" docker layer and just go from there with them. 

One outstanding issue in this PR that I haven't been able to quite pin down / fix, is getting the "compile-tron" docker layer to properly cache in GH Actions workflows... there should be no reason why it doesn't cache, and yet it recompiles tron every time. So any thoughts on that are totally welcome.


TJ-Everett (approved):
This is awesome work Rob. I have it up and running locally and looks like it built just fine.

bcamp136 (approved):

bcamp136:
Great work Rob!

Dockerfile
```diff
@@ -1,10 +1,129 @@
...
+# anything that needs to use forge/anvil/cast/etc. Since foundry's release process is so prone
+# to breaking things, we have opted to directly compile foundry from github using cargo, pinned to
+# a specific commit hash in their repo, thus avoiding all the issues we've had with their release
+# process and wacky install scripts.
+#
+# This layer is cached and not re-built unless there is some change to foundry.lock. It may be 
+# necessary to specifically update your Docker settings to increase the max available memory in
+# a container for this base stage to successfully build. 
+#
+################################################
```
excalq, line 13: Compliments on the thorough image stage documentation. This Dockerfile is the heart of all deployments so that's great to have here.

docker-scripts/deploy-tron.sh
```diff
@@ -0,0 +1,86 @@
...
+	forge script script/clientScripts/Application_Deploy_02_ApplicationFT1Pt2.s.sol --ffi --broadcast
+	forge script script/clientScripts/Application_Deploy_04_ApplicationNFT.s.sol --ffi --broadcast
+	source script/ParseApplicationDeploy.sh 3
+	forge script script/clientScripts/Application_Deploy_04_ApplicationNFTPt2.s.sol --ffi --broadcast
+	forge script script/clientScripts/Application_Deploy_05_Oracle.s.sol --ffi --broadcast 
+	source script/ParseApplicationDeploy.sh 4
+	forge script script/clientScripts/Application_Deploy_06_Pricing.s.sol --ffi --broadcast
+	source script/ParseApplicationDeploy.sh 5
+	forge script script/clientScripts/Application_Deploy_07_ApplicationAdminRoles.s.sol --ffi --broadcast
+} > /dev/null
```
bcamp136, line 44: I just realized that this is also hiding all of the script output when we launch the full stack locally. While it's not the end of the world, it is nice to be able to pull up these logs to troubleshoot potential issues in the tron deploy.

Dockerfile
```diff
@@ -1,10 +1,129 @@
...
+# RUST_LOG configures anvil output information. I think. 
+# CHAIN_ID is the chain-id Anvil runs on
+#
+################################################
+
+FROM compile-tron as anvil
+ENV FOUNDRY_PROFILE=docker
+ENV RUST_LOG=backend,api,node,rpc=warn
+ENV CHAIN_ID=31337
+ENTRYPOINT anvil --host 0.0.0.0 --chain-id $CHAIN_ID
```
excalq, line 64: Note that `ENTRYPOINT` runs `anvil` directly, without a shell. It also doesn't allow `docker run -it anvil <some-command>` to behave as expected if ENTRYPOINT were the default `/bin/sh -c`. `CMD` actually passes arguments to `ENTRYPOINT`.

See https://spacelift.io/blog/docker-entrypoint-vs-cmd

docker-scripts/deploy-tron.sh
```diff
@@ -0,0 +1,86 @@
...
+
+##
+# the --with-deploy-check flag is passed into this script by the main k8s.yml GH workflow,
+# so that it can confirm that the build will successfully deploy. Because that happens in GHA
+# in that case we override the FOUNDRY_PROFILE setting and start our own anvil service here
+# for it to deploy to and confirm success, and then GHA will just delete the whole test image
+# and push the tron and anvil builds to ECR. 
+##
+if [ $SCRIPT_MODE = "--with-deploy-check" ]; then
+	FOUNDRY_PROFILE=local
```
excalq: nit: indentation

docker-scripts/deploy-tron.sh
```diff
@@ -0,0 +1,86 @@
...
+    exit 1
+  else
+    echo "✅ All tests passed."
+    echo -e "\n\n!!! TODO in PR 3/3 !!!: GitHub Actions 'Deploy' workflow will run this in a step. This script will no longer invoke a deployment."
+  fi
+else
+  # For a normal deploy, keep the tron container running so that tron devs can exec 
+  # into it and run forge tests and other forge/cast/etc commands against a populated 
+  # .env file
+  tail -f /dev/null
```
excalq: I'm not a fan of this approach as we end up with a needlessly persistent service in K8s. The deployment is a `Job` in Doom, so it'd appear to be an non-completing job, which is a bit odd. We can merge this for now, but I think there's a better approach for local dev once we have a tool like Devspace in use.

excalq: Moreover, this script has two divergent purposes. One is to deploy, the other (which consumes most LOC) is the deploy-check. Perhaps we should just have a separate deploy check script which runs `deploy-tron.sh` before doing its checks?

.github/workflows/k8s.yml
```diff
@@ -29,18 +30,69 @@ jobs:
...
+        with:
+          context: .
+          file: ./Dockerfile
+          target: check-deploy
+          load: true
+          tags: ${{ env.TRON_TEST_TAG }}
+          cache-from: type=registry,ref=${{ env.TRON_BASE_TAG }}
+          cache-to: type=inline
+
+      - name: Test Tron Deployment
```
excalq:
```suggestion
      - name: Clean up deploy-check image 
```

.github/workflows/k8s.yml
excalq, line 1: Maybe we should rename this file (and the GH Action), as it doesn't do anything with k8s, other than push the ECR images which the DOOM stack deploys from.

docker-compose.yml
```diff
@@ -1,22 +1,17 @@
...
+  deploy:
     depends_on:
       - anvil
     build:
       context: .
-      dockerfile: Dockerfile.deploymentCheck
-    environment:
-      FOUNDRY_PROFILE: docker
+      dockerfile: Dockerfile
+      target: deploy-tron
```
excalq, line 17:
```suggestion
      target: deploy-tron
    # Keep the container running after successful deploy for local developer usage
    entrypoint:
      - /bin/sh
      - -c
    command:
      - ./docker-scripts/deploy-tron.sh --with-deploy-check && /usr/bin/tail -f /dev/null'
```

excalq:
@bfcoursewool Summarizing my thoughts here:

1. Thank you a ton for all the work on this and the DOOM PR - that's a lot of refactoring and removing old cruft! 
2. Awesome that Tron's `k8s/` directory is cleaned out, except for the one `k8s/tron/necessist-job.yaml` which isn't a DOOM concern. 
3. The deploy-tron.sh has the obvious purpose of deploying Tron. But most of its code is for special case "deploy-check". That should be separate, and run `deploy-tron.sh` as a dependency. This is of course complicated by the next item:
4. `deploy-tron` never terminates (except for failures). If this is useful only to local devs, lets achieve this by overriding `command` in `docker-file.yaml`.

docker-scripts/deploy-tron.sh
bfcoursewool: Yeah I thought about this too... It's definitely way better to have it be totally silent in the "deploy check" context, but then in the real deploy context, yeah, the outputs are pretty nice. I guess it shouldn't be too hard to make it conditionally do that...

Dockerfile
bfcoursewool: Nice, ok thanks for that... I haven't necessarily been super clear on the pros/cons/diffs between CMD and ENTRYPOINT, so I think it probably got set that way a while back and then I just kept it like it was.

bcamp136 (approved):

PR #391, opened 2024-05-21
Foundry version bump and small changes to docker scripts

bfcoursewool:
This is mostly just a couple of the very last changes I had intended to make on the new dockerfile setup that never got pushed before, and a foundry version bump.


bcamp136 (approved):

excalq (approved):
Great work!

PR #393, opened 2024-05-21
Tron Build/Push happens on merge to main

excalq:
Restores old behavior after refactor to deploy-check process (now a job step in this workflow.)


TJ-Everett (approved):
Looks good to me

bcamp136 (approved):

PR #394, opened 2024-05-22
adding nft upgradeable script calls in here

bfcoursewool:
Just waht it says


TJ-Everett (approved):
b e a utiful. Approved

PR #398, opened 2024-06-04
SE-632-Upgrade-protocol-and-handlers-on-testnet

ShaneDuncan602:


script/python/get_latest_facet_address.py
```diff
@@ -43,6 +43,10 @@ def main():
         print("Not Found")
         return
     enc = encode(["address[2]"], [[facet, diamond]])
+    # enc = encode(["(address,address)"], [(facet, diamond)])
+    # enc = encode(["address"], [facet])
+    # print(enc)
+    # print("0x01462DcB0DE6ea6b9Aee343C6c1280A17259969a","0x826149518A0D4Cf67b58fB8698b478CeaE96708E")
```
Palmerg4: Are these comments permanent?

docs/userGuides/rules/README.md
```diff
@@ -32,23 +32,6 @@ Rule application is a two step process. First, the rule admin must obtain a rule
...
-| [Token Minimum Hold Time](./TOKEN-MIN-HOLD-TIME.md) | The purpose of the token-min-hold-time rule is to reduce trade volitility by preventing transfers of tokens for a number of hours after ownership is acquired, either via minting or transfers. This rule allows developers to specifiy a number of hours, up to 43830 (5 years), that each tokenId must be held for. |
-| [Token Minimum Transaction](./TOKEN-MIN-TRANSACTION-SIZE.md)| The purpose of the token-min-tx-size rule is to prevent micro-trades or dust trades within an ecosystem. A developer can set the minimum number of tokens required per transfer to prevent these types of trades. |
```
Palmerg4, line 51: Are these definitions going to be relocated or purged? Do they need to be updated?

script/python/get_latest_facet_address.py
ShaneDuncan602: Good catch. These should be deleted.

docs/userGuides/rules/README.md
ShaneDuncan602: These were duplicates that already exist at the top of the README. Somehow they must've been copy/pasted twice so I deleted the redundant section.

PR #399, opened 2024-06-04
SE-808-Create-Reusable-Tests-for-individual-rule-gas-usage

Palmerg4:
This contains the reusable tests for individual gas usage. I wanted to get some feedback before I make the doc for this. Currently the gas reports provided by foundry arent exactly what we need for this, so I decided to use the gasLeft() built in. I wanted to get some feedback on readability, as I believe the test outputs aren't exactly the most readable gas reports.

As I mentioned, the doc isnt created for this just yet so Ill include the command here to run this report. Pretty normal --match-path command.

`forge test --ffi -vv --match-path test/util/GasReport/GasReport.t.sol`


test/util/GasReport/GasHelpers.sol
```diff
@@ -0,0 +1,25 @@
...
+    function stopMeasuringGas() internal virtual returns(uint256){
+        uint256 checkpointGasLeft2 = gasleft();
+
+        // Subtract 100 to account for the warm SLOAD in startMeasuringGas.
+        uint256 gasDelta = checkpointGasLeft - checkpointGasLeft2 - 100;
+
+        emit Gas_Log(string(abi.encodePacked(checkpointLabel, " Gas")), gasDelta);
+        return gasDelta;
+    }
+}
```
VoR0220: Nice

test/util/GasReport/GasReport.t.sol
```diff
@@ -0,0 +1,815 @@
...
+    }
+
+    function testERC20_TokenMinTxSize_Transfer() public endWithStopPrank {
+        _applyMinTxSizeSetUp();
+        _erc20TransferGasReport("ERC20_TokenMinTxSize_Transfer");
+    }
+
+
+/**********  Rule Setup Helpers **********/
+    function _erc20MintGasReport(string memory _label) public {
```
VoR0220: So I'm seeing this come up a lot as a reusable piece but it's not clear to me that there's any differentiation. It's just testing whether something has been minted. Why are we applying different labels to the same thing being tested?

test/util/GasReport/GasReport.t.sol
```diff
@@ -0,0 +1,815 @@
...
+
+    function _applyAccountDenyForNoAccessLevelSetUp() public {
+        ActionTypes[] memory actionTypes = createActionTypeArray(ActionTypes.P2P_TRANSFER, ActionTypes.BUY, ActionTypes.SELL, ActionTypes.MINT);
+        createAccountDenyForNoAccessLevelRuleFull(actionTypes);
+        switchToAccessLevelAdmin();
+        applicationAppManager.addAccessLevel(appAdministrator, 1);
+        applicationAppManager.addAccessLevel(user1, 1);
+    }
+
+    function _applyAccountMaxTradeSizeSetUp(address _handler) public {
```
VoR0220: shouldn't the gas usage begin as a modifier prior to execution and be applied on all these specific actions?

test/util/GasReport/GasReport.t.sol
TJ-Everett: This looks like the string for the output of the tests to say what rule set up was run 

Logs:
  ERC721_TokenMaxBuySellVolume_TransferFrom 93031
  ERC721_TokenMaxBuySellVolume_Burn 100136
  ERC721_TokenMaxBuySellVolume_SafeMint 141367

test/util/GasReport/GasReport.t.sol
TJ-Everett: I don't think so RJ, then we wouldn't get the user picture of how much gas is used for the transfers/mints/burns since that is more what we care about. The cost to create and set the rule is not part of the "Gas cost" that we test for, or at least haven't cared about prior

TJ-Everett (approved):
Looks good to me Gordon. I love the output when run with -vvv

test/util/GasReport/GasHelpers.sol
ShaneDuncan602, line 1: This directory needs to be camelCase instead of PascalCase.

ShaneDuncan602 (approved):
Just need the directory in the proper case. Great job on this. Very usable.

VoR0220 (approved):
Sorry about that! My misunderstanding. Now I'm able to look through it with a fresher eye and I see how this is supposed to be set up. Nice work. Approved!

docs/userGuides/gasReports/gasReportUsage.md
```diff
@@ -0,0 +1,80 @@
+# Individual Rule Gas Comparison ERC721 and ERC20
```
ShaneDuncan602: Can you add the standard header to this?

docs/userGuides/gasReports/gasReportUsage.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Purpose
+
+The gas report is used to determine how individual active rules effect gas consumption of basic token actions.
```
ShaneDuncan602: Wouldn't this be "affect"? I'm asking but I am not sure.

Palmerg4: Youre correct, it should be affect

docs/userGuides/gasReports/gasReportUsage.md
Palmerg4: The standard header being the project version and image, correct?

docs/userGuides/gasReports/gasReportUsage.md
VoR0220: It should be affect. Very common mistake to misuse so no biggie and I doubt anyone would have cared but just so we're clear.

ShaneDuncan602 (approved):
Good job!

PR #401, opened 2024-06-07
Completed Makefile functionality

mpetersoCode55:


script/makefile-tools/convertStringToByte32.sh
```diff
@@ -0,0 +1,29 @@
+  #!/bin/bash
```
excalq, line 1: Why not require a single arg `INPUT=${1:-'Test'}`, rather than using `-s ARG`?

excalq: (Which would remove the need for lines 5-22).

excalq (approved):
👍 Good stuff! As noted, I think the shell script's arg handling could be simplified.

script/makefile-tools/convertStringToByte32.sh
mpetersoCode55: Good catch! Implemented

Makefile
```diff
@@ -0,0 +1,77 @@
...
+# (-include to ignore error if it does not exist)
+-include .env
+
+###############################################
+# Cast Sends - Setters
+###############################################
+
+# Create a Account Min Max Token Balance rule. To use a specific string for the tag use the following syntax:
+# make setRule INPUT=tag
+# otherwise if INOUT is not specified, "Test" will be used for the tag.
```
ShaneDuncan602: INOUT

Makefile
```diff
@@ -0,0 +1,77 @@
...
+# Output is parsed and the created rule id is given as output to the make call
+createRule: createRuleCastSend
+	@echo $(shell sh ./script/makefile-tools/parseRuleID.sh)
```
ShaneDuncan602, line 14: This is great. Extremely helpful for clients because wading through that output sucks.

ShaneDuncan602 (approved):
This is great. I plan to use this often.

TJ-Everett (approved):
Approved for me. This is great work!

PR #404, opened 2024-06-13
Se 817 create a testnet token deployment of an lp in the amm with the token

mpetersoCode55:


script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -20,21 +20,42 @@ echo Please enter RPC URL
...
+    forge script script/clientScripts/DeployERC20Handler.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+else 
+    forge script script/clientScripts/Application_Deploy_02_ApplicationFT1.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+    sh script/ParseApplicationDeploy.sh 2 --chainid $CHAIN_ID
+    forge script script/clientScripts/Application_Deploy_02_ApplicationFT1Pt2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+fi
+
+forge script script/clientScripts/Application_Deploy_03_ApplicationFT2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+sh script/ParseApplicationDeploy.sh 6 --chainid $CHAIN_ID
+forge script script/clientScripts/Application_Deploy_03_ApplicationFT2Pt2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
```
ShaneDuncan602: So this is expecting the deployed ERC20 to not already be protocol supported. I have a feeling that people will likely set up their token, set rules, and then connect to the AMM. I think this probably needs to handle that. Can we add another conditional prompt once they select "y" for "Do you already have an ERC20 deployed (y or n)?" that allows them to choose whether it is protocol supported already?

ShaneDuncan602 (changes requested):
Great job. I just have one suggestion to handle protocol supported tokens that are already deployed.

script/clientScripts/Application_Deploy_Example.sh
mpetersoCode55: Good point! Implemented

ShaneDuncan602 (approved):
Love it.

PR #411, opened 2024-06-21
Invariant failure investigation

oscarsernarosero:
It looks like Foundry doesn't allow anymore uncaught reversions in the invariant tests. The failing invariant tests were fixed.


Palmerg4 (approved):
LGTM!

PR #416, opened 2024-07-11
Se 956 remove example token app manager references

ShaneDuncan602:


src/client/token/handler/diamond/ERC721HandlerMainFacet.sol
```diff
@@ -113,4 +113,12 @@ contract ERC721HandlerMainFacet is HandlerBase, HandlerUtils, ICommonApplication
 
         return true;
     }
+
```
TJ-Everett, line 116: Thank you for this!

TJ-Everett (approved):
Approved. Looks great

PR #417, opened 2024-07-15
Rename Tron -> Aquifi-rules-v1

excalq:
# Repository Naming Changes 

- Update mlc_config.json


bfcoursewool (approved):

.github/workflows/mlc_config.json
```diff
@@ -19,6 +19,10 @@
       "pattern": "https://github.com/thrackle-io/Tron",
```
jjbailey-io: This reference should also be updated.

jjbailey-io (changes requested):

.github/workflows/slither.yml
```diff
@@ -83,7 +86,8 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
+          # May 2, 2024 - Must be a release (nightly) tag
```
VoR0220, line 89: Why are we defining the nightly twice?

VoR0220 (approved):
Generally accomplishes the goal of renaming things. Just a question about the CI.

.github/workflows/slither.yml
jjbailey-io: Because they're two separate workflows that are being executed and both are installing foundry. By omitting the version in the second, it was not using the same version.

PR #418, opened 2024-07-23
Se 913 use tron versioning when syncing up ab is in doom

bfcoursewool:
Just updating the abi-aggregator to use the version it finds in package.json when storing the abis so that the admin-ui CI process which runs this script can make sure it has the most up to date ABI versions in the right spot.


bcamp136 (approved):

VoR0220 (approved):

bcamp136:
These changes have been applied in another branch making this PR unnecessary.

PR #419, opened 2024-07-23
Added coverage reporting

jjbailey-io:
- Added code coverage reporting
- Updated tests to use faster runners
- Updated workflows to use a consistent version of foundry

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,6 +16,15 @@ jobs:
     name: Code coverage evaluation
     runs-on: ubuntu-latest-m-16
     steps:
+      - name: Increase swapfile
+        # https://github.com/actions/runner-images/discussions/7188#discussioncomment-6750749
+        run: |
+          sudo swapoff -a
+          sudo fallocate -l 15G /swapfile
+          sudo chmod 600 /swapfile
+          sudo mkswap /swapfile
+          sudo swapon /swapfile
+          sudo swapon --show
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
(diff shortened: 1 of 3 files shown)


VoR0220 (approved):

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,82 @@
...
+          forge install dapphub/ds-test 
+          forge install OpenZeppelin/openzeppelin-contracts
+      - name: Build the repository
+        shell: bash
+        run: |
+          forge clean && forge build --optimize
+      - name: Run tests with coverage option
+        env:
+          ETHEREUM_RPC_KEY: ${{ secrets.ALCHEMY_KEY }}
+          POLYGON_RPC_KEY: ${{ secrets.POLYGON_ALCHEMY_KEY }}
```
TJ-Everett: Are these test coverage reports utilizing the alchemy key at all? Seems like this wouldn't be ideal if they are as that would be increasing usage on the alchemy account?

TJ-Everett (approved):
Approved. Looks great, just a question about the Alchemy keys referenced

.github/workflows/coverage.yml
VoR0220: Unless we were running tests off of a forked alchemy environment.

Palmerg4 (approved):

.github/workflows/coverage.yml
jjbailey-io: Oops... yeah, that's what you get for copy/pasting from other repos. v2 will have this issue too.

ShaneDuncan602 (approved):
Looks good to me.

VoR0220:
tests failing. Uh oh. @jjbailey-io did you cancel these?

PR #421, opened 2024-07-30
Deployment script updates

TJ-Everett:


script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -28,11 +30,14 @@ read GAS_NUMBER
...
-forge script script/DeployAllModulesPt3.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
-forge script script/DeployAllModulesPt4.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+DEPLOYMENT=$(echo "$DEPLOYMENT" | tr '[:upper:]' '[:lower:]')
+if [ "$DEPLOYMENT" = "y" ]; then 
+  forge script script/DeployAllModulesPt1.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+  sh script/ParseProtocolDeploy.sh --chainid $CHAIN_ID
+  forge script script/DeployAllModulesPt2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+  forge script script/DeployAllModulesPt3.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+  forge script script/DeployAllModulesPt4.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+fi
```
ShaneDuncan602: This should probably be DeployProtocol.sh. It does all of these steps with the same parameters.

ShaneDuncan602 (commented):
The DeployERC20Handler was broken up because it was too large for testnet deployment, correct? If so, I think we should go ahead and split up the ERC721 handler script or create a ticket to do so.

script/clientScripts/Application_Deploy_Example.sh
TJ-Everett: Great point I will get that switched over to call the single script instead

TJ-Everett:
> The DeployERC20Handler was broken up because it was too large for testnet deployment, correct? If so, I think we should go ahead and split up the ERC721 handler script or create a ticket to do so.

I will create a ticket for this. Thanks Shane

script/clientScripts/Application_Deploy_Example.sh
TJ-Everett: I need a bash-fu pointer on how to pass the GAS_NUMBER variable from the first script into the DeployProtocol.sh script

mpetersoCode55: you should be able to export it in your current script so that it's available in the environment DeployProtocol.sh is called in. So for this specific example:
export GAS_NUMER=$GAS_NUMER

script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -58,25 +63,25 @@ if [ "$DEPLOYED" = "y" ]; then
...
-forge script script/clientScripts/Application_Deploy_04_ApplicationNFTUpgradeablePt2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
-# Optional: If you would like to run through tests using the upgradeable token
-# sh script/SubUpgradeableTokenForRegularToken.sh
-forge script script/clientScripts/Application_Deploy_05_Oracle.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
-sh script/ParseApplicationDeploy.sh 4 --chainid $CHAIN_ID
-forge script script/clientScripts/Application_Deploy_06_Pricing.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
-sh script/ParseApplicationDeploy.sh 5 --chainid $CHAIN_ID
-forge script script/clientScripts/Application_Deploy_07_ApplicationAdminRoles.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
\ No newline at end of file
+    forge script script/clientScripts/Application_Deploy_03_ApplicationFT2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
```
mpetersoCode55, line 71: So this change basically say if you answered the question "Do you already have an ERC20 deployed" with yes then just deploy the handler diamond. Otherwise deploy everything else.
This works, but I think we could make it more configurable. Originally that question was only the deciding factor for if we deploy just an ERC20 handler diamond or the example ERC20 and ERC20 Handler (ApplicationFT1.s.sol)
My recommendation would be to have this question still only determine that. Then have a separate question about whether or not the user would like to deploy the other example contracts.

mpetersoCode55 (approved):
Nice bashing man! 
I'm approving because I didn't find any actual issues, but I do have one suggestion on how we can make the script a little more configurable/versatile (take it or leave it, the way it currently is won't cause any issues just requires a little bit of tribal knowledge when running the script)

ShaneDuncan602 (approved):
Nice job!

PR #423, opened 2024-08-02
SE-935-Remove-AppManager-References-from-Deployment-Test-Scripts

ShaneDuncan602:
…ployments with the DeployExampleApplication.sh


deployAppERC721Test.sh
```diff
@@ -16,7 +16,7 @@ GREEN='\033[0;32m'
 YELLOW='\033[0;33m'
 NC='\033[0m' # No Color
 
-./foundry-version-check.sh
+# ./foundry-version-check.sh
```
VoR0220: Do we get rid of this? If not, maybe we can get a comment?

VoR0220 (approved):
Just one small thing and other than that, this looks good to go.

deployAppERC721Test.sh
ShaneDuncan602: Nope. Good catch. This should not be commented out.

TJ-Everett (approved):
Looks great! Approved

PR #424, opened 2024-08-06
Update TokenRegistered Event & Add UpdateRegisteredToken fn

TJ-Everett:


test/util/TestCommonFoundry.sol
```diff
@@ -663,6 +663,7 @@ abstract contract TestCommonFoundry is TestCommon, EndWithStopPrank, EnabledActi
         switchToAppAdministrator();
         erc721.connectHandlerToToken(address(handler));
         /// register the token
+        emit AD1467_TokenRegistered(name, address(erc721), 1);
```
Palmerg4, line 667: Missing an expectEmit here

src/client/application/AppManager.sol
```diff
@@ -652,20 +652,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
...
         if (!isTokenRegistered[_tokenAddress]) {
             _addAddressWithMapping(tokenList, tokenToIndex, isTokenRegistered, _tokenAddress);
             registeredHandlers[ProtocolTokenCommon(_tokenAddress).getHandlerAddress()] = true;
-            emit AD1467_TokenRegistered(_token, _tokenAddress);
+            try IERC165(_tokenAddress).supportsInterface(0x80ac58cd) returns (bool isERC721) {
+                if (isERC721){
+                    tokenType = 1;
+                    emit AD1467_TokenRegistered(_token, _tokenAddress, tokenType);
+                }
+                tokenType = 0;
```
Palmerg4: Why set tokenType to 0 here?

TJ-Everett: My thought on the try catch was 
try sets the isERC721 bool 
>>>>If true - set to 1 
>>>>>else set to 0 
on catch 
>>>>>>>set to 0 
This way if the ERC721 doesn't support erc165 it can still register the token - they just need to use the new update function

src/client/application/AppManager.sol
```diff
@@ -652,20 +652,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
...
+     * @param _token The token identifier of registered token
+     * @param _tokenAddress Address corresponding to the tokenId
+     * @param _tokenType The token type to update registered token
+     */
+    function updateRegisteredToken(string calldata _token, address _tokenAddress, uint8 _tokenType) external onlyRole(APP_ADMIN_ROLE) {
+        if (_tokenAddress == address(0)) revert ZeroAddress();
+        if (!isTokenRegistered[_tokenAddress]) {
+            registerToken(_token, _tokenAddress); 
+        }
+        emit AD1467_TokenRegistered(_token, _tokenAddress, _tokenType);
```
Palmerg4, line 696: This event seems a bit redundant, do we want this event to fire when only updating the token name?

src/client/application/AppManager.sol
```diff
@@ -652,20 +652,48 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
...
+    /**
+     * @dev This function Updates the Registered Token to an ERC721 token that does not support ERC165 interface. 
+     * @notice Token type is not stored on chain and is only to update events parameters for off chain databasing. 
+     * @param _token The token identifier of registered token
+     * @param _tokenAddress Address corresponding to the tokenId
+     * @param _tokenType The token type to update registered token
+     */
+    function updateRegisteredToken(string calldata _token, address _tokenAddress, uint8 _tokenType) external onlyRole(APP_ADMIN_ROLE) {
+        if (_tokenAddress == address(0)) revert ZeroAddress();
+        if (!isTokenRegistered[_tokenAddress]) {
```
Palmerg4: I think we can just call registerToken here, it has this check included

src/client/application/AppManager.sol
TJ-Everett: I was thinking that this would be better on Doom since they are already listening for this specific event to set the token type in the indexer. If we want this to be more specific we can have a different event name since its being updated? 

Idea here is that we are not actually storing the tokenType on chain just a trigger for the indexer.

src/client/application/AppManager.sol
Palmerg4: Shouldnt it be an else block though?

TJ-Everett: Ahh, yes. Thank you!

Palmerg4 (changes requested):
Just a few small things. One question I have is do we want to support 1155 here as well in this ticket?

TJ-Everett:
> Just a few small things. One question I have is do we want to support 1155 here as well in this ticket?

I don't think so. I think we should (if we go the 1155 route in V1) do an entire planned effort for that. We could write tests and integration strategies and include any updates in that. However I can happily be overruled on that.

src/client/application/AppManager.sol
```diff
@@ -652,20 +652,46 @@ contract AppManager is IAppManager, AccessControlEnumerable, IAppLevelEvents, IA
...
+    function registerToken(string calldata _token, address _tokenAddress) public onlyRole(APP_ADMIN_ROLE) {
+        uint8 tokenType; 
         if (_tokenAddress == address(0)) revert ZeroAddress();
         tokenToAddress[_token] = _tokenAddress;
         addressToToken[_tokenAddress] = _token;
         if (!isTokenRegistered[_tokenAddress]) {
             _addAddressWithMapping(tokenList, tokenToIndex, isTokenRegistered, _tokenAddress);
             registeredHandlers[ProtocolTokenCommon(_tokenAddress).getHandlerAddress()] = true;
-            emit AD1467_TokenRegistered(_token, _tokenAddress);
+            try IERC165(_tokenAddress).supportsInterface(0x80ac58cd) returns (bool isERC721) {
```
ShaneDuncan602, line 669: Can we add a comment in here that explains which interface this is checking for and the value meaning for tokenType?

ShaneDuncan602 (approved):
Great job.

Palmerg4 (approved):
Looks great!

PR #426, opened 2024-08-07
SE-951-Create-FeeCollected-Event-for-ERC20

ShaneDuncan602:


src/common/IEvents.sol
```diff
@@ -148,8 +148,11 @@ interface ITokenHandlerEvents is IAppManagerAddressSet{
  */
 
 interface IApplicationEvents is IAppManagerAddressSet{
-    /// Application Handler
-    event AD1467_HandlerConnected(address indexed handlerAddress, address indexed assetAddress); 
+    
+    ///ProtocolERC20
+    event AD1467_NewTokenDeployed(address indexed appManagerAddress);
```
Palmerg4, line 153: These events were just removed in Taylers ticket, which is now merged into main

test/util/UtilProtocolERC20.sol
```diff
@@ -121,28 +109,38 @@ contract UtilProtocolERC20 is ERC20, ERC165, ERC20Burnable, ERC20FlashMint, Prot
...
     }
 
+    /**
+     * @dev This transfers all the P2P transfer fees to the individual fee sinks
+     * @param from sender address
+     * @param amount number of tokens being transferred
+     */
+    function _handleFees(address from, uint256 amount) internal returns (uint256) {
+        address[] memory targetAccounts;
+        int24[] memory feePercentages;
```
Palmerg4, line 126: Whats the reason this is an int and not a uint?

test/util/UtilProtocolERC20.sol
```diff
@@ -121,28 +109,38 @@ contract UtilProtocolERC20 is ERC20, ERC165, ERC20Burnable, ERC20FlashMint, Prot
...
+        uint256 fees = 0;
+        (targetAccounts, feePercentages) = FeesFacet(handlerAddress).getApplicableFees(from, balanceOf(from));
+        for (uint i; i < feePercentages.length; ++i) {
+            if (feePercentages[i] > 0) {
+                // trim the fee and send it to the target fee sink account
+                uint fee = (amount * uint24(feePercentages[i])) / 10000;
```
Palmerg4, line 132: feePercentages type casted to a uint here making me thing the usage of int on ln 126 is supposed to be uint

TJ-Everett: The fee is an int (vs uint) at creation so that we can send negative integers to the protocol for fee a discount of fees if they are eligible for that (a discount is a reduction of fees. That is returned by the protocol as a unsigned int for the actual transfer value so the code doesn't freak out. 

So this is int here inside the token because the fees facet returns the int24 vs uint24

TJ-Everett (approved):
Approved for me!

Palmerg4 (approved):
Approved for me, int for feePercentage makes sense to me now

src/common/IEvents.sol
ShaneDuncan602: Rebased with main and they're now gone.

PR #427, opened 2024-08-07
SE-1000 ensure protocol events are fired correctly, add ruleId to app…

Palmerg4:
Added lots of expectEmits throughout tests and helpers, added ruleId to AD1467_ApplicationHandlerActionActivated and AD1467_ApplicationHandlerActionDeactivated


TJ-Everett (commented):
Waiting on final word from Doom crew on how they want events handled for the ruleId param or if they even want that.

ShaneDuncan602 (approved):
Good job on this. Sending 0 in the events for non oracle rules is not ideal but it's pretty much our only way forward to make it work with the UI.

PR #432, opened 2024-08-12
Se 999 reconfigure application deploy.sh

TJ-Everett:


script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -24,6 +24,8 @@ echo "Rule Processor Deployment Required? (y or n)"
 read DEPLOYMENT
```
ShaneDuncan602, line 24: I know you didn't change this but this question is a little confusing without the context. Can we change it to, "Does The Protocol(Rule Processor) need to be deployed to the target chain?"

script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -64,23 +87,155 @@ if [ "$DEPLOYED" = "y" ]; then
...
+    echo
+    echo "Not a valid answer (y or n)"
+    echo "Do you want to deploy an ERC721 (y or n)?"
+    read ERC721_DEPLOYMENT
+    ERC721_DEPLOYMENT=$(echo "$ERC721_DEPLOYMENT" | tr '[:upper:]' '[:lower:]')  
+    done
+    if [ "$ERC721_DEPLOYMENT" = "y" ]; then
+      forge script script/clientScripts/Application_Deploy_04_ApplicationNFT.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
+      sh script/ParseApplicationDeploy.sh 3 --chainid $CHAIN_ID 
+      forge script script/clientScripts/Application_Deploy_04_ApplicationNFTPt2.s.sol --ffi --broadcast --rpc-url $ETH_RPC_URL --gas-price $GAS_NUMBER
```
ShaneDuncan602: After all of these deployments, can we output the addresses in the terminal? Some have them and some don't. It would be nice if they were a different color, like blue or something.

script/clientScripts/Application_Deploy_Example.sh
```diff
@@ -64,23 +87,155 @@ if [ "$DEPLOYED" = "y" ]; then
...
+    read ERC20_DEPLOYMENT
+    ERC20_DEPLOYMENT=$(echo "$ERC20_DEPLOYMENT" | tr '[:upper:]' '[:lower:]')  
+    while [ "y" != "$ERC20_DEPLOYMENT" ] && [ "n" != "$ERC20_DEPLOYMENT" ] ; do
+    echo
+    echo "Not a valid answer (y or n)"
+    echo "Do you want to deploy an ERC20 (y or n)?"
+    read ERC20_DEPLOYMENT
+    ERC20_DEPLOYMENT=$(echo "$ERC20_DEPLOYMENT" | tr '[:upper:]' '[:lower:]')  
+    done
+    if [ "$ERC20_DEPLOYMENT" = "y" ]; then
```
ShaneDuncan602, line 102: While testing, I was able to say no to deploying an ERC20 and never go the option to deploy an ERC721

ShaneDuncan602 (changes requested):
Since this is a client facing script, I think we should prompt them for all the addresses and not rely on the .env. The clustered scripts could still rely on it to run in groups, but I don't think we should expect the users to ever enter their addresses in there. It should all be prompts. I think this blows the scope of this ticket though so it probably requires some discussion.

TJ-Everett:
Ticket was OBE: Closing PR

PR #441, opened 2024-09-09
SE-1095-Add-licensing-information

ShaneDuncan602:
Also updated foundry version to nightly-5ac78a9cd4b94dc53d1fe5e0f42372b28b5a7559

### DIFF
--- a/src/example/LICENSE
+++ b/src/example/LICENSE
@@ -1,0 +1,445 @@
+This software is available under your choice of the GNU General Public
+License, version 2 or later, or the Business Source License, as set
+forth below.
+
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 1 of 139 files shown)


src/example/LICENSE
```diff
@@ -0,0 +1,445 @@
...
+
+Additional Use Grant: Any uses listed and defined at
+                      v3-core-license-grants.uniswap.eth
+
+Change Date:          The earlier of 2023-04-01 or a date specified at
+                      v3-core-license-date.uniswap.eth
+
+Change License:       GNU General Public License v2.0 or later
+
+-----------------------------------------------------------------------------
```
TJ-Everett: Is this an example section or should we update this?

TJ-Everett (commented):
Just a quick question on that section for the example licensing

src/example/LICENSE
ShaneDuncan602: Good catch. This needs to be updated.

TJ-Everett (approved):
Approved. Great work on this

PR #443, opened 2024-09-10
SE-1103-Increment-version-to-2.1.0-and-Add-Verification-to-Deployments

ShaneDuncan602:


TJ-Everett (approved):
Looks great! Even got the newest tests in there updated for versioning. Ready to rock an roll

PR #447, opened 2024-09-18
Remove deployed token dependency in DeploymentProcessor.t.sol

TJ-Everett:


test/util/TestCommonFoundry.sol
```diff
@@ -267,6 +267,17 @@ abstract contract TestCommonFoundry is TestCommon, EndWithStopPrank, EnabledActi
...
+     * @dev Deploy and set up the AppManager
+     */
+    function setUpAppManager() public endWithStopPrank {
+        switchToSuperAdmin();
+        applicationAppManager = _createAppManager();
+        switchToAppAdministrator(); // app admin should set up everything after creation of the appManager
+        applicationAppManager.setNewApplicationHandlerAddress(address(_createAppHandler(ruleProcessor, applicationAppManager)));
+        applicationHandler = ApplicationHandler(applicationAppManager.getHandlerAddress());
+    }
+
```
ShaneDuncan602: I think the createAppManager function does the same thing as this one. You should be able to use it.

TJ-Everett: Brilliant. Made the change and re running tests

ShaneDuncan602 (approved):
Looks awesome

PR #450, opened 2024-09-24
DemoSetup script - anvil already running error catch

dmulvi:
If anvil (or something else on port 8545) is already running, anvil will fail to start in the demo script. This just adds a quick check for the error in anvil_output.txt and prints a message for the user if caught. 

This also sets the output filename to be `.test_env` instead of `test_env` to match the message.


TJ-Everett (approved):
Nice! I Like this a lot. Good idea and approved for me.

PR #453, opened 2024-09-27
adds $ to ETH_RPC_URL command example to properly pull value from env

dmulvi:
There are 3 commands in the deploy protocol readme that are missing the $ in front of ETH_RPC_URL. This fixes that.


ShaneDuncan602 (approved):
Great catch. Approved for me.

TJ-Everett (approved):
Approved for me too

PR #455, opened 2024-10-08
SE-994 Thrackle `foundryup` (Precompiled, Versioned Foundry Tools)

excalq:
## Use of Thrackle's `foundryup` to provide versioned, pre-compiled Foundry binaries.

1. Updates `foundry.lock` to `v0.2.0`, the current versioned release in `thrackle-io/foundry`, away from a nightly build hash.
2. Updates Documentation regarding use of `foundryup`, as we recommend installing Thrackle's version, rather than from `foundry.paradigm.xyz`.
3. `Dockerfile`: Loads precompiled, versioned binaries, via the modified `foundryup`.

This produces a Docker image which now builds in 6.5 minutes (without cache)! 
image

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -26,12 +26,7 @@ jobs:
       #    sudo swapon /swapfile
       #    sudo swapon --show
       
-      - name: Install Foundry
-        run: |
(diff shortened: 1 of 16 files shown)


TJ-Everett (approved):
This looks good to me. Approved

ShaneDuncan602 (approved):

PR #456, opened 2024-10-08
docs: update rules applicability table

dmulvi:
- remove "Rule" from end of rule names for consistency
- fix link for account-max-value-out-by-access-level
- re-ogranize rules to list by application grouping, then token
- remove 'AMM' from level column for oracle rule


TJ-Everett (approved):
Looks great! Approved

PR #463, opened 2024-10-28
Update Max Trade Size rule documentation

TJ-Everett:
Remove unnecessary import from IProtocolTokenHandler.sol


ShaneDuncan602 (approved):
Much clearer now

PR #464, opened 2024-10-28
SE-848-Output-rule-engine-gas-estimates-to-documentation_B

ShaneDuncan602:


docs/userGuides/deployment/DEPLOYMENT-DIRECTORY.md
```diff
@@ -3,8 +3,8 @@
 
 | Version | Environment     | Contract Name        | Contract Address |
 | :--- |:----------------|:---------------------| :--- | 
-| v2.2.0 | Base Mainnet | RuleProcssorDiamond | 0x474d3c9fd6f35d741058dc2d6496263669ea49b6
-| v2.2.0 | Base Sepolia | RuleProcssorDiamond | 0x8fe43b3bcf8410b913dd1c84dc8af28daafd9777
+| v2.2.0 | Base Mainnet | RuleProcessorDiamond | 0x474d3c9fd6f35d741058dc2d6496263669ea49b6
```
TJ-Everett, line 6: Eagle eyed catch on this one!

test/util/MinimalERC721LegacySell.sol
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: UNLICENSED
```
TJ-Everett, line 1: Am I understanding the purpose of this file and the erc20 is to test the legacy versions of the tokens we have supported in the past? I.e. the different protocol hook we used to have?

docs/userGuides/gasReports/GAS_ESTIMATES_PER_RULE.md
```diff
@@ -0,0 +1,117 @@
+# Gas Estimate Per Rule ERC721
+[![Project Version][version-image]][version-url]
+
+---
+
+## Purpose
+
+This chart shows how individual active rules affect gas consumption of basic token actions.
+
+Integrating the Rules Engine without activating any rules add n gas units.
```
mpetersoCode55: Should "n gas units" be "no gas units"?

TJ-Everett (approved):
Approved for me. I really like the refactor on this. The final test file looks so much cleaner

mpetersoCode55 (approved):

test/util/MinimalERC721LegacySell.sol
ShaneDuncan602: That was the original intention of this contract but the old hook was never actually used. I repurposed it to make use of its action parameter. It makes it possible to test any action without setting up an elaborate transaction for sells and buys.

docs/userGuides/gasReports/GAS_ESTIMATES_PER_RULE.md
ShaneDuncan602: Good catch. This sentence is all out of whack.

PR #465, opened 2024-10-29
Se 804 modify abi scripts to replace diamond json in deployment directory

VoR0220:
This fixes UpgradeAFacet script and also fixes Revert A Facet. The flow works like this. You input the name of the facet you want to upgrade and the timestamp directory of which diamond you want to upgrade. Likewise for Revert a Facet but also requires the address of what Facet you want to revert to. Adding documentation clarification now.

### DIFF
--- a/docs/userGuides/architecture/common/DIAMOND-UPGRADE-REVERSION.md
+++ b/docs/userGuides/architecture/common/DIAMOND-UPGRADE-REVERSION.md
@@ -9,7 +9,7 @@ If a diamond upgrade reversion is required, the next steps must be followed:
     ```
     DEPLOYMENT_OWNER_KEY=<YOUR_DEPLOYMENT_OWNER_KEY>
     FACET_NAME_TO_REVERT=<THE_NAME_OF_THE_FACET_CONTRACT_TO_REVERT>
-    DIAMOND_TO_UPGRADE=<THE_NAME_OF_THE_DIAMOND_AS_IT_IS_RECORDED_IN_THE_DEPLOYMENTS>
+    FACET_TIMESTAMP=<THE_TIMESTAMP_OF_THE_DIAMOND_TO_REVERT> # Find this in the deployments directory, its the name of the folder for when the diamond was deployed
     REVERT_TO_FACET_ADDRESS=<THE_PREVIOUS_RECORDED_ADDRESS_OF_THE_FACET>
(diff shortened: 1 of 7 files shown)


script/RevertAFacetUpgrade.s.sol
```diff
@@ -27,21 +27,18 @@ contract RevertAFacetupgrade is Script, DeployBase {
...
     address revertToFacetAddress;
-    bool recordAllChains;
 
     /**
      * @dev This is the main function that gets called by the Makefile or CLI
      */
     function run() external {
 
         privateKey = vm.envUint("DEPLOYMENT_OWNER_KEY");
-        recordAllChains = vm.envBool("RECORD_DEPLOYMENTS_FOR_ALL_CHAINS");
```
TJ-Everett, line 40: Is the idea for removing this, that we technically already have this recorded since this is reverting back to a different facet?

TJ-Everett (approved):
Looks great, approved for me.

script/RevertAFacetUpgrade.s.sol
VoR0220: iirc this is because it's never used again.

docs/userGuides/architecture/common/DIAMOND-UPGRADE-REVERSION.md
```diff
@@ -9,7 +9,7 @@ If a diamond upgrade reversion is required, the next steps must be followed:
     ```
     DEPLOYMENT_OWNER_KEY=<YOUR_DEPLOYMENT_OWNER_KEY>
     FACET_NAME_TO_REVERT=<THE_NAME_OF_THE_FACET_CONTRACT_TO_REVERT>
-    DIAMOND_TO_UPGRADE=<THE_NAME_OF_THE_DIAMOND_AS_IT_IS_RECORDED_IN_THE_DEPLOYMENTS>
+    FACET_TIMESTAMP=<THE_TIMESTAMP_OF_THE_DIAMOND_TO_REVERT> # Find this in the deployments directory, its the name of the folder for when the diamond was deployed
```
ShaneDuncan602, line 12: I don't see this variable in the .env

ShaneDuncan602 (changes requested):
This looks really close. I just tested a live update though and found that it isn’t updating the address of the facet with the diamonds.json. Since that file is the new source of truth for facet addresses, it should also replace the address of the old facet with the new one. Also, we should have the FACET_TIMESTAMP in the .env

ShaneDuncan602 (approved):
Great job. This looks ready to me.

PR #466, opened 2024-10-29
Removed references to renamed rule (Max Transaction Size Per Period)

TJ-Everett:
Removed from docs/userGuide/rules


ShaneDuncan602 (approved):
Looks ready

PR #471, opened 2024-11-11
Ran the upgrade version script for you

VoR0220:
I added a couple lines because I saw it wasn't upgrading the version in certain python scripts that might now have the version needed. When I ran it it seemed to update the docs just fine? Adding them in to the userGuides.

### DIFF
--- a/test/util/TestCommonFoundry.sol
+++ b/test/util/TestCommonFoundry.sol
@@ -735,7 +735,7 @@ abstract contract TestCommonFoundry is TestCommon, EndWithStopPrank, EnabledActi
         erc721.connectHandlerToToken(address(handler));
         /// register the token
         applicationAppManager.registerToken(name, address(erc721));
-        HandlerVersionFacet(address(handler)).updateVersion("2.1.0");
+        HandlerVersionFacet(address(handler)).updateVersion("2.2.0");
     }
 
     function deployAndSetupERC20MinLegacySell(string memory name, string memory symbol) internal endWithStopPrank returns (MinimalERC20LegacySell erc20, HandlerDiamond handler) {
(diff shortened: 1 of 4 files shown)


test/util/TestCommonFoundry.sol
```diff
@@ -735,7 +735,7 @@ abstract contract TestCommonFoundry is TestCommon, EndWithStopPrank, EnabledActi
         erc721.connectHandlerToToken(address(handler));
         /// register the token
         applicationAppManager.registerToken(name, address(erc721));
-        HandlerVersionFacet(address(handler)).updateVersion("2.1.0");
+        HandlerVersionFacet(address(handler)).updateVersion("2.2.0");
```
ShaneDuncan602, line 738: I don't see where this was added to the versioning script. Was this already being changed and just got missed?

TJ-Everett: +1 to this Question.

TJ-Everett (commented):
Just a comment about if an updated file was missed or needs to be added to the script. 
Also just a psa that we want to ensure that our commit messages are applicable to the changes we make within the scope of the ticket. That way we can reference back to these and know what was changed within the commit.

test/util/TestCommonFoundry.sol
VoR0220: No I just ran the script and this was the output? I'm confused myself. The script works on my local.

ShaneDuncan602: Hmm, might have been a merge issue that was never detected.

ShaneDuncan602 (approved):
Good job!

mpetersoCode55 (approved):

PR #475, opened 2024-11-18
SE-1112-Full-stack-Liquidity-feature-polishing

bfcoursewool:
Just slightly modifying the deploy script for docker deploys (ie, the dev stack), so that we also deploy an unregistered ERC20 (WOLFY), that we can work with in the UI to use as a an xToken in a pool.


TJ-Everett (approved):
Approved. Looks good to me.

PR #480, opened 2024-12-10
Update NPM package contents to include deployment and utility scripts

TJ-Everett:


package.json
```diff
@@ -11,6 +11,22 @@
...
+    "script/DeployAllModulesPt4.s.sol",
+    "script/DeployProcessorDiamond.s.sol",
+    "script/DiamondScriptUtil.sol",
+    "script/EnabledActionPerRuleArray.sol",
+    "script/ParseApplicationDeploy.sh",
+    "script/ParseProtocolDeploy.sh",
+    "script/RevertAFacetUpgrade.s.sol",
+    "script/SetupProtocolDeploy.sh",
+    "script/SubUpgradeableTokenForRegularToken.sh",
+    "script/UpgradeAProtocolFacet.s.sol",
```
ShaneDuncan602, line 30: Can't we just put the containing folders instead of listing the files? For instance, "script", "src", "lib", etc.? Also, the ticket says "everthing" and it's bold and uppercase. We probably want to add the deployments and test folders or get confirmation that we shouldn't.

TJ-Everett: Great questions! Resolution is to add deployments directory but not tests or full scripts folder

ShaneDuncan602 (approved):
Looks good!

PR #484, opened 2025-01-16
Forte Rules Engine Gas Reports

TJ-Everett:
Updated GasReports.t.sol to no longer be used 
Added GasReportV2 for ERC20 transfer focused gas tests


mpetersoCode55 (approved):
Looks great! Approved

PR #491, opened 2025-02-27
RL-266 Corrected Bytes strings and Getters

TJ-Everett:
Updated getter for accountMaxReceivedByAccessLevel in rules diamond


ShaneDuncan602 (approved):
Looks great.

